Issue #1, opened 2021-01-26
Technology stack for CCM


pushyamig:
Currently the tech stack followed internally are Python/Django. But with current CCM redesign that is not necessarily the ideal choices. One proposal is experiment with  Full stack Java script. 

CCM Design wireframes
https://github.com/tl-its-umich-edu/UX-Work/tree/master/ccm/wireframes

chrisrrowland:
Did some experimenting with svelte/typescript/material [here ](https://github.com/chrisrrowland/svelte1)

pushyamig:
We are moving forward with React stack and this issue served the purpose

PR #8, opened 2021-02-09
Add basic setup for Sapper/Svelte/Express/Docker (#10)

ssciolla:
This PR adds a basic project setup for Sapper, Svelte, Express, TypeScript, and Docker, allowing us to experiment with a single-server approach with this combination of technologies. Docker artifacts are functional but works in progress. The PR works towards resolving issue #1, but does not resolve it on its own. This PR resolves issue #10.


pushyamig:
I was reviewing this PR and found these things
1. docker-compose up is resulting in errors and failing at this `sapper dev` due to `sh: 1: sapper: not found` 
2. I am assuming the Svelte code is bundled into a single JS file but I Don't see that file being generated at all. But this is only my Assumption

chrisrrowland:
> I was reviewing this PR and found these things
> 
> 1. docker-compose up is resulting in errors and failing at this `sapper dev` due to `sh: 1: sapper: not found`

It started up ok for me but I can't access the site in the browser.

```
Creating ccm_web ... done
Attaching to ccm_web
ccm_web | Executing test start commands
ccm_web | 
ccm_web | > canvas-course-manager@0.0.1 dev /app
ccm_web | > sapper dev
ccm_web | 
ccm_web | ✔ server (4.3s)
ccm_web | ✔ client (4.3s)
ccm_web | > Listening on http://localhost:3000
ccm_web | ✔ service worker (907ms)
```

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    ports:
+      - 3001:3000
+      # Added to eliminate repeated console errors.
+      # I need to look into this more. Seems like only a dev environment problem.
+      # https://github.com/sveltejs/sapper-template/issues/175
+      - 10000:10000
+    volumes:
+      # Facilitates re-loading
+      - ./ccm_web/src:/app/src:z
+      - ./ccm_web/static:/app/static:z
```
ssciolla, line 21: @pushyamig and @chrisrrowland, this should resolve the problem. Basically, the previous volume mapping seemed to be wiping out `node_modules` in `ccm_web`. The other files in `ccm_web` but not in `src` or `static` probably need a rebuild anyway when they're changed.

pushyamig:
I see the PR is working and I need to launch the browser with URL http://localhost:3001, the log message says launching from the 3000. I am assuming this is from the Sapper default port. I think we might want to control the port of choice from Docker compose. But for Prototype this is file

pushyamig:
The Reloading feature is working when I changed the file content in the `index.svelte`. In Myla we have to run a separate container for making the file available to Django. Here it doesn't look like that is the case. How are these files served up?

pushyamig:
What is the integration point between Sapper and Express? how are they connected?

pushyamig:
>  I am assuming this is from the Sapper default port. I think we might want to control the port of choice from Docker compose. But for Prototype this is fine

Port is controlled from Docker, ignore this comment

ssciolla:
> What is the integration point between Sapper and Express? how are they connected?

See `ccm_web/src/server.ts`. The Express server application loads Sapper as middleware.

ssciolla:
> The Reloading feature is working when I changed the file content in the `index.svelte`. In Myla we have to run a separate container for making the file available to Django. Here it doesn't look like that is the case. How are these files served up?

I'm still wrapping my head around all this works myself. My current understanding is that when changes are saved, they are seen somehow by Sapper, and they are re-generated and run in a directory called `__sapper__` (which you can see at the root inside the container, `docker exec -it ccm_web /bin/bash`). If you're running with `dev`, things get populated under `__sapper__/dev`. Right now I'm also building for prod in the `Dockerfile` (`npm run build`), and that generates an optimized version under `__sapper__/build`. We could probably make that conditional in the `Dockerfile`, which I've noted in a comment. What you're seeing in the browser is the result of `node __sapper__/dev` or `/build`. Again, I don't have mastery over how this works yet, so please correct me if you see something working differently.

pushyamig (approved):
This Looks Good for the first attempt. @chrisrrowland what do you say?

ccm_web/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Establish base image to build upon
+FROM node:14-slim AS node
+WORKDIR /app/
+
+EXPOSE 3000
```
pushyamig, line 5: I would suggest to keep the port different from 3000 since it is interfering with Myla ports

pushyamig:
> I'm still wrapping my head around how all this works myself. My current understanding is that when changes are saved, they are seen somehow by Sapper, and they are re-generated and run from a directory called __sapper__ (which you can see at the root inside the container, docker exec -it ccm_web /bin/bash). If you're running with dev, things get populated under __sapper__/dev. Right now I'm also building for prod in the Dockerfile (npm run build), and that generates an optimized version under __sapper__/build. We could probably make that conditional in the Dockerfile, which I've noted in a comment. What you're seeing in the browser is the result of node __sapper__/dev or /build. Again, I don't have mastery over how this works yet, so please correct me if you see something working differently.

I believe rollup (a WebPack alternative) is involved in generating the files under at least /build, but maybe also dev.


That's how I felt as well. I am also making sense of this because this step of hot reloading has been simplified and don't need to put a great effort to this. But as we work more understanding grows and we can be in better position to assessment

ccm_web/Dockerfile
chrisrrowland: Change this to 3333 and do
```
    ports:
      - 3333:3333
```
in docker-compose.yml and the startup link works as well as not being on 3000

ssciolla: Sure, I can change to `3333` if you both think that makes sense.

chrisrrowland: I don't know that making a decision here based on myla is particularly important, but its just a number so what does it hurt.  
I  mostly want to be able to click the link produced on startup and matching the ports up in docker-compose.yaml seems to fix that part of it.

ssciolla: K, I don't know what the best practice is here. I'm not aware of any particular issues with mapping to the same port, so sure, I'll do it.

ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/8/commits/7bd104a36edce63c8fcffab4b4e41bb18767f34a

chrisrrowland (approved):

PR #24, opened 2021-03-01
Add Docker-ized basic stack with Express, React, and TypeScript (#14)

### LINKED ISSUE #14: Getting the Initial Local setup going with React Stack
This issue can be characterized having 
1. Have a Docker setup
2. Integrating with React Ecosystem along with ExpressJS. Use of NextJS is the Developer choice if faster for development.
3. Integrating with LTI launch, this can be addressed as separate issue. Database of choice seems to be Postgres but confirmation from MIDatabase  https://teamdynamix.umich.edu/TDNext/Apps/31/Tickets/TicketDet.aspx?TicketID=633962 would help back our decision.

ssciolla:
Features included:
* Basic Express server in TypeScript (`ccm_web/server`)
* Basic React client in TypeScript (`ccm_web/client`)
* Single image/container and `package.json` for development and production
* `docker-compose` workflow for development
* Live reloading on save across the stack (does not yet include Hot Module Replacement)
* Sample API route (see `components/TestConsumer.tsx` and the route in `app.ts`)
* Linting configuration using ESLint and Standard JS rules

The PR resolves #14.


ccm_web/client/Dockerfile.dev
```diff
@@ -0,0 +1,12 @@
+FROM node:14-slim AS node
+WORKDIR /app/
```
pushyamig, line 2: is this file being used?

ssciolla: Yes, by `docker-compose` for the development set up.

ccm_web/client/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "devDependencies": {
+    "typescript": "^4.2.2",
+    "@testing-library/jest-dom": "^5.11.9",
+    "@testing-library/react": "^11.2.5",
+    "@testing-library/user-event": "^12.7.3",
+    "@types/jest": "^26.0.20",
+    "@types/node": "^12.20.4",
+    "@types/react": "^17.0.2",
+    "@types/react-dom": "^17.0.1"
```
pushyamig, line 18: How is Typescript is being transpiled to Javascript? I don't see babel or webpack being used?

README.md
```diff
@@ -1 +1,47 @@
...
+Taken together, all the stages in `ccm_web/Dockerfile` will build an optimized image for production.
+
+You can build a production image, run a container, and use the application locally by performing
+the following steps, replacing `{PORT}` with the port you want the application to run on:
+
+1. Build one image with the server and static artifacts, including the bundled JavaScript frontend.
+
+```
+docker build -t ccm --build-arg PORT={PORT} ./ccm_web
+```
```
pushyamig, line 35: I see the build failure for this case
```
TypeScript error in /build/client/src/App.tsx(1,19):
Could not find a declaration file for module 'react'. '/build/node_modules/react/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`  TS7016
```

ssciolla: Uh-oh, let me take a look. I tested this, but I might have had some cache or directories disrupting the outcome.

ccm_web/client/package.json
ssciolla: My understanding at this point is that it's being taken care of by `react-scripts`. See the `react-scripts` `package.json` here: https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/package.json

We could certainly do some of this ourselves (i.e. using `webpack`), but this seems to work?

README.md
ssciolla: Reproduced. I guess some of those Typescript-related packages do need to be `dependencies`...

README.md
```diff
@@ -1 +1,47 @@
-## Canvas Course Manager
\ No newline at end of file
+## Canvas Course Manager
+
+## Development
```
pushyamig, line 3: I would really like to hear, the development set up is running on 2 container (client and server) and prod doesn't seems to that way

This setup seems to be new to me and I cannot connect why client and server running on 2 different instances. what is the purpose they needs to be running in separate ports?

ssciolla: We can discuss. Maybe there is another way to do this. The separate servers here I believe are most to allow us to leverage `react-scripts` and its reloading-for-dev and building-for-prod functionality. I was mostly seeing this pattern when I researched online, but I could have been missing something.

Yes, the production workflow is different, the Express application serves an optimized build as static assets.

ccm_web/client/package.json
pushyamig: But when you look at the react-scripts package.json, it is coming with lot more than extra package than what we need. For example if we don't do SASS why we need it etc. With SPA this will make it big single JS file (I really cannot see how big is the file currently KB or MB), load time increases as we add more stuff and code spitting will be needed soon. This is my 2 cents

README.md
ssciolla: Okay, I think I fixed this build problem here: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/24/commits/b5d981fdfce69048ce79f24bd4a4a77aafd34521

I'm looking into alternatives for development setups.

README.md
pushyamig: I see this [setup](https://levelup.gitconnected.com/how-to-setup-environment-using-react-webpack-express-babel-d5f1b572b678) with express and React using webpack ( with out docker though ) https://github.com/Jatin-8898/express-react-webpack-babel-setup

ssciolla: Thanks, I'll take a look. I think if I go through the `webpack` tutorial, I can probably do something like this without too much trouble. `react-scripts` (`create-react-app`) does seem to handle a lot of this stuff for you, but maybe with a lean setup like this, it's not needed, and we can limit development to one server.

pushyamig: I think Babel is not needed with Type script since ts-loader handles both types and transpiling. I believe, react-script is convenient but I feel it has lot of stuff and it is concealing what is going on underneath.

I think we still 2 watchers in Dev mode one looking at the express(BE) changes,`nodemon` is probably doing that. One looking at React changes (FE). The interplay will be interesting to see like how the react bundle changes noticed by BE. If BE reloads after a FE bundle change it is bit unclear right now.

ssciolla:
Okay, I'm going to continue testing this and see if I can tidy up `package.json` and minify CSS to decrease image size for production (we're at `239MB` for production), but I believe I've successfully transitioned away from `react-scripts` and separate servers to one `express` server with `webpack` middleware for development. I'm going to open this for review so we can keep moving forward. There's more we can do around tree-shaking, hot module replacement, and code splitting, but we can build upon this work later.

README.md
ssciolla: See my latest commits. FE changes will be registered by `webpack`. BE changes will trigger a full reset, including a new `webpack` bundling process, but I'm not sure if we can avoid that given the current one server configuration.

README.md
ssciolla: With webpack and `ts-loader` handling the TypeScript transformations, I think we can keep all of the `@types` packages as `devDependencies`. I will test this more.

ccm_web/client/package.json
ssciolla: I'm gonna mark this as resolved, but feel free to reopen if there's a lingering issue after the switch to custom `webpack`.

pushyamig:
Reviewing now! Nice work with webpack loader.

pushyamig:
What is the linter are you planning to use? if that is setup early on it sets up the stage further on

ccm_web/server/app.ts
```diff
@@ -0,0 +1,61 @@
...
+    ? devConfig.output.publicPath : undefined
+
+  app.use(
+    webpackDevMiddleware(compiler, {
+      publicPath: publicPath,
+      // Just copies what's used in memory to client/build (changing doesn't do anything)
+      writeToDisk: true
+    })
+  )
+
```
ssciolla, line 46: This works out pretty nicely in development, but it's a little awkward having to import `webpack` and related dependencies even in the production images. Now or later, we might want to think about either having a separate `dev` `app.ts`. Alternatively, we could use `webpack-dev-server`, but that starts to look similar to `react-scripts`. There might be a way to run two servers in one container? I haven't thought that through too much.

ssciolla:
> What is the linter are you planning to use? if that is setup early on it sets up the stage further on

Mmm, good question. I'll work on that. Are you okay with `standard`? I don't really care either way about 2 or 4 spaces for indents, but we should be consistent. I'd prefer not to use semicolons, but would be fine either way. Looks like `ts-standard` is a thing. I'll do that unless I hear otherwise. Looks like I may need to clean up some indentation already 😭

pushyamig:
> I'd prefer not to use semicolons, but would be fine either way. 

Me too. I got used to not doing that with Myla, I did not know what your choice is. I am ok using `ts-standard`

ccm_web/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# Prod stage
+FROM node:14-slim AS prod
+ENV NODE_ENV=production
+WORKDIR /app
+
+COPY --from=base /base/package.json ./
+RUN npm install --production
+
+COPY --from=build /build/dist/ ./
```
pushyamig, line 29: the `dist` folder in the container needs to map to local disk space, since as we are keep changing text new file will be created but We need to delete it without going into container. We need to see the `dist` file otherwise we forget to delete it and container size increases and we run into space issues

ccm_web/nmon_config.json
```diff
@@ -0,0 +1,12 @@
+{
```
pushyamig, line 1: what is the purpose of this file?

ssciolla: This configures `nodemon`, which is how changes to the backend (`app.ts`) re-load. See the `dev` script in `package.json`.

ccm_web/Dockerfile
ssciolla: I DMed you to discuss this. This step will only be performed for production builds. Happy to explain. I could of course be doing something wrong, so I'd like some further explanation.

ccm_web/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "ccm_web",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "dev": "nodemon --config nmon_config.json server/app.ts",
+    "build-client": "webpack --config webpack/webpack.prod.ts",
+    "build-server": "tsc",
+    "preprod": "npm run build-client && npm run build-server",
+    "prod": "NODE_ENV=production node dist/server/app.js"
```
pushyamig, line 10: It feels bit weird the `app.js` is in `dist` folder? isn't dist more or less temporary folder?

ssciolla: Where would you want to put it? You never edit `app.js` directly, it's always going to be generated by the TypeScript compiler.

ccm_web/webpack/webpack.dev.ts
```diff
@@ -0,0 +1,12 @@
+import * as webpack from 'webpack'
+import { merge } from 'webpack-merge'
+
+import commonConfig from './webpack.common'
+
+
+const devConfig: webpack.Configuration = merge(commonConfig, {
+    mode: 'development',
+    devtool: 'inline-source-map'
+})
```
pushyamig, line 10: I really don't feel the source map is working since i still see compiled stuff
Screen Shot 2021-03-05 at 2 58 01 PM

this is minor but can be handled later on

ccm_web/package.json
pushyamig: one a second thought on this, From the `Dockerfile` I can see last line is `CMD node server/app.js` and that is technically starting the server in production mode. is the line `"prod": "NODE_ENV=production node dist/server/app.js"` really called somewhere?

ssciolla: This command is to allow to run in production without Docker using `npm run prod`. I'm not using it in Docker because I'm hiding `dist` (coping only the contents of dist into the last stage, not the `dist` directory itself). Does that clarify? I could add a comment somewhere about this if you want.

ccm_web/webpack/webpack.dev.ts
ssciolla: Hm, it's close. I think what your saying is ideally you'd be able to view TypeScript there? Or at least get an accurate line number for the TypeScript file?

ccm_web/package.json
pushyamig: OK, make sense now. Adding a comment would be good.

ccm_web/webpack/webpack.dev.ts
pushyamig: sourcemap will exactly output the file as is in IDE. The benefit of this when debugging with breakpoint it is clear and it reflect with written code than compiled code

ssciolla: Working on it, hopefully won't be too difficult.

jonespm: What about just the `source-map` option?

ssciolla: Should be fixed with this change: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/24/commits/2e0ae9ddfe6133ddef63bfca3c70a7b3eece394c

We'll probably want to add a source map for production, probably the one @jonespm recommended.

pushyamig: you should not add source map to production since it adds to the launch time. you webpack.config reflects that as of now. should be good

docker-compose.yml
```diff
@@ -12,11 +12,13 @@ services:
     environment:
       - PORT=4000
       - NODE_ENV=development
+    env_file: .env
```
pushyamig, line 15: you added the `.env` configuration without adding file. The build is  failing if you merge it as is. either remove this line or add the file

ccm_web/webpack/webpack.dev.ts
ssciolla: @pushyamig, `webpack` actually says differently: https://webpack.js.org/guides/production/#source-mapping Different options seem to have different performance implications.

https://webpack.js.org/configuration/devtool/#devtool

docker-compose.yml
ssciolla: Wups, I did not mean to commit that. I will remove.

ssciolla: Done. https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/24/commits/969d576677f1df660a3d1c703420fc181a598d51

pushyamig:
This looks Good and nice Job! This is ready to merge from me. I think we will learn more things as we go and make improvement to the setup. Merging this allows for further development.

pushyamig (approved):

ssciolla:
> Me too. I got used to not doing that with Myla, I did not know what your choice is. I am ok using `ts-standard`

Setting up `ts-standard` was not as easy as I expected. I got it working for the `server`, but I can't get it to recognize the `client` as well. I might look into `eslint` configured with TypeScript and a few rules, as I might be able to specify multiple projects. In any case, not going to finish it today. I'll return to it on Monday, though I might make this a separate issue.

ccm_web/client/tsconfig.json
```diff
@@ -1,26 +1,15 @@
 {
+  "extends": "../tsconfig.json",
```
pushyamig, line 2: does this means this is extending all the rules from the root `tsconfig.json`?

ssciolla: Yes, that is my understanding. Any repeated properties in this `tsconfig.json` will override what's in the base config.

https://www.typescriptlang.org/tsconfig#extends

pushyamig:
The new changes looks good. I don't see anything obvious from stopping this from merging. We can keep on improving the set as we go.

ccm_web/server/app.ts
```diff
@@ -0,0 +1,59 @@
...
+  app.use(express.static(prodBuildPath))
+
+  app.get('*', (req, res) => {
+    res.sendFile(path.join(prodBuildPath, 'index.html'))
+  })
+}
+
+// Start the server
+app.listen(port, () => {
+  console.log(`server started at http://localhost:${port}`)
```
pushyamig, line 58: This log message can be tailored saying Dev or proc env in Launch

ccm_web/package.json
ssciolla: Done, see `README.md`.

ccm_web/server/app.ts
ssciolla: See issue #29.

ccm_web/server/app.ts
ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/24/commits/858461ef2c79e99ce303bc6d937346f315d091d6

ssciolla:
@pushyamig , is this enough for now? We'll need to handle the production host use `app.listen`, but we're not ready to test or implement that. We can create an issue.

ssciolla:
Thanks for the review, @pushyamig, going to squash and merge now.

PR #35, opened 2021-03-16
First pass at integrating Material  UI

chrisrrowland:
Added Material UI dependency.

Created a skeleton for a landing page.

No theme created, no navigation added.   Just something for people to check out if its a reasonable starting point.

### DIFF
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 
 import ConsumerTest from './components/ConsumerTest'
-import Home from './components/Home'
+import Home from './pages/Home'
 import './App.css'
 
 function App (): JSX.Element {
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -1,26 +1,11 @@
 import React from 'react'
+import Home from './components/Home'
 import './App.css'
-
-import ConsumerTest from './components/ConsumerTest'
 
 function App (): JSX.Element {
   return (
     <div className='App'>
-      <header className='App-header'>
-        <p>
-          Edit <code>src/App.tsx</code>, and refresh browser to reload!
-        </p>
-        <a
-          className='App-link'
-          href='https://reactjs.org'
-          target='_blank'
-          rel='noopener noreferrer'
-        >
-          Learn React
-        </a>
-        <br />
-        <ConsumerTest />
-      </header>
+      <Home/>
     </div>
   )
 }
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -1,4 +1,6 @@
 import React from 'react'
+
+import ConsumerTest from './components/ConsumerTest'
 import Home from './components/Home'
 import './App.css'
 
@@ -6,6 +8,7 @@ function App (): JSX.Element {
   return (
     <div className='App'>
       <Home/>
+      <ConsumerTest/>
     </div>
   )
 }
--- a/ccm_web/client/src/components/FeatureCard.tsx
+++ b/ccm_web/client/src/components/FeatureCard.tsx
@@ -1,0 +1,72 @@
+import React from 'react'
+import { makeStyles } from '@material-ui/core/styles'
+import Card from '@material-ui/core/Card'
+// import CardActions from '@material-ui/core/CardActions'
+import CardContent from '@material-ui/core/CardContent'
+import Grid from '@material-ui/core/Grid'
+import Typography from '@material-ui/core/Typography'
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    backgroundColor: '#FAFAFA'
+  },
+  centered: {
+    width: '100%'
+  },
+  title: {
+    fontSize: 14
+  },
+  pos: {
+    // marginBottom: 12
+  },
+  tall: {
+    height: '400px'
+  },
+  short: {
+    height: '188px',
+    padding: '0px'
+  }
+}))
+
+function FeatureCard (props: FeatureCardProps): JSX.Element {
+  const classes = useStyles()
+
+  function sizeClass (size: string): string {
+    if (size === 'tall') {
+      return classes.tall
+    } else {
+      return classes.short
+    }
+  }
+
+  return (
+    <Card className={`${classes.root} ${sizeClass(props.size)}`} variant="outlined">
+        <CardContent>
+            <Grid container xs={12}>
+                <div className={classes.centered}>
+                    {props.icon}
+                    {/* {React.cloneElement(props.icon, { fontSize: 'large' })} */}
+                </div>
+                <div className={classes.centered}>
+                    <Typography className={classes.title} color="textPrimary" gutterBottom>
+                    {props.title}
+                    </Typography>
+                    <Typography variant="body2" component="p" color="textSecondary" >
+                    {props.description}
+                    </Typography>
+                </div>
+            </Grid>
+        </CardContent>
+    </Card>
+  )
+}
+
+interface FeatureCardProps {
+  title: string
+  description: string
+  icon: JSX.Element
+  size: 'tall' | 'short'
+}
+
+export { FeatureCard as default }
+export type { FeatureCardProps }
--- a/ccm_web/client/src/components/FeatureCard.tsx
+++ b/ccm_web/client/src/components/FeatureCard.tsx
@@ -17,7 +17,6 @@ const useStyles = makeStyles((theme) => ({
   title: {
     fontSize: 14
   }
-
 }))
 
 function FeatureCard (props: FeatureCardProps): JSX.Element {
--- a/ccm_web/client/src/components/FeatureCard.tsx
+++ b/ccm_web/client/src/components/FeatureCard.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import { makeStyles } from '@material-ui/core/styles'
 import Card from '@material-ui/core/Card'
-// import CardActions from '@material-ui/core/CardActions'
 import CardContent from '@material-ui/core/CardContent'
 import Grid from '@material-ui/core/Grid'
 import Typography from '@material-ui/core/Typography'
--- a/ccm_web/client/src/components/FeatureCard.tsx
+++ b/ccm_web/client/src/components/FeatureCard.tsx
@@ -25,23 +25,23 @@ function FeatureCard (props: FeatureCardProps): JSX.Element {
 
   return (
     <Card className={`${classes.root}`} variant="outlined">
-        <CardContent>
-            <Grid container>
-                <Grid item xs={12}>
-                    <div className={classes.centered}>
-                        {props.icon}
-                    </div>
-                    <div className={classes.centered}>
-                        <Typography className={classes.title} color="textPrimary" gutterBottom>
-                        {props.title}
-                        </Typography>
-                        <Typography variant="body2" component="p" color="textSecondary" >
-                        {props.description}
-                        </Typography>
-                    </div>
-                </Grid>
-            </Grid>
-        </CardContent>
+      <CardContent>
+        <Grid container>
+          <Grid item xs={12}>
+            <div className={classes.centered}>
+              {props.icon}
+            </div>
+            <div className={classes.centered}>
+              <Typography className={classes.title} color="textPrimary" gutterBottom>
+                {props.title}
+              </Typography>
+              <Typography variant="body2" component="p" color="textSecondary" >
+                {props.description}
+              </Typography>
+            </div>
+          </Grid>
+        </Grid>
+      </CardContent>
     </Card>
   )
 }
--- a/ccm_web/client/src/components/FeatureCard.tsx
+++ b/ccm_web/client/src/components/FeatureCard.tsx
@@ -47,6 +47,7 @@ function FeatureCard (props: FeatureCardProps): JSX.Element {
 }
 
 interface FeatureCardProps {
+  id: string
   title: string
   description: string
   icon: JSX.Element
--- a/ccm_web/client/src/components/FeatureCard.tsx
+++ b/ccm_web/client/src/components/FeatureCard.tsx
@@ -8,53 +8,38 @@ import Typography from '@material-ui/core/Typography'
 
 const useStyles = makeStyles((theme) => ({
   root: {
-    backgroundColor: '#FAFAFA'
+    backgroundColor: '#FAFAFA',
+    height: 200
   },
   centered: {
     width: '100%'
   },
   title: {
     fontSize: 14
-  },
-  pos: {
-    // marginBottom: 12
-  },
-  tall: {
-    height: '400px'
-  },
-  short: {
-    height: '188px',
-    padding: '0px'
   }
+
 }))
 
 function FeatureCard (props: FeatureCardProps): JSX.Element {
   const classes = useStyles()
 
-  function sizeClass (size: string): string {
-    if (size === 'tall') {
-      return classes.tall
-    } else {
-      return classes.short
-    }
-  }
-
   return (
-    <Card className={`${classes.root} ${sizeClass(props.size)}`} variant="outlined">
+    <Card className={`${classes.root}`} variant="outlined">
         <CardContent>
-            <Grid container xs={12}>
-                <div className={classes.centered}>
-                    {props.icon}
-                    {/* {React.cloneElement(props.icon, { fontSize: 'large' })} */}
-                </div>
-                <div className={classes.centered}>
-                    <Typography className={classes.title} color="textPrimary" gutterBottom>
-                    {props.title}
-                    </Typography>
-                    <Typography variant="body2" component="p" color="textSecondary" >
-                    {props.description}
-                    </Typography>
-                </div>
+            <Grid container>
+                <Grid item xs={12}>
+                    <div className={classes.centered}>
+                        {props.icon}
+                    </div>
+                    <div className={classes.centered}>
+                        <Typography className={classes.title} color="textPrimary" gutterBottom>
+                        {props.title}
+                        </Typography>
+                        <Typography variant="body2" component="p" color="textSecondary" >
+                        {props.description}
+                        </Typography>
+                    </div>
+                </Grid>
             </Grid>
         </CardContent>
     </Card>
@@ -65,7 +50,7 @@ interface FeatureCardProps {
   title: string
   description: string
   icon: JSX.Element
-  size: 'tall' | 'short'
+  ordinalty: number
 }
 
 export { FeatureCard as default }
--- a/ccm_web/client/src/components/FeatureCard.tsx
+++ b/ccm_web/client/src/components/FeatureCard.tsx
@@ -24,7 +24,7 @@ function FeatureCard (props: FeatureCardProps): JSX.Element {
   const classes = useStyles()
 
   return (
-    <Card className={`${classes.root}`} variant="outlined">
+    <Card className={`${classes.root}`} variant="outlined" tabIndex={props.ordinalty}>
       <CardContent>
         <Grid container>
           <Grid item xs={12}>
--- a/ccm_web/client/src/components/Home.tsx
+++ b/ccm_web/client/src/components/Home.tsx
@@ -13,7 +13,8 @@ import PersonAddOutlinedIcon from '@material-ui/icons/PersonAddOutlined'
 const useStyles = makeStyles((theme) => ({
   root: {
     flexGrow: 1,
-    minWidth: 275
+    minWidth: 275,
+    padding: 25
   }
 }))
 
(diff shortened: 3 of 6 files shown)


ccm_web/client/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+const useStyles = makeStyles((theme) => ({
+  root: {
+    backgroundColor: '#FAFAFA'
+  },
+  centered: {
+    width: '100%'
+  },
+  title: {
+    fontSize: 14
+  },
```
ssciolla, line 18: Yeah, some of this should be in a `theme` or the root CSS file. But you already know that!

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+    icon: <LibraryBooksOutlinedIcon fontSize='large'/>,
+    size: 'short'
+  }
+
+  const addSectionsGroupsProps: FeatureCardProps = {
+    title: 'Add Sections / Groups',
+    description: 'Add sections, groups through csv viles into your own course',
+    icon: <AccountCircleOutlinedIcon fontSize='large'/>,
+    size: 'tall'
+  }
```
ssciolla, line 42: I think we will reach a point where some of this data is coming from the backend. To prepare for that, maybe we want to just offload this into a JSON file, figure out how to map data we can expect to come from the backend with the front-end stuff (mostly the icons, I guess).

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+                        <Grid item xs={12} sm={6}>
+                        <FeatureCard {...addNonUMUsersProps}/>
+                        </Grid>
+                    </Grid>
+                </Grid>
+                <Grid item xs={12} sm={2}>
+                    <FeatureCard {...addSectionsGroupsProps}/>
+                </Grid>
+            </Grid>
+        </div>
```
ssciolla, line 81: Looks like this is using a different indentation (4 spaces?) and some things aren't indented enough. I wonder if there's more linting configuration needed for JSX...

ccm_web/client/src/components/Home.tsx
ssciolla: The backend doesn't need to know about icon choices, but perhaps we can just use a simple object to map task (still thinking about the entity name) names or IDs to their icons.

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+import PersonAddOutlinedIcon from '@material-ui/icons/PersonAddOutlined'
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    flexGrow: 1,
+    minWidth: 275
+  }
+}))
+
+function Home (): JSX.Element {
```
ssciolla, line 20: I'm wondering if we should make this component a `Page` or a container, rather than a simple component under `components`. In my limited experience, distinguishing between common, elemental components and higher-order components (that use other components and contain data fetching) seems to be a useful way to organize a React project.

ssciolla: If we go the page route, we can either create a `Layout` component later or use some framing/header/footer/navigation stuff in each separate page.

ssciolla (commented):
Some initial thoughts, kind of thinking into the future. No major concerns so far. I don't think everything I mentioned needs to be done for this to be merged, so use your discretion and ask me for another review when you think it's merge-able.

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const gradebookToolsProps: FeatureCardProps = {
+    title: 'Gradebook Tools',
+    description: 'Trim the gradebook from Canvas, or trim the gradebook from a third party to correct format',
+    icon: <LibraryBooksOutlinedIcon fontSize='large'/>,
+    size: 'short'
+  }
+
+  const addSectionsGroupsProps: FeatureCardProps = {
+    title: 'Add Sections / Groups',
```
pushyamig, line 38: creating Groups is not done in CCM anymore. It is now natively supported in canvas. You can remove the groups text and just keep the `Add Sections`

ccm_web/client/src/components/FeatureCard.tsx
pushyamig: I feel the Home outline is missing a Layout. As things needs to be centered. It should have a consistence margin on top/bottom/left/right.

ccm_web/package-lock.json
```diff
@@ -1,8 +1,7137 @@
 {
   "name": "ccm_web",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
pushyamig, line 4: @chrisrrowland was this file auto generated when you doing a docker build or you had to run the `npm install` manually on your machine?

ccm_web/client/src/components/Home.tsx
pushyamig: @ssciolla you mentioned some of these can come from the Backend, after LTI launch happens and based on the role are you going to send 
```
Person A, subaccount_admin, feature_list: all
Person B, instrutor, Feature-list: merge_sections, add_non_umich_user, Gradebook_tools
```

BE simple can send person role and UI can map it up to appropriate features. 

It can be done in either way, I am fine either way

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+    icon: <PersonAddIcon fontSize='small'/>,
+    size: 'short'
+  }
+
+  const addNonUMUsersProps: FeatureCardProps = {
+    title: 'Add Non-UM Users',
+    description: 'Enroll non-UM users to your available sections',
+    icon: <PersonAddOutlinedIcon fontSize='small'/>,
+    size: 'short'
+  }
```
pushyamig, line 56: Just a thought, can this entire section can be separated to a separate component and can be made as list

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+                            <FeatureCard {...addUMUsersProps}/>
+                        </Grid>
+                        <Grid item xs={12} sm={6}>
+                        <FeatureCard {...addNonUMUsersProps}/>
+                        </Grid>
+                    </Grid>
+                </Grid>
+                <Grid item xs={12} sm={2}>
+                    <FeatureCard {...addSectionsGroupsProps}/>
+                </Grid>
```
pushyamig, line 79: instead of separate `<grid>` we can use `map` operation to populate the list with single `Grid` element. This will eliminate if the styling of the grid needs to be change let say you need to add `lg` property then you don't need to do it on each `grid` element representing a feature. 

This is just a thought, you can also work on it when BE sends data and it more clear that way

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const gradebookToolsProps: FeatureCardProps = {
+    title: 'Gradebook Tools',
+    description: 'Trim the gradebook from Canvas, or trim the gradebook from a third party to correct format',
+    icon: <LibraryBooksOutlinedIcon fontSize='large'/>,
+    size: 'short'
+  }
+
+  const addSectionsGroupsProps: FeatureCardProps = {
+    title: 'Add Sections / Groups',
+    description: 'Add sections, groups through csv viles into your own course',
```
pushyamig, line 39:
```suggestion
    description: 'Add sections, groups through csv files into your own course',
```

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,85 @@
...
+    minWidth: 275
+  }
+}))
+
+function Home (): JSX.Element {
+  const classes = useStyles()
+
+  const mergeSectionProps: FeatureCardProps = {
+    title: 'Merge Sections',
+    description: 'Combine sections taught by yourself or others into one Canvas site for easier management',
```
pushyamig, line 25: I think this description needs to change since instructor can only merge his sections he teaches but admins can merge across different courses. We don't want description to be different for different role types.

How about `Combine sections into one Canvas site for easier management`

ccm_web/client/src/components/Home.tsx
chrisrrowland: @pushyamig that was my thought of how it would work, with the UI determining what to display based on the roles it is passed.  That seems like a simple and logical and probably practical implementation to me.  

The text/icons coming from the backend seems like it could be a useful idea if that sort of thing was meant to be configurable or changing rapidly, or perhaps changing between releases.  I'm not sure  any of those things will be the case, though.

For now I just plan to leave this as is b/c I figure it's going to end up changing when we figure out how this is going to work.

ccm_web/client/src/components/Home.tsx
chrisrrowland: What is  the  issue here?  I've noticed that the linter complains sometime about mismatched indentation, but I've always fixed it when it complains.

Do you know if it's expecting tabs or spaces?

ccm_web/client/src/components/Home.tsx
ssciolla: Yeah, that might be okay, @pushyamig, need to think about it a little more. There still might be advantages to keeping some of this data in a static file, rather than in a React component.

ccm_web/package-lock.json
ssciolla: I don't think a `package-lock.json` change in the container would be propagated to the repo through any volume mapping, so must have been done locally?

ccm_web/client/src/components/Home.tsx
ssciolla: Yeah, I'd prefer the use of `map` with an array, but that's not exactly feasible unless we move away from the wireframe landing page design (which he was trying to replicate here).

ccm_web/package-lock.json
pushyamig: @ssciolla I think the way the Docker file is setup is it is taking both package.json and package-lock.json and when build happens it will always go by the lock file dependency tree. We might miss the getting any latest updates on package. 
How do the package update happens automatically?

chrisrrowland: package-lock.json already existed.  I added material-ui dependencies which accounted for most of the changes.  I believe I also did an npm install on eslint when I didn't realize it was already installed, resulting in an upgrade there.
Typescript also got an upgrade, but I don't recall why.  I don't think it hurts though.

ccm_web/client/src/components/Home.tsx
chrisrrowland: The layout is currently kind of funky b/c of the wireframe with 40/40/20 widths and the half height boxes in the center.  If they are all made the same it will be easy to do what you suggest.

ccm_web/client/src/components/Home.tsx
ssciolla: Indentation seems to be 4 spaces instead of 2 (which is enforced in the TS, I believe). I'd prefer to see that in the JSX as well? I also see at least one `FeatureCard` that needs to be indented again.

ccm_web/client/src/components/Home.tsx
chrisrrowland: Sure, I just copied from the design doc.

ccm_web/client/src/components/Home.tsx
ssciolla: The `standard` rule (first here: https://standardjs.com/rules.html#rules) is 2 spaces, if that helps.

ccm_web/package-lock.json
pushyamig: Yeah.. the process is same with other project. 
But does it make sense to automate the process.
1. During the docker build don't add the lock file during the npm install. This will get all the latest package updates
2. This creates a new lock file and once the build is done due to volume mapping the new lock file is updated or mapped local.
3. This way we will always end up will latest dependencies
4. This is just a proposal, no work is needed at this moment.

ccm_web/client/src/components/Home.tsx
pushyamig: Go for it,  make all the cards of same size

ccm_web/package-lock.json
ssciolla: Yeah, I need to think about this more, and may want to discuss. I kind of like the current approach.

ccm_web/client/src/components/Home.tsx
pushyamig: @ssciolla, In my IDE I configured to use the TypeScript Standard Style. I don't see MY IDE is  complaining with extra indentation. Am I missing anything?

Screen Shot 2021-03-17 at 10 15 45 AM

ssciolla: You are not missing anything. I am just wondering why the linting configuration is not complaining about the extra indentation. This isn't anyone's fault (except maybe mine), it's probably something that needs to be changed in `.eslintrc.js`.

ccm_web/client/src/App.tsx
```diff
@@ -1,26 +1,11 @@
...
-        <a
-          className='App-link'
-          href='https://reactjs.org'
-          target='_blank'
-          rel='noopener noreferrer'
-        >
-          Learn React
-        </a>
-        <br />
-        <ConsumerTest />
```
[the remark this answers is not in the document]
chrisrrowland: Added this back.

ccm_web/client/src/components/Home.tsx
chrisrrowland: I think my editor is trying to use tabs by default.. my manual fix when the linter was complaining has been to remove the tab and put 2 spaces.  I guess  maybe its ending up with leading tabs and that's what you're seeing?
It's  not complaining on my end.  I'll see if i can figure out how to change vscode to use spaces.

chrisrrowland: Hmm it was using spaces instead of tabs, just 4 instead of 2... but i was fixing it wherever i saw a complaint so not sure what the issue is.

ccm_web/client/src/components/Home.tsx
chrisrrowland: @ssciolla So you want a containers folder like myla at the same level as components?

ccm_web/client/src/components/Home.tsx
ssciolla: Yeah, in VSCode, you can change that by clicking where says "Spaces" on the bottom of the screen.

ccm_web/client/src/components/Home.tsx
ssciolla: Yeah, at that level. Do you have strong preference between `containers` or `pages`? I have a slight preference for the `page` idea, I think it's a little less abstract than `container`.

ccm_web/client/src/components/Home.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  const addNonUMUsersProps: FeatureCardProps = {
+    title: 'Add Non-UM Users',
+    description: 'Enroll non-UM users to your available sections',
+    icon: <PersonAddOutlinedIcon fontSize='small'/>,
+    ordinalty: 5
+  }
+
+  const cards: FeatureCardProps[] = [mergeSectionProps, gradebookToolsProps, addSectionsGroupsProps, addUMUsersProps, addNonUMUsersProps]
```
ssciolla, line 59: Not a big difference, but you could just define these all at once, rather than using variables. Also, `ordinality` idea is nice, but we may as well just give them IDs (e.g. `id: 0`).

ccm_web/client/src/components/Home.tsx
chrisrrowland: Pages makes more sense to me. I don't know how a component is really all that different than a 'container' anyways

ccm_web/package-lock.json
chrisrrowland: The purpose of the package-lock file is so you don't get unintended upgrades, right?
I don't see what is not working with the current setup.

ccm_web/client/src/components/Home.tsx
chrisrrowland: I think we wait to see how this is going to be integrated with the backend/role information for now.

There is little to the Home component besides this information as it is.

ccm_web/client/src/components/Home.tsx
chrisrrowland: I don't like the idea of ordinality being tied to something not intended for that purpose, but they probably should have an id b/c i was using the title as the key.

ccm_web/client/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+    fontSize: 14
+  }
+
+}))
+
+function FeatureCard (props: FeatureCardProps): JSX.Element {
+  const classes = useStyles()
+
+  return (
+    <Card className={`${classes.root}`} variant="outlined">
```
ssciolla, line 27: The specific of this might change later when the link is added, but what do you think about adding `tabIndex={0}` just to make these keyboard accessible so we remember that's needed later?

ccm_web/client/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <CardContent>
+            <Grid container>
```
ssciolla, line 29: Screen Shot 2021-03-17 at 4 04 05 PM

Seems like maybe there's too much empty space beneath the description. Not exactly sure what's causing that. Also, these particular icons are pretty small, maybe they need to be `fontSize='large'` like the others?

ssciolla (approved):
I'm mostly fine with you merging as is, but I had a bunch more comments. This is good work. Feel free to resolve what you care about and merge; no need to ping me again for a review.

ssciolla:
@chrisrrowland, word to the wise, please try to use use an issue branch in the future. It might have become a headache if this wasn't merged before someone's work was.

ccm_web/client/src/pages/Home.tsx
```diff
@@ -0,0 +1,81 @@
...
+    id: 'GradebookTools',
+    title: 'Gradebook Tools',
+    description: 'Trim the gradebook from Canvas, or trim the gradebook from a third party to correct format',
+    icon: <LibraryBooksOutlinedIcon fontSize='large'/>,
+    ordinalty: 2
+  }
+
+  const addGroupsProps: FeatureCardProps = {
+    id: 'AddGroups',
+    title: 'Add Groups',
```
pushyamig, line 42: There is not Add groups feature, Please replace it with create sections

pushyamig, line 42:
```suggestion
    title: 'Create Sections',
```

ccm_web/client/src/pages/Home.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const gradebookToolsProps: FeatureCardProps = {
+    id: 'GradebookTools',
+    title: 'Gradebook Tools',
+    description: 'Trim the gradebook from Canvas, or trim the gradebook from a third party to correct format',
+    icon: <LibraryBooksOutlinedIcon fontSize='large'/>,
+    ordinalty: 2
+  }
+
+  const addGroupsProps: FeatureCardProps = {
+    id: 'AddGroups',
```
pushyamig, line 41:
```suggestion
    id: 'CreateSections',
```

ccm_web/client/src/pages/Home.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const gradebookToolsProps: FeatureCardProps = {
+    id: 'GradebookTools',
+    title: 'Gradebook Tools',
+    description: 'Trim the gradebook from Canvas, or trim the gradebook from a third party to correct format',
+    icon: <LibraryBooksOutlinedIcon fontSize='large'/>,
+    ordinalty: 2
+  }
+
+  const addGroupsProps: FeatureCardProps = {
```
pushyamig, line 40:
```suggestion
  const createSectionsProps: FeatureCardProps = {
```

ccm_web/client/src/pages/Home.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const addNonUMUsersProps: FeatureCardProps = {
+    id: 'addNonUMUsers',
+    title: 'Add Non-UM Users',
+    description: 'Enroll non-UM users to your available sections',
+    icon: <PersonAddOutlinedIcon fontSize='small'/>,
+    ordinalty: 5
+  }
+
+  const cards: FeatureCardProps[] = [mergeSectionProps, gradebookToolsProps, addGroupsProps, addUMUsersProps, addNonUMUsersProps]
```
pushyamig, line 64:
```suggestion
  const cards: FeatureCardProps[] = [mergeSectionProps, gradebookToolsProps, createSectionsProps, addUMUsersProps, addNonUMUsersProps]
```

ccm_web/package-lock.json
ssciolla: I agree. Anyways, let's discuss separately, nothing you need to do here.

pushyamig:
I forgot to mention, can you add a title to the page 'Canvas course manager`

ccm_web/client/src/components/Home.tsx
ssciolla: I'm fine waiting as well, this can be refactored later.

ccm_web/client/src/components/Home.tsx
ssciolla: I see your point. Thanks for adding the `id`.

ssciolla:
> I forgot to mention, can you add the text 'Canvas course manager` on the page

This might be part of the layout/navigation discussion. We could put it somewhere now as an `h1`, but it might change.

pushyamig:
> This might be part of the layout/navigation discussion. We could put it somewhere now as an h1, but it might change.

ok

ccm_web/client/src/pages/Home.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <Grid container spacing={3}>
+                {cards.sort((a, b) => (a.ordinalty < b.ordinalty) ? -1 : 1).map(p => {
+                  return (
+                    <Grid key={p.id} item xs={12} sm={4}>
+                        <FeatureCard {...p}/>
+                    </Grid>
+                  )
+                })}
+            </Grid>
+        </div>
```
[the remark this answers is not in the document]
chrisrrowland: I had adjusted the default indentation but I hadn't reformatted after doing it since the linter wasn't complaining.. made that change  now.

ccm_web/client/src/components/FeatureCard.tsx
chrisrrowland: Yeah the font size is wrong, i forgot to change it from the original smaller boxes.

I'm not particularly focused on the fine details of how the css looks right now, but I can make them smaller..  I feel like we might be focusing too much on details though

ccm_web/client/src/pages/Home.tsx
chrisrrowland: Apparently I heard this backwards 🙃

ccm_web/client/src/components/FeatureCard.tsx
chrisrrowland: Good idea, thanks

chrisrrowland: Added `tabIndex={props.ordinalty}`

ccm_web/client/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import { makeStyles } from '@material-ui/core/styles'
+import Card from '@material-ui/core/Card'
+// import CardActions from '@material-ui/core/CardActions'
```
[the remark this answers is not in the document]
chrisrrowland: I hear ya, seems like this is  inevitably going to be uncommented when it's made clickable so I just left it.  Maybe it bothers you more that I left it in than it would bother me to delete it and add it back in lol

ccm_web/client/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+interface FeatureCardProps {
+  title: string
+  description: string
+  icon: JSX.Element
+  ordinalty: number
+}
+
+export { FeatureCard as default }
+export type { FeatureCardProps }
```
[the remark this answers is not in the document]
chrisrrowland: Frankly, I don't know and didn't think about it that much.  Would it be better to not have as default?

ccm_web/client/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import { makeStyles } from '@material-ui/core/styles'
+import Card from '@material-ui/core/Card'
+// import CardActions from '@material-ui/core/CardActions'
+import CardContent from '@material-ui/core/CardContent'
+import Grid from '@material-ui/core/Grid'
+import Typography from '@material-ui/core/Typography'
```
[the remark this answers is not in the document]
chrisrrowland: I don't know how that works, maybe you can brief me sometime

ccm_web/client/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+                    </div>
+                </Grid>
+            </Grid>
+        </CardContent>
+    </Card>
+  )
+}
+
+interface FeatureCardProps {
+  id: string
```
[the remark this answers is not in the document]
chrisrrowland: Is there a reason it should be a number rather than a string?  I thought about that but it seemed like a string would have an more implicit meaning.  I guess it takes up a few more bytes?

pushyamig (approved):

ccm_web/client/src/components/FeatureCard.tsx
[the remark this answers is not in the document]
ssciolla: This is probably open to discussion. Efficiency probably doesn't matter here as much. In my understanding, semantic meaning in a key is actually not desirable, because if it changes then the key might change.

ccm_web/client/src/components/FeatureCard.tsx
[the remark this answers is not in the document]
ssciolla: It's not a big deal either way. There may perks to doing it your way, you're kinda using `default` to indicate what the core `export` is.

Issue #38, opened 2021-03-16
Adopt a logging approach for server


ssciolla:
Thus far we've been mainly using `console.log` and `console.error`, but there are likely better ways to approach this so that we can include timestamps and log origin, as well have access to more levels like `debug`.

[`Log4js`](https://www.npmjs.com/package/log4js) and [`winston`](https://www.npmjs.com/package/winston) seem like popular options with TypeScript support.

In addition, those of us from Django backgrounds will miss request logging. There is express middleware for this; [morgan](https://www.npmjs.com/package/morgan) seems like a popular option.

pushyamig:
Test passes. I don't see any console log message in the server folder of ccm_web and a configuration of log level is established as well

PR #40, opened 2021-03-17
Add LTI Integration using ltijs (#23)

### LINKED ISSUE #23: Building Block: LTI 1.3 Integration
This issue is mainly have the LTI integrated into the CCM. [LTIJS](https://github.com/Cvmcosta/ltijs) library seems to be promising library as analyzed here #4 

CCM tool would be added at course level, more info is [here](https://canvas.instructure.com/doc/api/file.navigation_tools.html#course_navigation). 

The LTI 1.3 configuration should be only visible to admins in the course. 
Configuration of LTI tool 
Visibility: admins
Default: visible  // we can disable the tool and give an option to the user to enable it using the Setting => Navigation
$Canvas.membership.roles: Give exact users role in Canvas like Sub-Account Admin, Support Consultant etc
$Canvas.user.isRootAccountAdmin: Optional can be used to find out if user root admin or not

LTI tool Launch access list is described [here](https://docs.google.com/spreadsheets/d/1pm5y9FX0zrDX7Zy3mOyDLxmA-iKoWfmlxvbtNkWg6Zw/edit#gid=1360549473)

ssciolla:
This PR aims to resolve #23.

#### Features (Coming)
- Database service setting up PostgreSQL instance and application database
- Configuration approach and validation
- Separation of api routing
- Client handling of LTI key (JWT)
- `usePromise` from ROHQ
- `docker-compose-prod.yml` for production workflow testing

#### To Do
- Extend existing @types for `litjs` and write types for `litjs-sequelize`
- Handle production host name

####
Resources
* [ltijs](https://cvmcosta.me/ltijs/#/)
* [Typing with external libraries blogpost from Credera](https://www.credera.com/insights/typescript-adding-custom-type-definitions-for-existing-libraries/)


ssciolla:
Hi @pushyamig, @chrisrrowland, @jonespm, I am still working on a couple things related to this PR, but most of the work is additive (I might open separate issues for some of them). It would be great if people started trying it out and offering feedback. I will try to have some of the other pieces done soon.

docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
     environment:
       - PORT=4000
       - NODE_ENV=development
+    env_file: .env
```
chrisrrowland, line 15: Would make sense to have the .env file sit outside of the repo directory?
Just thinking about how it would be lost if you ever blew away your local repo.

chrisrrowland (commented):
Not really a lot to add  here.
Got it up and running locally with help from @ssciolla 
looks good so far

docker-compose.yml
jonespm: It looks like nestjs is just using [dotenv](https://www.npmjs.com/package/dotenv) rather than a json configuration, but it loads the .env directly rather than having something like docker-compose populate the environment.

ssciolla: I tried `config` and thought about using JSON Schema library, but ultimately, I just decided to keep it simple for now and use environment variables. I'm validating them using TypeScript and building a nested `config` object here: https://github.com/ssciolla/canvas-course-manager-next/blob/issue-23-lti/ccm_web/server/config.ts There's probably a way a better way to do this, but there are some advantages to the simplicity of environment variables and not having to worry about a `dotenv`, `config`, package or whatever. ROHQ uses a similar pattern, so I was partly inspired by that.

pushyamig:
I am reviewing it now

ccm_web/client/src/api.ts
```diff
@@ -0,0 +1,21 @@
+import getLTIKey from './utils/getLTIKey'
```
pushyamig, line 1: is this file a collection of all api calls for mergeSection, create sections etc on client side? This is a good organization

ccm_web/client/src/models.ts
```diff
@@ -0,0 +1,3 @@
+export interface HelloMessageData {
+  message: string
```
pushyamig, line 2: what is purpose of the file? and how it will be used for future? A short comment would be use full.

ccm_web/client/src/api.ts
ssciolla: Yep, that's the idea!

pushyamig: few comments on file usefulness/purpose will be helpful?

ccm_web/client/src/models.ts
ssciolla: I'll add a comment. `models` (it could be called something else) would be used to define interfaces for common entity types/objects in the application. For instance, in the future we could define `Section`, `Course`, or `Feature` objects that could be used throughout the application. This will hopefully reduce some duplication, promote common patterns.

ccm_web/server/config.ts
```diff
@@ -0,0 +1,75 @@
...
+  let server
+  let lti
+  let db
+
+  try {
+    server = {
+      port: validate<number>('PORT', Number(env.PORT), isNumber, 4000)
+    }
+    lti = {
+      encryptionKey: validate<string>('LTI_ENCRYPTION_KEY', env.LTI_ENCRYPTION_KEY, isString, 'LTIKEY'),
```
pushyamig, line 55: Some how the `LTI_ENCRYPTION_KEY` is commented out in the .env.sample. But regardless if not present looks like you are setting the some string here. I assume that is needed for Prod and non-prod env. Can we generate some random encrypted like we do in Django? so that way we can eliminate messages like below
`Value undefined for LTI_ENCRYPTION_KEY is invalid; using fallback LTIKEY`

pushyamig:
> Client handling of LTI key (JWT)
What do you mean by it? I suppose this is talking about idtoken? What is reason for client handling it. As of now I cannot pin point where exactly this is happening?

ccm_web/server/config.ts
ssciolla: Yes, we will want to do that (use a better encryption key). I think if you un-comment the line in your `.env` and change it, it should work properly. It gets passed in to the `ltijs` configuration.

ccm_web/client/src/models.ts
pushyamig: This file seems to be synonymous to `Config.ts`, isn't it? If Yes the client and server can have same names like `ConfigClient.ts` and `ConfigServer.ts`. If No adding a comment would help

ssciolla:
> > Client handling of LTI key (JWT)
> 
> What do you mean by it? I suppose this is talking about idtoken. What is reason for client handling it. As of now I cannot pin point where exactly this is happening

You can see how this is being used in `client/api.ts` and `client/utils/getLTIKey`. The JWT or LTIKey as `ltijs` calls it is being stored on the `window` after a successful launch; you need this key in your request headers to the server API, otherwise you'll be unauthorized. I'm actually working on improving this a bit so that client-side routing doesn't disrupt this, so the exact implementation may change.

ccm_web/client/src/models.ts
ssciolla: Mmm, not exactly, although the `models.ts` might define interfaces for some configuration data. This will also include interfaces for data originating from the Canvas API (e.g. `Section`) and other places in our application (like a database table).

pushyamig:
>  you need this key in your request headers to the server API, otherwise you'll be unauthorized. 

Ok, I See what's going on. But can we store is variable somewhere in the Clientside Cache or as a react drill down variables? 
If I understand this as you keep moving from page to page using react routes this `idk` will be removed from the browser. Any API call further cannot access this from `idk`

ssciolla:
> > you need this key in your request headers to the server API, otherwise you'll be unauthorized.
> 
> Ok, I See what's going on. But can we store is variable somewhere in the Clientside Cache or as a react drill down variables?
> If I understand this as you keep moving from page to page using react routes this `idk` will be removed from the browser. Any API call further cannot access this from `idk`

Yep, that's what I'm working on. We can start passing it as a `prop` to where it's needed, after grabbing it once on launch at the root of the application. Some kind of client-side store might be useful, but not sure if it's necessary yet.

pushyamig:
> We can start passing it as a prop to where it's needed, after grabbing it once on launch at the root of the application.
Just some suggestion, you can send the `idk` when send global props like user role. So that way client can access this always.

To be this `idk` sound sort of CSRF token, to keep track of if session is coming from same user. I believe the LTIJS workflow is binding `idk` that way. 
But I wish there is a better way ExpressJS can automatically handle this. But I feel this is OK. Let me see why this `idk` flow is built in this way. Just for my understanding.

docker-compose.yml
```diff
@@ -21,3 +22,20 @@ services:
       - ./ccm_web/dist:/base/dist:z
     command: npm run dev
     container_name: ccm_web
+    depends_on:
+      - database
+
+  database:
+    image: postgres:10.4-alpine
```
pushyamig, line 29: What is the reason to chose 10.4 version when [latest](https://www.postgresql.org/about/news/postgresql-13-released-2077/) is 13?

ssciolla: I think I copied this from ROHQ. I assumed that's what the MiDatabase team is offering. Is there somewhere we can check this?

pushyamig: I can check with them

pushyamig: So MiDatabase team offer Postgres in AWS (Cloud) through the ITS Container Service.https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.DBVersions.
@ssciolla We can choose any version as long as it's available on AWS. Let's go with latest??

ssciolla: Okay, I'll update it to a reasonable latest version.

ccm_web/client/src/hooks/useGlobals.ts
```diff
@@ -0,0 +1,44 @@
...
+    } else if (
+      globals.environment === 'development' &&
+      ltiKey === undefined &&
+      !globals.useLTI
+    ) {
+      isAuthenticated = true
+    }
+  } else if (getGlobalsError !== undefined) {
+    isAuthenticated = false
+  }
```
ssciolla, line 39: I don't actually know how good this `useGlobals` approach is, but I'll attempt to explain it. 

First, for some reason, `ltijs` is not locking down the application route if you don't have an LTI Key (it is for the other routes, so not sure what's going on there. Maybe it needs to be accessible for the launch?). So we need some way of locking down the landing page to unauthenticated users. Hence, the `/api/globals` call. MyLA's behavior is kinda similar, but it uses `globals` from the `index` template.

I would like to allow developers to still use the application without LTI, but I'm not sure adding an alternate authentication scheme is necessary. To that end, I'm trying to use some `Globals` data to allow users to be seen as authenticated if it's a development environment (they'll be able to access Globals data because on that path, the `api` routes will be unprotected). I don't think this is risky from a security perspective, but I could be missing something.

pushyamig: > I would like to allow developers to still use the application without LTI, 

With out the LTI how will you know what course, who the user is, what is user role? Based on that we are showing the feature in the landing page. If he has to merge sections then how will we know what courses to merge and from which term. We get the data from Launch information and allow him to do necessary feature action

ssciolla: I understand that, and perhaps we can only take this idea so far.

One idea is that we could allow developers to configure a course and username (that actually exists), which would be sent over through `Globals` (this would be in a future PR). In the case of a real LTI integration, the `Globals` object could include the real user, course, and roles from the LTI launch.

chrisrrowland: My thought, which may or may not make sense here, would be an interface that the contains all the relevant properties that lti provides.  We could then have something like an LTIIdentityProfvider when using LTI and a EnvVarIdentityProvider or ConfigFileIdentityProfier or something when using development (or testing?) if desired.  Perhaps someone else might want some other sort of IdentityProvider to integrate with their platform (unlikely)

Not even sure if Identity is the correct term here, I'm not sure what LTI is providing.

pushyamig: @ssciolla I appreciate your enthusiasm for providing an alternative approach without LTI. But I feel we should park that idea for now - my 2 cents.

ssciolla: Okay, I can rewind some of this and adjust `AppHandler`.

ssciolla: @chrisrrowland, something like you what you are describing is what I had mind, but I guess we can postpone or make it a separate issue. In the meantime, if developing the UI under LTI constraints bothers you enough, we can work on getting `storybook` set up.

chrisrrowland: @ssciolla It doesn't bother me.  It adds a minor level of annoyance but I can survive.

That said I understand wanting to shelf that as a nice to have but It doesn't seem like it really adds much effort to create the interface and the LTI implementation and leave any alternatives for a later effort.  

I am not in the weeds on this, though, so it could certainly be  more difficult than I'm envisioning lol

ssciolla: Thanks for the additional thoughts, @chrisrrowland. I want to think about this a little more and separate it from this PR. I'm not really sure about my current approach; it would be better if the non-LTI alternative also provided some kind of key for authenticating to API endpoints. That definitely would make this more complicated (how much so, not sure). If any of my current approach is use-able in the future, it'll just be a copy and paste job from these commits.

ssciolla: @pushyamig, see https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/40/commits/464fd0ccf98c3af0ad0112b2aa6e90c6d1e3841b

I would still like to know if others think my approach to locking down `/` makes sense. See `useGlobals` and `App.tsx`.

ccm_web/client/src/hooks/useGlobals.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const getGlobals = async (): Promise<Globals> => await api.getGlobals(ltiKey)
+  const [doGetGlobals, getGlobalsLoading, getGlobalsError] = usePromise(
+    getGlobals,
+    (value: Globals) => setGlobals(value)
+  )
+  useEffect(() => {
+    /* eslint-disable @typescript-eslint/no-floating-promises */
+    doGetGlobals()
+  }, [])
```
pushyamig, line 24: is this useEffect is making sure that the globals is fetched just once?

ssciolla: Yes.

ccm_web/client/src/hooks/useGlobals.ts
pushyamig: The look down approach look ok. I don't see any concerns there. 
I am still trying to wrap my head around the promise/async/await with fetch. This looks bit complicated just to drilling downs what's happening in the `usePromise`. But at the high level i understand what it is doing

ssciolla: Yeah, I don't have a mastery of it either. `async`, `await,` and `Promise` are all still pretty confusing to me; I plan to try to learn more about them as time allows.

ccm_web/server/appHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+        serverAddon: serverAddon,
+        staticPath: staticPath
+      }
+    )
+
+    provider.app.use('/api', this.apiRouter)
+
+    // Set lti launch callback
+    // When receiving successful LTI launch redirects to app.
+    provider.onConnect((token: any, req: Request, res: Response) => {
```
pushyamig, line 114: As part of LTI launch you should receive an `idtoken`. is this `token` hold that info. I am trying to console.log this but it dumps out something else. 
I don't see you are extracting the Launch variables like role, course, terms, user info. IS that something you are still working on?

ssciolla: Yeah, I know we need to do this, but I wasn't sure the best way to go about it yet. I think some of the information is already stored in the database, or it could be extracted from the JWT sent to the client. Yeah, there's something weird about logging the token here I've seen, I'm not sure exactly what's happening. I'll investigate and report back. We may want to split handling that info out into a separate issue.

ccm_web/client/src/hooks/useGlobals.ts
ssciolla: Marking this thread as resolved. Ask more questions if you see other issues related to this stuff.

ccm_web/server/appHandler.ts
pushyamig: Yeah.. it is storing it in the different tables. 
`select * From idtokens; select * from accesstokens;select * From contexttokens;`
LTI JS is implementing this storing the idtoken in different ways. It is not in one place but separate tables.

Either you have to access the `idtoken` when request comes to `app.ts` but I am assuming LTIJS is handling it before it is available in request.

Understanding how the access the id_token content is important. Even if it stores in DB we need to store it server cache or session. since we don't want to keep on querying the DB when say merge feature required data than BE needs to aware if he is an instructor or service admin

ssciolla: For now, how do you feel about just pulling the custom fields (username, roles, course) you want from the `contexttokens` table and sending them up with `/api/globals`? If we pass them down the component tree, they'll be available. Later we could also look at something like `react-redux`: https://react-redux.js.org/introduction/getting-started

pushyamig: I don't yet feel there is a need to redux. This is my thinking...
The Frontend needs to know what is the user role for showing the appropriate content. And based on the request the backend will server up the content. Sending the details of course/term/user I feel could be kept BE. BE need to be aware of user and use the appropriate scoped token to serve up the content.

 With the scoped token approach the token needs to be stored in DB once generated. but should be loaded in from DB and stored in session/cache so that further API call it doesn't need a DB lookup.

ssciolla: Mmm, I see. Can we push this off to a separate issue / can you create such an issue and assign it to me? I'd like to do some data modeling for user info, decide what ORM/ database client to use (probably `sequelize` to minimize dependencies) and think more about this caching approach. I don't think processing this information is a required part of getting LTI working (related but not necessarily in scope for the issue).

pushyamig: OK, This can be done as a separate issue.

pushyamig: w.r.t to Redux or some Client caching. If you look at the mockup we show lot of sections in the courses like merge sections, add_non_um_user, 3rd party gradebook trimmer. We can cache sections info. between view. But practically it would be a rare case that a instructor merges a sections and add non-umich user on the same day to keep that in cache.

We can very well make the API call for each case. but later if we see redundant patterns like this we could improve. 

Some thoughts....

ccm_web/server/localTypes/ltijs/index.d.ts
```diff
@@ -0,0 +1,274 @@
+/*
+Minimum temporary types for ltijs, building upon existing @types/ltijs and preliminary updates by the same author
+See https://github.com/paulschwoerer/DefinitelyTyped/commit/9e9bec14b02e3c78051c07862d2c305a7ac98c39
+This file does not include types for service classes (Grade, DeepLinking, NamesAndRoles, DynamicRegistration).
+
+References
+- https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html
+- https://stackoverflow.com/questions/45099605/ambient-declaration-with-an-imported-type-in-typescript
+*/
```
ssciolla, line 9: I've continued to wrestle with the `ltijs` types here, but this is where I'm at for now. Instead of overriding or patching, I'm basically just including all the types I think we need for now (we may be even be able to manage with less). Things seem to be working. I will see if I can open a PR with a few changes either against the WIP version or DefinitelyTyped.

Does it seem okay to copy-and-paste and update these types here? Not sure if that's in violation of copyright or whatnot. The license for DT is MIT, though the original type author may hold some rights.

ccm_web/server/app.ts
```diff
@@ -1,63 +1,21 @@
...
-  })
-}
-
-// Start the server
-// TO DO: handle production host
-app.listen(port, () => {
-  console.log(`Server started on ${host} and port ${port}`)
-})
+// TO DO: need to implement database availability check
+setTimeout(() => appHandler.startApp(), 10000)
```
ssciolla, line 21: This is basically a `sleep` to prevent the `ltijs` server from crashing if the `postgres` service isn't ready. 10 seconds is enough usually for `postgres` service to start up from scratch. If it's not from scratch (i.e. there's already a volume), it should only take a few seconds, and you can change this to `3000` or whatever.

We will want to implement a database availability check, either in a `start.sh` script or in a utility function. This feels like something that can be done later/separately, but let me know if you disagree. If we do it programmatically in JavaScript, we may want to use the ORM anyway (can be done at the same time as #46).

ccm_web/client/src/App.test.tsx
```diff
@@ -1,9 +1,8 @@
...
-test('renders learn react link', () => {
-  render(<App />)
-  const linkElement = screen.getByText(/learn react/i)
-  expect(linkElement).toBeInTheDocument()
-})
+// test('renders something', () => {
+//   render(<App />)
+//   const linkElement = screen.getByText(/something/i)
+//   expect(linkElement).toBeInTheDocument()
+// })
```
ssciolla, line 8: I'm not a huge fan of commented code, but this seems like a decent reminder to add some FE tests. We could also just delete the file and testing dependencies for now.

docker-compose.yml
ssciolla: I updated it to 13 for now, but see #49.

ccm_web/server/appHandler.ts
ssciolla: Thanks for creating the issues. Definitely need to think more about how much caching will be needed, but that's a later step.

ccm_web/server/localTypes/ltijs/index.d.ts
ssciolla: How is this? https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/40/commits/1117466cf171c582b0076d502d299fbbedd8c8e3

I also wonder if I need to include anything about code patterns taken from ROHQ, specifically `usePromise` and `handleErrors`. I asked James about `usePromise`, he said using it was fine, I didn't need to mention him, but that project carries an Apache license. I'm new to re-purposing code like this.

pushyamig: I feel Code share from ROHQ should be fine since it is internal project from ITS. Good you checked with ROHQ team members.

pushyamig:
this is an observation when the server side code change it restarts the server, I feel the server should not restart but just add the incremental changes on top of it. I am not saying this is good or bad but can the incremental update to server behavior be achieved without a restart? this is a question not asking for rework.

pushyamig:
>  2021-03-31 18:52:38.485 UTC [57] WARNING:  could not open statistics file "pg_stat_tmp/global.stat": Operation not permitted
I do see these warning, would like fix this in future. I can create issue.

pushyamig:
> (node:33) Warning: Accessing non-existent property 'MongoError' of module exports inside circular dependency

I prefer this warning to be removed as well, in future

ccm_web/server/app.ts
pushyamig: I feel that this is only for local development and not in non-prod or prod.SO I feel this timeout check needs if(dev) => setTimeout.
I feel setting a timeout is suffice if we encounter if this workaround is not working we can address it later.

ccm_web/server/appHandler.ts
```diff
@@ -0,0 +1,145 @@
...
+    const webpackDevMid = webpackDevMiddleware(compiler, {
+      publicPath: String(publicPath),
+      // Just copies what's used in memory to client/build (changing doesn't do anything)
+      writeToDisk: true
+    })
+    return { middleware: webpackDevMid, publicPath: String(publicPath) }
+  }
+
+  // ltijs docs: https://cvmcosta.me/ltijs/#/
+  async setupLTI (): Promise<Express> {
```
pushyamig, line 70: This Block of `setLTI` console.log doesn't work. even I enable `DEBUG=provider:*`. 
I wanted to know what is in side Token.
`provider.onConnect(async (token: IdToken, req: Request, res: Response) => {`

How did you determine the token is `IdToken`?

ssciolla: I don't know what's going on with this honestly. I will take a look. I believe it's `IdToken` based on the source code and the `@types` updates that the contributor proposed.

ssciolla:
> Warning: Accessing non-existent property 'MongoError' of module exports inside circular dependency

This error has something to do with a dependency of `ltijs`, so I don't think there's anything we can do about it ourselves. We'll have to wait for the maintainer to fix it or request a dependency update.

https://developer.mongodb.com/community/forums/t/warning-accessing-non-existent-property-mongoerror-of-module-exports-inside-circular-dependency/15411/6

ssciolla:
> > 2021-03-31 18:52:38.485 UTC [57] WARNING:  could not open statistics file "pg_stat_tmp/global.stat": Operation not permitted
> 
> I do see these warning, would like fix this in future. I can create issue.

We can fix this later if we stick with `postgres`.

ccm_web/server/app.ts
ssciolla: Mmm, that's a good point. If I add this, it will still break when using `docker-compose-prod.yml`...

ccm_web/server/appHandler.ts
ssciolla: So you can see the token logged if you use the production workflow offered through `docker-compose-prod.yml`. So maybe something `provider.redirect` is causing an issue here?

ssciolla: I now am wondering if this callback is getting disrupted by the Webpack Middleware. I'll focus on this tomorrow morning, try to dig into how this interacts with ltijs.

ccm_web/server/app.ts
pushyamig: ok, You can leave this for now and can deal later. But in reality we don't use docker-compose in prod this is more for testing purpose.

pushyamig:
> 2021-03-31 18:52:38.485 UTC [57] WARNING: could not open statistics file "pg_stat_tmp/global.stat": Operation not permitted
I do see these warning, would like fix this in future. I can create issue.
We can fix this later if we stick with postgres.

Sure

pushyamig:
> Warning: Accessing non-existent property 'MongoError' of module exports inside circular dependency
This error has something to do with a dependency of ltijs, so I don't think there's anything we can do about it ourselves. We'll have to wait for the maintainer to fix it or request a dependency update.
developer.mongodb.com/community/forums/t/warning-accessing-non-existent-property-mongoerror-of-module-exports-inside-circular-dependency/15411/6

Ok.
It's weird that we did not enable mongodb but still see the error. As you say probably in the configuration and LTIJS author might need to do something. We can open an issue with LTIJS library saying why this is warning is there for non-mongodb projects.

ssciolla:
> It's weird that we did not enable mongodb but still see the error. As you say probably in the configuration and LTIJS author might need to do something. We can open an issue with LTIJS library saying why this is warning is there for non-mongodb projects.

I believe this has been fixed in patch release of `ltijs` (see latest commit). The default database is `mongodb`; I agree it would be nice to have this not be a dependency if we're not using it, but I imagine that change is not a small one.

ccm_web/server/appHandler.ts
ssciolla: I'm fairly confident now the problem is the Webpack middleware, but I don't have a solution yet. 😞

pushyamig: Probably may be tied with how webpack is restarting the server when BE changes happened. How exactly is webpack interfering here?

pushyamig:
are we still sticking with postgres? I though there was a fix around Mysql DB issue we were seeing.

ssciolla:
> are we still sticking with postgres? I though there was a fix around Mysql DB issue we were seeing.

I was looking at the dependency updates earlier this week. In theory, yes, we should be able to switch to MySQL, but I couldn't get the updates to work for me. Right now I've pinned them to the previous minor versions. I will give it another shot now.

pushyamig:
It's ok, we can stick with Postgre, I don't want you keep going back and forth on this

ssciolla:
> It's ok, we can stick with Postgre, I don't want you keep going back and forth on this

We want to do it eventually, so I'll get started on it in another branch.

pushyamig:
The LTI serverless wiring seems to be working both in prod and development mode

- Both the FE and BE are updating as server and client file changes.
- Debugger attaches with nodemon change and keep remain attached after restart
- Webpack stuff seems to work as well
- the `static` folder routing for react dump don't seems to have a real location. Somehow it just work with `app.use('/static', express.static(this.envOptions.staticPath))` . Not an concern but curious
- API call all works as supposed to

The LTIJS in serverless mode seem to work, I will go in details Monday morning and give an approval.

docker-compose.yml
ssciolla: > Would make sense to have the .env file sit outside of the repo directory?
> Just thinking about how it would be lost if you ever blew away your local repo.

I would like to discuss the ideal configuration approach at a later date, if that's okay. I rarely need to wipe out my repository, but there are probably scenarios I'm not thinking of where that would be nice. An approach that uses something like `dotenv` explicitly would probably be nice to allow for easy use outside of Docker (i.e. straight `node`/`npm`, local database instance). I'm going to mark this is as resolved for now, but if there's a serious concern still here, please re-open and state what you'd like to see.

README.md
```diff
@@ -1,5 +1,19 @@
 ## Canvas Course Manager
```
pushyamig, line 1: The readme should have documentation of LTI configuration in canvas. That will help with new people join the team

ccm_web/server/appHandler.ts
```diff
@@ -0,0 +1,103 @@
...
+        // Set secure to true if the testing platform is in a different domain and https is being used
+        // Set sameSite to 'None' if the testing platform is in a different domain and https is being used
+        cookies: { secure: true, sameSite: 'None' }
+      }
+    )
+
+    // Redirect to the application root after a successful launch
+    provider.onConnect(async (token: IdToken, req: Request, res: Response) => {
+      console.log(token.userInfo)
+      return res.sendFile(path.join(this.envOptions.staticPath, 'index.html'))
```
pushyamig, line 58: Is Line#58 needed? since I see you are doing the same thing in Line#93-95? Felt redundant having in both places.

README.md
ssciolla: Okay, I will add some info. However, I want to avoid parroting similar docs we've already written for other projects. I will add a paragraph describing the Canvas configuration and provide a sample LTI config in JSON.

ccm_web/server/appHandler.ts
ssciolla: I'll check.

ccm_web/server/localTypes/ltijs/index.d.ts
ssciolla: Just checking to see if the documentation I provided in `localTypes/ltijs/index.d.ts` looks like it's enough to you, @pushyamig

pushyamig: Looks Good to me.

ccm_web/server/appHandler.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  // ltijs docs: https://cvmcosta.me/ltijs/#/
+  async setupLTI (): Promise<Express> {
+    const { db, lti } = this.config
+
+    const dbPlugin = new Database(
+      db.name,
+      db.user,
+      db.password,
+      { host: db.host, dialect: 'postgres', logging: false }
```
pushyamig, line 38: Not sure what is this `logging: false` means, if this is for debugging more message then this should be coming from the `.env`. You can leave it for now since now a issue for merging this code

ssciolla: This is a `sequelize` option. The `ltijs-sequelize` docs had `false` there. It likely prevents a bunch of SQL being echoed to the logs. It seems okay to me to have it be disabled (the default is `console.log`, looks like: https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html), people can remove it or change the value to a function if they want to see the `ltijs` database operations.

pushyamig:
This PR is in Good shape to Merge. I feel having this merged and more people testing the LTI work while working on there respective issues will bring in new insights and possible improvement to it. As we learn and grow with Node/Express we might do something better that what it is today. Nice work!

pushyamig (approved):

ssciolla:
@pushyamig, regarding backend logging, I created an issue a while back regarding this here: #38. I did some research, seems like we have some good options there for multiple levels of logging, more detailed messages, and API request logging for Express.

pushyamig:
> @pushyamig, regarding backend logging, I created an issue a while back regarding this: #38. I did some research, seems like we have some good options there for multiple levels of logging, more detailed messages, and API request logging for Express

It make sense to work on this before we start building more features. We can discuss about this in our meeting

config/lti_dev_key_sample.json
```diff
@@ -0,0 +1,31 @@
...
+  "scopes": [],
+  "extensions": [
+    {
+      "platform": "canvas.instructure.com",
+      "settings": {
+        "platform": "canvas.instructure.com",
+        "default": "visible",
+        "placements": [
+          {
+            "placement": "course_navigation",
```
pushyamig, line 12:
```suggestion
            "visibility": "admins",
```

config/lti_dev_key_sample.json
```diff
@@ -0,0 +1,31 @@
...
+            "message_type": "LtiResourceLinkRequest",
+            "target_link_uri": "https://{ccm_app_url}"
```
pushyamig, line 14: You need to add `"visibility": "admins"` since this tool should only be visibility for admins in a course

config/lti_dev_key_sample.json
```diff
@@ -0,0 +1,31 @@
+{
+  "title": "Canvas Course Manager",
+  "scopes": [],
+  "extensions": [
+    {
+      "platform": "canvas.instructure.com",
+      "settings": {
+        "platform": "canvas.instructure.com",
+        "default": "visible",
```
pushyamig, line 9: I don't think this is needed

config/lti_dev_key_sample.json
ssciolla: Done.

config/lti_dev_key_sample.json
ssciolla: Done. https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/40/commits/05a1cde52f98dfe7504581dab3fc6005873d06c3

ssciolla:
Thank you very much, @pushyamig, for your review. Apologies to everyone for a long PR process and too many changes and moving parts. I will strive to do better at keeping PRs small and limited in scope in the future. I'm merging now.

PR #43, opened 2021-03-24
Basic breadcrumb implementation.

chrisrrowland:
Added basic breadcrumbs.
Automatically constructed from the route so this *should* continue to work when we get more than 2 deep.

Each route is now automatically added as defined by the featureCardProps array defined in Home.tsx (for now)

All features currently direct to the placeholder Merge component (because they have to have a component defined) but use appropriate routes.

### DIFF
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -13,6 +13,10 @@ const useStyles = makeStyles((theme) => ({
   }
 }))
 
+interface TitleTypographyProps {
+  to?: string
+}
+
 function App (): JSX.Element {
   const classes = useStyles()
   return (
@@ -34,17 +38,12 @@ function App (): JSX.Element {
                       const last = index === pathnames.length - 1
                       const to = `/${pathnames.slice(0, index + 1).join('/')}`
 
-                      return last
-                        ? (
-                            <Typography color="textPrimary" key={to}>
-                              {featureCardProps.filter(feature => { return feature.route.substring(1) === value })[0].title}
-                            </Typography>
-                          )
-                        : (
-                            <RouterLink color="textPrimary" to={to} key={to}>
-                              {featureCardProps.filter(feature => { return feature.route.substring(1) === value })[0].title}
-                            </RouterLink>
-                          )
+                      const feature = featureCardProps.filter(feature => { return feature.route.substring(1) === value })[0]
+                      const titleTypographyProps: TitleTypographyProps = last ? { to: to } : {}
+
+                      return <Typography color='textPrimary' key={to} {...titleTypographyProps}>
+                        {feature.title}
+                      </Typography>
                     })}
                   </Breadcrumbs>
                 )
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -6,7 +6,7 @@ import NavigateNextIcon from '@material-ui/icons/NavigateNext'
 import ConsumerTest from './components/ConsumerTest'
 import Home from './pages/Home'
 import './App.css'
-import AllFeatures from './models/FeatureCardData'
+import allFeatures from './models/FeatureUIData'
 
 const useStyles = makeStyles((theme) => ({
   breadcrumbs: {
@@ -21,7 +21,7 @@ interface TitleTypographyProps {
 
 function App (): JSX.Element {
   const classes = useStyles()
-  const features = AllFeatures
+  const features = allFeatures
   return (
     <div className='App'>
       <Router>
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -19,11 +19,11 @@ interface TitleTypographyProps {
   to?: string
 }
 
-function HomeBreadcrumb (pathnames: string[]): JSX.Element {
+function HomeBreadcrumb (isLink: boolean): JSX.Element {
   const typography = (<Typography color='textPrimary'>
                         Canvas Course Manager
                       </Typography>)
-  return pathnames.length > 0
+  return isLink
     ? (<Link component={RouterLink} to='/'>{typography}</Link>)
     : (typography)
 }
@@ -41,7 +41,7 @@ function App (): JSX.Element {
                 const pathnames = location.pathname.split('/').filter(x => x)
                 return (
                   <Breadcrumbs aria-label="breadcrumb" separator={<NavigateNextIcon fontSize="small" />}>
-                    {HomeBreadcrumb(pathnames)}
+                    {HomeBreadcrumb(pathnames.length > 0)}
                     {pathnames.map((value, index) => {
                       const last = index === pathnames.length - 1
                       const to = `/${pathnames.slice(0, index + 1).join('/')}`
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -19,9 +19,26 @@ interface TitleTypographyProps {
   to?: string
 }
 
+function HomeBreadcrumb (pathnames: string[]): JSX.Element {
+  return pathnames.length > 0
+    ? (
+      <Link component={RouterLink} to='/'>
+        <Typography color='textPrimary'>
+          Canvas Course Manager
+        </Typography>
+      </Link>
+      )
+    : (
+      <Typography color='textPrimary'>
+        Canvas Course Manager
+      </Typography>
+      )
+}
+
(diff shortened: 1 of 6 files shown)


ccm_web/client/src/App.tsx
```diff
@@ -1,18 +1,61 @@
...
+                const pathnames = location.pathname.split('/').filter(x => x)
+                return (
+                  <Breadcrumbs aria-label="breadcrumb">
+                    <Link component={RouterLink} to='/'>
+                      <Typography color='textPrimary'>
+                        Canvas Course Manager
+                      </Typography>
+                    </Link>
+
+                    {pathnames.map((value, index) => {
```
pushyamig, line 33: @chrisrrowland  These seems to be variables used below and I can see types are missing. should have been `(value: string, index: int)`. 

Trying to understand why is my IDE is complaining type are missing. We do have strict mode enabled for type. why aren't these flagged.

I am trying to understand how the types linting is working here. Same is true for line 24 where `const pathnames = location.pathname.split('/').filter(x => x) ` can be expressed as
`const pathnames: string = location.pathname.split('/').filter(x => x)`

@ssciolla you can also give some input

chrisrrowland: I'm not sure.  I have seen warnings before when it says the type can be inferred and is not necessary.  
In this case I don't get a warning if the types are explicitly specified or left out.
I'm using ESLint 2.1.19 in VSCode

chrisrrowland: What is the lint error code? @pushyamig

pushyamig: Currently, I don't see any errors with regards to why types are missing.

pushyamig: Ok, I and Chris had conversation certain time types are inferred.

pushyamig (commented):
This looks good. I will look forward for the when 3 level of bread crumbs get wired

ccm_web/client/src/App.tsx
```diff
@@ -1,18 +1,61 @@
...
+                        ? (
+                            <Typography color="textPrimary" key={to}>
+                              {featureCardProps.filter(feature => { return feature.route.substring(1) === value })[0].title}
+                            </Typography>
+                          )
+                        : (
+                            <RouterLink color="textPrimary" to={to} key={to}>
+                              {featureCardProps.filter(feature => { return feature.route.substring(1) === value })[0].title}
+                            </RouterLink>
+                          )
```
ssciolla, line 47: You should be able to simplify this a bit. The element's children in each case is identical.

ssciolla, line 47: You should be able to simplify this a bit, the code inside of each element is identical.

ccm_web/client/src/pages/Home.tsx
```diff
@@ -24,15 +25,17 @@ const mergeSectionProps: FeatureCardProps = {
   description: 'Combine sections into one Canvas site for easier management',
   icon: <MergeTypeIcon fontSize='large' />,
   ordinality: 1,
-  route: '/merge-sections'
+  route: '/merge-sections',
+  component: MergeSections
```
ssciolla, line 29: I think we're starting to get to the point where it does make sense to extract these objects from `Home`. They could go in a `features.ts` module that gets imported by `Home` and `App`. That would also give us an opportunity to simplify some code variable names from `featureCardProps` to `feature`. If it were me, I'd create a `Feature` interface that's identical to the `FeatureCardProps` one and then adjust the `FeatureCard` component to accept a `Feature`. I think it will help us out in the long run if we start to think of these objects in a way less attached to one use case (`*Card`).

ccm_web/client/src/App.tsx
```diff
@@ -1,18 +1,61 @@
...
+
 function App (): JSX.Element {
+  const classes = useStyles()
   return (
     <div className='App'>
       <Router>
+        <div className={classes.breadcrumbs}>
+          <Route>
+              {({ location }) => {
+                const pathnames = location.pathname.split('/').filter(x => x)
```
ssciolla, line 24: I think this is cool and extensible, but something about it feels a bit fragile to me (all the string manipulation and matching?) or overly complex. What happens if there is a trailing slash? (legitimate question, I haven't thought it through yet.)

Here's perhaps one alternative. Each component in `pages` uses a `Layout` component that includes a `BreadCrumbs` component and a footer of some sort, as well as allowing children. You can pass hard-coded crumb information (that you know because you're in a component) into `Layout`. One positive side effect of this would be that we minimize code the amount of code in `App`.

Food for thought. Happy to discuss the pros and cons with you if you'd like.

ssciolla (commented):
I'm going to spend some more time thinking about this, but here are some initial thoughts.

ccm_web/client/src/App.tsx
chrisrrowland: Changed this. Not sure if its simplified or not.  I think the line count was unchanged.

ccm_web/client/src/pages/Home.tsx
chrisrrowland: I refactored this, however the configuration still requires some UX specific configuration.
Now there is  feature.ts that contains non UX related config (everything except the icon and the component for the feature) and the rest is configured in FeatureCard.  That could be put into yet another ts file too, though its not clear to me that would  present any big advantage?

ccm_web/client/src/App.tsx
chrisrrowland: I'd like to see how this plays out. It's based on how the  material docs demonstrated interacting with react-router.

ssciolla: Okay, can you link to the documentation you're talking about? Danke

ccm_web/client/src/App.tsx
ssciolla: Looks a little better to me. Interesting that you don't need the `RouterLink` component...

ccm_web/client/src/App.tsx
```diff
@@ -1,18 +1,61 @@
...
+                return (
+                  <Breadcrumbs aria-label="breadcrumb">
+                    <Link component={RouterLink} to='/'>
+                      <Typography color='textPrimary'>
+                        Canvas Course Manager
+                      </Typography>
+                    </Link>
```
ssciolla, line 37: BreadCrumbs look nice in terms of spacing and size. Accessibility/UX folks may encourage us to make the linked crumbs look noticeable in some way besides the hover underline. Do the wireframes give a specific separator to use? Looks like you can pass that to `Breadcrumbs`: https://material-ui.com/components/breadcrumbs/#custom-separator
Screen Shot 2021-03-29 at 3 58 15 PM

ccm_web/client/src/App.tsx
ssciolla: Mmm, here I guess: https://material-ui.com/components/breadcrumbs/#integration-with-react-router

I'll mark this as resolved.

ccm_web/client/src/App.tsx
chrisrrowland: At least one of the wireframes has a ">" separator but I overlooked that detail. I can certainly change that.

ccm_web/client/src/App.tsx
```diff
@@ -1,18 +1,62 @@
...
+import Home from './pages/Home'
 import './App.css'
+import allFeatures from './models/FeatureUIData'
+
+const useStyles = makeStyles((theme) => ({
+  breadcrumbs: {
+    paddingLeft: 25,
+    paddingTop: 25
+  }
+}))
```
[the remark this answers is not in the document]
chrisrrowland: Agreed there is no global theme at the moment

ccm_web/client/src/models/FeatureUIData.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { ComponentType } from 'react'
+import AccountCircleOutlinedIcon from '@material-ui/icons/AccountCircleOutlined'
+import LibraryBooksOutlinedIcon from '@material-ui/icons/LibraryBooksOutlined'
+import MergeTypeIcon from '@material-ui/icons/MergeType'
+import PersonAddIcon from '@material-ui/icons/PersonAdd'
+import PersonAddOutlinedIcon from '@material-ui/icons/PersonAddOutlined'
```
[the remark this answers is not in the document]
chrisrrowland: It didn't seem possible to do this with the Icons, at least the way I tried

Issue #26, opened 2021-03-05
Feature: Rename of Course name


pushyamig:
This feature is Admin/subaccount feature only. When ever the course name is updated course code needs to be updated

API: https://umich.test.instructure.com/api/v1/courses/:id?course[name]=updata course name&course[course_code]=updata course name

Details about the API listed [here](https://docs.google.com/spreadsheets/d/1pm5y9FX0zrDX7Zy3mOyDLxmA-iKoWfmlxvbtNkWg6Zw/edit#gid=578161201)

ssciolla:
@pushyamig, can we close this now that the client and server are wired up?

pushyamig:
Test passed!

PR #47, opened 2021-03-30
Course Rename

chrisrrowland:
I've done some work to create an inline text field editor for course rename.

I'm not sure how the data is going to get fed into it, and how it is going to save so it's going to have to be modified, but hopefully the general idea is sound.  Right now it is just passed a string, the course name, and a function to pass a string to to save changes.

That function is defined in Home.tsx where the InlineTextEdit element is declared.  I'm sure this will annoy @ssciolla :)

You can enable edit mode by clicking on the pencil icon, or clicking on the disabled text field (I haven't figured out how to override the cursor on the text field thus far)

Click save or cancel to save or cancel.  Enter or num-pad enter save. Escape cancels.  Only changes will be saved.
Trying to save 'reject' as a course name demonstrates a failed  save.

Success/failure notification through snackbar (toast) added to main app

### DIFF
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { BrowserRouter as Router, Route, Switch } from 'react-router-dom'
+import { SnackbarProvider } from 'notistack'
 
 import ConsumerTest from './components/ConsumerTest'
 import Home, { mergeSectionProps } from './pages/Home'
@@ -9,14 +10,16 @@ import './App.css'
 function App (): JSX.Element {
   return (
     <div className='App'>
-      <Router>
-        <Switch>
-          <Route exact={true} path="/" component={Home} />
-          <Route path={mergeSectionProps.route} component={MergeSections} />
-          <Route render={() => (<div><em>Under Construction</em></div>)} />
-        </Switch>
-      </Router>
-      <ConsumerTest/>
+      <SnackbarProvider maxSnack={3}>
+        <Router>
+          <Switch>
+            <Route exact={true} path="/" component={Home} />
+            <Route path={mergeSectionProps.route} component={MergeSections} />
+            <Route render={() => (<div><em>Under Construction</em></div>)} />
+          </Switch>
+        </Router>
+        <ConsumerTest/>
+      </SnackbarProvider>
     </div>
   )
 }
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -20,8 +20,10 @@ interface TitleTypographyProps {
   to?: string
 }
 
+const breadcrumbVariant = 'h5'
+
 function HomeBreadcrumb (isLink: boolean): JSX.Element {
-  const typography = (<Typography color='textPrimary'>
+  const typography = (<Typography color='textPrimary' variant={breadcrumbVariant}>
                         Canvas Course Manager
                       </Typography>)
   return isLink
@@ -50,7 +52,7 @@ function App (): JSX.Element {
                         const feature = features.filter(f => { return f.route.substring(1) === value })[0]
                         const titleTypographyProps: TitleTypographyProps = last ? { to: to } : {}
 
-                        return (<Typography color='textPrimary' key={to} {...titleTypographyProps}>
+                        return (<Typography color='textPrimary' variant={breadcrumbVariant} key={to} {...titleTypographyProps}>
                           {feature.data.title}
                         </Typography>)
                       })}
--- a/ccm_web/client/src/components/InlineTextEdit.tsx
+++ b/ccm_web/client/src/components/InlineTextEdit.tsx
@@ -12,14 +12,25 @@ interface InlineTextEditProps {
 
 const useStyles = makeStyles((theme) => ({
   root: {
+    textAlign: 'left',
     padding: '5px',
     '& .MuiInputBase-root.Mui-disabled': {
       color: 'rgba(0, 0, 0, 0.6)' // (default alpha is 0.38)
+    },
+    '& svg': {
+      paddingTop: '6px',
+      paddingBottom: '7px'
     }
+  },
+  inputRow: {
+    height: '60px',
+    paddingTop: '6px',
+    paddingBottom: '7px',
+    fontSize: '30px'
   },
   editIcon: {
     cursor: 'pointer',
-    textAlign: 'left'
+    fontSize: '30px'
   },
   inputArea: {
     width: '500px'
@@ -70,20 +81,20 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
   return (
     <form className={classes.root} noValidate autoComplete="off">
       <Grid container className={classes.inputArea}>
-        <Grid item xs={12} sm={8}>
-          <TextField onClick={toggleEdit} inputProps={{ style: { fontSize: 30 } }} ref={textInput} id="standard-basic" placeholder='Course Name' value={tempName} onKeyDown={(e) => keyPress(e.code)} onChange={(e) => setTempName(e.target.value)} disabled={!isEditing}/>
+        <Grid item xs={8} sm={8}>
+          <TextField className={classes.inputRow} aria-readonly={false} onClick={toggleEdit} inputProps={{ style: { fontSize: 30 } }} ref={textInput} id="standard-basic" placeholder='Course Name' value={tempName} onKeyDown={(e) => keyPress(e.code)} onChange={(e) => setTempName(e.target.value)} disabled={!isEditing}/>
         </Grid>
-      {!isEditing
-        ? (<Grid item xs={4} className={classes.editIcon} ><EditIcon onClick={toggleEdit}/></Grid>)
-        : (<Grid container item xs={12} sm={4}>
-            <Grid item xs={6} sm={6} >
-              <Button disabled={courseName === tempName} onClick={save}>Save</Button>
-            </Grid>
-            <Grid item xs={6} sm={6} >
-              <Button onClick={cancel}>Cancel</Button>
-            </Grid>
-          </Grid>)
-      }
+        {!isEditing
+          ? (<Grid item xs={4} className={classes.inputRow}><EditIcon className={classes.editIcon} fontSize='inherit' onClick={toggleEdit}/></Grid>)
+          : (<Grid container item xs={12} sm={4} className={classes.inputRow}>
+              <Grid item xs={6} sm={6} >
+                <Button disabled={courseName === tempName} onClick={save}>Save</Button>
+              </Grid>
+              <Grid item xs={6} sm={6} >
+                <Button onClick={cancel}>Cancel</Button>
+              </Grid>
+            </Grid>)
+        }
       </Grid>
     </form>
   )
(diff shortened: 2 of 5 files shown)


ccm_web/package.json
```diff
@@ -18,6 +18,8 @@
     "css-loader": "^5.1.1",
     "express": "^4.17.1",
     "html-webpack-plugin": "^5.2.0",
+    "keycode-js": "^3.1.0",
+    "notistack": "^1.0.5",
```
ssciolla, line 22: We should discuss the choice of these dependencies in a meeting and make sure they align with the rest of the project goals. `notistack` seems pretty neat, though it makes sense to do a little research to figure out what else is out there. I don't quite understand `keycode-js` (or the need for it) yet; seems like it's cool way to customize keyboard control, but customization may get in the way of accessibility (by violating expectations). I'm guessing (but could be wrong) that we can get by without it, and I generally would like to minimize dependencies.

ssciolla (commented):
Seems like a nice start on some stuff we'll need. It touches a couple different significant issues (which I don't think are documented yet): form fields and notifications, probably others. We may need to do some client-side validation for entries, but that could be done in the future, so long as the backend has validation.

ccm_web/package.json
chrisrrowland: keycode-js is just a package that defines a bunch of constants so you can do this

```
  const keyPress = (code: string): void => {
    if (code === CODE_ENTER || code === CODE_NUMPAD_ENTER) {
      save()
    } else if (code === CODE_ESCAPE) {
      cancel()
    }
  }
```
instead of
```
  const keyPress = (code: string): void => {
    if (code === 'Enter' || code === 'NumpadEnter') {
      save()
    } else if (code === 'Escape') {
      cancel()
    }
  }
```

chrisrrowland: I chose notistack becasue it was recommended by the [material-ui documentation](https://material-ui.com/components/snackbars/#notistack).  That carried enough weight for me but I didn't shop around, either.

pushyamig:
Looking at the way it is implemented the Canvas Course Manger should be higher heading order. Currently it should be smaller size than CCM course rename edit feature

Screen Shot 2021-03-31 at 11 59 55 AM
Screen Shot 2021-03-31 at 12 00 06 PM

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { useRef, useState } from 'react'
+import { makeStyles } from '@material-ui/core/styles'
+import { Button, Grid, TextField } from '@material-ui/core'
+import { Edit as EditIcon } from '@material-ui/icons'
+import { useSnackbar } from 'notistack'
+import { CODE_ENTER, CODE_NUMPAD_ENTER, CODE_ESCAPE } from 'keycode-js'
+
+interface InlineTextEditProps {
```
pushyamig, line 8: Can this be renamed to `CourseNameTextEditProps`? Same goes with function name. That way it can be understood as it is renaming course

ccm_web/client/src/pages/Home.tsx
```diff
@@ -67,8 +68,18 @@ function Home (): JSX.Element {
...
   const cards: FeatureCardProps[] = [mergeSectionProps, gradebookToolsProps, createSectionsProps, addUMUsersProps, addNonUMUsersProps]
 
+  const saveCourseName = async (courseName: string): Promise<void> => await new Promise<void>((resolve, reject) => {
+    console.log('saveCourseName ' + courseName)
+    if (courseName === 'reject') {
+      reject(new Error('Rejected as requested'))
+    } else {
+      return resolve()
+    }
+  })
```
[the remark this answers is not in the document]
pushyamig: The LTI needs to provide the Course Name after the LTI launch. So far it looks like the this will hold in React Props, once the course rename is happened then it need to persisted into props. Probably this can use react state lifting.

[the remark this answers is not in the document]
pushyamig: I feel the saveCourseNamme function could be taken out Home, Probably into `InlineTextEdit`? Since all the req/res for features should be handled inside feature function. This is just a though No need to rework here. Since you are building UI with out API endpoint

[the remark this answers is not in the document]
ssciolla: In my limited experience, I have found it more intuitive to have API interactions at the highest level in a container (here `Home`), and then passed down as needed to child components (somewhat related to the lifting state up idea). We can discuss, but I think having a consistent pattern for where API calls are set up and handled will make it easier to expand and modify.

ccm_web/package.json
pushyamig: I do agree with @ssciolla that we don't want to pull a lot of dependencies since it make the main.js file bulk. `keycode-js` seems to be cool, but we can define our own constants as separate file and then just use what is needed.

ccm_web/client/src/App.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import { BrowserRouter as Router, Link as RouterLink, Route, Switch } from 'react-router-dom'
+import { SnackbarProvider } from 'notistack'
```
pushyamig, line 3: what is the reason for choosing an external Snack bar Library? did you felt that it supported the use cases for CCM that default material-UI  did not had?

ccm_web/client/src/pages/Home.tsx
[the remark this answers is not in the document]
pushyamig: > n my limited experience, I have found it more intuitive to have API interactions at the highest level in a container (here Home), 

OK, valid this is a new territory and we will grow as we learn

But I feel since this is Home file, the API can exits here but the internals can be pulled out to a separate function, that way Home is not cluttered.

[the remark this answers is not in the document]
ssciolla: Sure, I agree we shouldn't be `fetch`ing here, we can modularize it so it's relatively terse in this component.

pushyamig:
@chrisrrowland, I have converted this Course Rename feature as an Epic.  I see that this PR is dealing with building the UI component and once the LTI support sending the roles is in place along with an API end point then we can finish this https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/26

ccm_web/client/src/components/InlineTextEdit.tsx
chrisrrowland: It's more generic than course renaming, or at least intended to be.  It's basically just an inline editable text field.

chrisrrowland: I need to make the error message something that gets passed in, rather than having a course name specific message, but I was unsure precisely where that should be handled and didn't do that at this point.

ssciolla: Being generic is a good goal, but we probably need to rename some variables (e.g. `const [courseName, setCourseName`]) 😆

ccm_web/client/src/App.tsx
chrisrrowland: notistack was [recommended ](https://material-ui.com/components/snackbars/#notistack)as a 'complementary'  project by material ui.

It allows multiple notifications which I believe is desirable, and you don't have to manually manage state.

ccm_web/client/src/components/InlineTextEdit.tsx
pushyamig: Ok, In case it is generic I saw some of the variable names as `courseName` so I got bit confused. Of it is generic then course rename text should be removed and use more generic terms

chrisrrowland: > Being generic is a good goal, but we probably need to rename some variables (e.g. `const [courseName, setCourseName]`) 😆

Yes good point. I will fix that as well.

pushyamig: Yeah.. I agree it is bit difficult to thing through if the backend is still not wired up yet. The InlineTextEdit should take an error message from the calling function and should not include feature specific text. This can be handled when backend API endpoint and LTI support is added

chrisrrowland: I wasn't sure where it made sense to handle error/success, but I will create properties for the error/success messages for now.

chrisrrowland: This should be better now

chrisrrowland:
> Looking at the way it is implemented the Canvas Course Manger should be higher heading order. Currently it is smaller size than CCM course rename edit feature
> 
> Screen Shot 2021-03-31 at 11 59 55 AM
> Screen Shot 2021-03-31 at 12 00 06 PM

I think I will resolve this on a new PR related to breadcrumbs

ccm_web/client/src/App.tsx
pushyamig: ok

chrisrrowland:
> > Looking at the way it is implemented the Canvas Course Manger should be higher heading order. Currently it is smaller size than CCM course rename edit feature
> > Screen Shot 2021-03-31 at 11 59 55 AM
> > Screen Shot 2021-03-31 at 12 00 06 PM
> 
> I think I will resolve this on a new PR related to breadcrumbs

Actually its such a small change, I'm just including it here.

ccm_web/package.json
chrisrrowland: I think it's only [importing ](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/47/files#diff-51ae223315fee7af3e83fc1b165b53f1857295b24e0dc9a5953253ebcc064cb1R6) what is used?  I'm not entirely sure how minimizing the amount of code imported works but I thought this was safe

ccm_web/client/src/pages/Home.tsx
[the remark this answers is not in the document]
chrisrrowland: This was a placeholder for the sort of promise based functionality I was guessing would eventually be implemented, so I could build out the flow.  I don't think its super important to worry about where this code that is not going to be permanent lives at the moment.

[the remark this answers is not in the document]
pushyamig: Sure, this is general discussion so that next time when actual implementing then you know team preference how this should be structured

ccm_web/package.json
ssciolla: I guess I'd want to know why `keyPress` is even necessary (if it is, then something like this library might make sense). People who will be using keyboards to navigate the site would mostly be using tabs and Enter (probably not escape). Once they are done typing, I would think they would just hit Tab and then Enter once they're on the Save or Cancel buttons. Is that wrong or short-sighted somehow?

chrisrrowland: It is so when the user is editing the course name escape will fire the cancel event (same as the cancel button) and Enter will complete the edit (same as the save button)

pushyamig:
I see this is functional and don't see huge concerns. We discussed the area of improvement and that can be dealt later

pushyamig (approved):

ccm_web/package.json
ssciolla: Mmm, not sure. These behaviors seem reasonable to me, but UX team may say something else. I'm not familiar with regular uses of `esc` in an accessibility context. `Enter` is more common, and this logic seems right to me, but there may be another `onEnter` prop you could handle this with.

The concern with the dependency is that it's one more thing to keep updated/maintain, and you're only using it for two constants so far. If this were a game of some sort where you were using tons of keys, it would make mores sense to me. You could write your own constants here, or safely use the strings, right?

All that said, I don't see this is as a huge issue, or a blocker for me.

PR #57, opened 2021-04-19
Change database flavor from PostgreSQL to MySQL (#49)

### LINKED ISSUE #49: Switch from PostgreSQL to MySQL
In order to save costs and limit barriers to collaboration, we decided to change the database technology from PostgreSQL to MySQL. However, changing this is currently blocked by a bug in `ltijs-sequelize`: https://github.com/Cvmcosta/ltijs-sequelize/issues/2 For now, we can stick with PostgreSQL for finishing #23, and we can make the change once `ltijs-sequelize` is updated.

ssciolla:
This PR makes minor modifications to the back end and Docker artifacts to change the database system/image in use from PostgreSQL to MySQL. It also includes updates to the `ltijs` and `ltijs-sequelize` dependencies, some of which were needed to support MySQL properly. The PR aims to resolve #49.


pushyamig:
I am planning to review this by end of today, if not tomorrow

docker-compose-prod.yml
```diff
@@ -18,13 +18,19 @@ services:
       - database
 
   database:
-    image: postgres:13
+    image: mysql:5.7
```
pushyamig, line 21: is there any reason we are still need to stick with older version? I know our other app still are on 5.7 when 8 is available. We can talk to DBA team if they support 8 or not? does LTIJS support latest version of MySQL as well?

docker-compose-prod.yml
```diff
@@ -18,13 +18,19 @@ services:
...
     environment:
-      - POSTGRES_DB=ccm
-      - POSTGRES_USER=admin
-      - POSTGRES_PASSWORD=admin
+      - MYSQL_ROOT_PASSWORD=ccm_root_pw
+      - MYSQL_HOST=ccm_db_prod
+      - MYSQL_PORT=3306
+      - MYSQL_DATABASE=ccm
+      - MYSQL_USER=admin
+      - MYSQL_PASSWORD=admin
```
pushyamig, line 30: Can we have same mysql property names as in `.env` file?

pushyamig:
The PR is functional and work well on both development and Prod build.

docker-compose-prod.yml
jonespm: I would also fix the version to be specific to what we're likely running. Last I checked it was mysql:5.7.26 but might have updated to 5.7.33. There are some minor issues that have arisen from using a different database locally verses in production.

I'm also thinking we should change the flag of 
`    restart: on-failure`

in all of our apps to either `unless-stopped` or just `off` so it doesn't restart all of the MySQL containers after a restart. It's a little annoying having to kill them off, maybe I'm the only one.

jonespm: Last time I checked they only supported MySQL 5.7. There's some general debate among MySQL users on whether to update to MySQL 8 or just switch to MariaDB. (Which is more generally more compatible with 5.7 and has different features)

I'd just use whatever MiDatabase supports but we'd want to wait till they update and we could check.

jonespm (approved):
Looks good to me. ltij-sequelize still is using text for most of the columns which hopefully will be improved before CCM releases to production. We'd have this problem with either MySQL or PostgeSQL though. Only using MongoDB would we avoid that.

docker-compose-prod.yml
pushyamig: OK, we will stick with 5.7. See conversation in the TLDev slack channel if you want.

docker-compose-prod.yml
ssciolla: I'm not sure what you're asking for here. These need to be set up this way in the MySQL container. I could change the database parameters in the application container so they use these names. Is that what you're asking for? I guess I slightly prefer not including the database flavor name in the variables, but could be convinced otherwise. I also think `db.name` is clearer in our application code than `db.database`.

https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/3f7536de69f87cbe80902473209f874aefb1e270/ccm_web/server/config.ts#L62-L67

docker-compose-prod.yml
ssciolla: @jonespm

Regarding the `restart` policy, I think I'll just remove it for now (default is `off`). It doesn't seem super important, and the `on-failure` behavior could get annoying (though I don't know if I've ever experienced it myself).

How much of a risk is it to stick just with `mysql:5.7`?  We always test with MiDatabase instances before we roll to production. I feel like the time spent debugging problems with patch differences (which should in theory be no time) could be less than keeping up with each patch change the MiDatabase team rolls out.

docker-compose-prod.yml
pushyamig: Yes! OK if you leave it off. I don't have a convincing reason but I looks like 2 different DB configuration options from face of it one for local and other for non-prod

ssciolla: Hm, I'll follow up with you directly, I would like to understand your thinking here.

docker-compose-prod.yml
ssciolla: Regarding the patch version question, I'm going to merge this as is, and when/if we decide to do this for other applications, we can do it here as well.

docker-compose-prod.yml
ssciolla: Going to keep it as is for now.

ssciolla:
Thanks for the reviews! Merging now.

PR #59, opened 2021-04-21
Add server logging approach using `winston` (#38)

### LINKED ISSUE #38: Adopt a logging approach for server

ssciolla:
This PR adds a server logging approach similar to our other applications using the `winston` library, which seems widely used (though has a lot of open PRs/issues). On a high level, a `baseLogger` is set with most configuration; configuration validation logic uses the default logging; the minimum log level shown is set in `app.ts` after the configuration is loaded; and each module sets up a child logger, passing `__filename` so the logger format function can determine the module. The PR aims to resolve #38.

Note: I am choosing to wait on determining the best HTTP request logging approach until we decide whether we will add an additional framework/tool for building out the internal application API. I will separate that piece out into a new issue.

Resources:
* [`winston`](https://github.com/winstonjs/winston)

### DIFF
--- a/ccm_web/server/logger.ts
+++ b/ccm_web/server/logger.ts
@@ -1,0 +1,21 @@
+import path from 'path'
+import winston, { format } from 'winston'
+
+const logFormat = format.printf(
+  ({ level, message, timestamp, filePath }) => {
+    const moduleName = path.basename(filePath, '.ts')
+    return `${String(timestamp)} - ${moduleName} - ${level} - ${message}`
+  }
+)
+
+const baseLogger = winston.createLogger({
+  transports: [new winston.transports.Console()],
+  format: format.combine(
+    format.timestamp(),
+    format.colorize(),
+    logFormat
+  ),
+  exitOnError: false
+})
+
+export default baseLogger
--- a/ccm_web/server/logger.ts
+++ b/ccm_web/server/logger.ts
@@ -3,7 +3,7 @@ import winston, { format } from 'winston'
 
 const logFormat = format.printf(
   ({ level, message, timestamp, filePath }) => {
-    const moduleName = path.basename(filePath, '.ts')
+    const moduleName = path.basename(filePath)
     return `${String(timestamp)} - ${moduleName} - ${level} - ${message}`
   }
 )
--- a/ccm_web/package-lock.json
+++ b/ccm_web/package-lock.json
@@ -160,6 +160,16 @@
         "regenerator-runtime": "^0.13.4"
       }
     },
+    "@dabh/diagnostics": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/@dabh/diagnostics/-/diagnostics-2.0.2.tgz",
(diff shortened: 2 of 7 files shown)


pushyamig:
This works ok, I did not see any oddities in the configuration. Only thing is LTIJS Debug and winston logging don't go hand in hand, which is OK (not a requirement)

pushyamig (approved):

lsloan (commented):
My Q about this is where do the log entries appear?  Are they going out to the console only?  Is anything being logged to files?

ssciolla:
> My Q about this is where do the log entries appear? Are they going out to the console only? Is anything being logged to files?

@lsloan, we can add additional transports, like a file transport, here if we want to:
https://github.com/ssciolla/canvas-course-manager-next/blob/1485eedd98a752b2bdbc2ff2bff893268aeb6cd8/ccm_web/server/logger.ts#L11-L12

See https://www.npmjs.com/package/winston#transports

That didn't seem necessary yet, as we have Jenkins and OpenShift logs. Are you partial to file logs? This doesn't seem like something we need to decide now.

jonespm:
For running this on Openshift I think console is ideal.

ccm_web/server/logger.ts
```diff
@@ -0,0 +1,21 @@
...
+    const moduleName = path.basename(filePath)
+    return `${String(timestamp)} - ${moduleName} - ${level} - ${message}`
+  }
+)
+
+const baseLogger = winston.createLogger({
+  transports: [new winston.transports.Console()],
+  format: format.combine(
+    format.timestamp(),
+    format.colorize(),
```
jonespm, line 15: This might be something we want to be able to disable. It probably looks good in dev but not sure what this is going to look like on Openshift or Kibana?

jonespm (approved):

jonespm:
It looks like winston and log4js are the two most popular libraries, this looks fine to me to use this one.

ccm_web/server/logger.ts
ssciolla: Hm, was trying to keep configuration out of this file, but we could branch on this based on `NODE_ENV` (`production` or `development`), or just remove it. What do you think? We could also wait until this is deployed in a QA setting and see what happens.

ssciolla: I'm going to wait before changing this, but I'm glad you flagged in, I am not sure how OpenShift will respond.

pushyamig:
Test passes. I don't see any console log message in the `server` folder of `ccm_web`

PR #63, opened 2021-04-26
Fixes #62 - Display full message when an error is caught in the main loop

### LINKED ISSUE #62: Display full message when an error is caught in the main loop
When an error is caught in the main loop a message of

"An error occurred while starting the application." is printed without any additional context. This should also display with a full stack trace to help track down the cause of the error.

The easiest way to test this is to have a `.env` file without setting `LTI_CLIENT_ID` (As is the current default in the `.env.sample`)

jonespm:


pushyamig (approved):
Looks Good!

PR #66, opened 2021-04-29
Basic ORM integration with Sequelize

### LINKED ISSUE #46: Select and integrate Database ORM to use
We might need store data into database like scoped token of each user, Logging which user is storing which feature when making an API call. So selecting a ORM and doing that way is practical approach.

pushyamig:
This is the first pass at ORM integration. I am opening as Draft since it needs some improvements with using JS files for dbConfig, still exploring the best options `sequelize-typescript` or `NestJS` ORM library

Fixes #46


ccm_web/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "dev:server": "nodemon --inspect=0.0.0.0:9229 --config nmon_config.json server/app.ts",
     "dev:client": "webpack --watch --config webpack/webpack.dev.ts",
-    "dev": "npm run dev:server & npm run dev:client",
+    "dev": "wait-port ccm_db:3306 && sequelize db:migrate & npm run dev:server & npm run dev:client",
```
pushyamig, line 8: @ssciolla indeed you are right, npx was redundant and simple library names was sufficient to run the command.

pushyamig:
I am closing the PR since i am continue this work over here #80

PR #68, opened 2021-05-03
Re-structure and improve backend using NestJS (#67)

### LINKED ISSUE #67: Explore better Framework than express
Express being un opinionated framework seems to be adding little structure. The API endpoint work can benefit from having a better structure and similarly DB ORM as well

ssciolla:
This PR introduces an additional framework, [NestJS](https://docs.nestjs.com/), to the Node backend to make API endpoint development more efficient and provide additional structure and testing capabilities. To create this PR, I did the following:

1) Added the starter project (https://github.com/nestjs/typescript-starter) under `server` using the CLI
2) Moved/combined/cut various artifacts (e.g. `package.json`, `.eslintrc.js`, and `tsconfig.json`) balancing the needs of the existing project with the demands of NestJS
3) Migrated existing code for serving static, configuration, logging, and LTI into the NestJS application, using modules, providers, controllers, and middleware as appropriate.

Feedback is welcome. The PR will resolve #67.

To Do
- [x] Get `ltijs` initialization to run on module load
- [x] Move `AppController` and `AppService` to a different module (`BaseModule`)
- [x] Fix `jest` tests for existing API routes
- [x] Tidy up npm commands
- [x] Tidy up `tsconfig.json`
- [x] Fix end-to-end tests
- [x] Document test running

Separate issues?
- [ ] Add Swagger?
- [ ] Dynamically load `LtiModule` based on configuration flag?
- [ ] Assess static approach
- [ ] Try to resolve client routing passthrough issue


config/lti_dev_key_sample.json
```diff
@@ -25,7 +25,7 @@
     "roles": "$Canvas.membership.roles",
     "isRootAccountAdmin": "$Canvas.user.isRootAccountAdmin"
   },
-  "public_jwk_url": "https://{ccm_app_url}/keys",
-  "target_link_uri": "https://{ccm_app_url}",
-  "oidc_initiation_url": "https://{ccm_app_url}/login"
+  "public_jwk_url": "https://{ccm_app_url}/lti/keys",
+  "target_link_uri": "https://{ccm_app_url}/lti",
+  "oidc_initiation_url": "https://{ccm_app_url}/lti/login"
 }
```
ssciolla, line 31: @pushyamig, @lsloan, @jonespm, please note the change in configuration here; all the `ltijs` routes start with `/lti` now.

ccm_web/server/src/app.module.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { AppController } from './app.controller'
+import { AppService } from './app.service'
+import { LtiModule } from './lti/lti.module'
+
+import { validateConfig } from './config'
+
+@Module({
+  imports: [
+    ConfigModule.forRoot({
```
pushyamig, line 12: I see the ConfigModule in NestJS supports loading of configuration from yaml, I suppose this would be similar to json as well. What are your thoughts on changing it to JSON?

ssciolla: I'm not opposed to using a YAML or JSON configuration file. The NestJS documentation does suggest that using environment variables is a best practice, but I'm open. I think this could be done in a separate issue though.

pushyamig:
I am planning on complete the review by tomorrow.

config/lti_dev_key_sample.json
pushyamig: Nice improvement on having LTI specific URL with `lti` routes

ccm_web/package.json
```diff
@@ -3,19 +3,34 @@
...
     "build-client": "webpack --config webpack/webpack.prod.ts",
     "build-server": "tsc",
     "preprod": "npm run build-client && npm run build-server",
-    "prod": "NODE_ENV=production node dist/server/app.js"
+    "prod": "NODE_ENV=production node dist/server/src/main.js",
+    "prebuild": "rimraf dist",
+    "build": "nest build",
+    "start": "nest start",
+    "start:dev": "nest start --watch",
+    "start:debug": "nest start --debug --watch",
```
pushyamig, line 17: Line 16 and 17 seems to be doing what Line 6 is doing. I see that you have installed `@nest/cli` and `nodemon` packages. Can we support either one of them? So this way less libraries. Anyway we won't be using nest CLI in Prod/Non-prod and just Locally.

pushyamig:
One observation, Angular is built around Typescript support. I feel NestJS is following same. I was looking at other NodeJs frameworks few only written in typescript. In that **narrow** perspective this would be a good choice also we made choice to write BE in typescript as well.

ssciolla:
> Midddleware → Middleware with extra middle. 😉

@lsloan, yep, do you have a suggestion about this? We could probably move the work of `LTIService` into the middleware.

lsloan (commented):
I know this PR is still in draft mode, so this is a just a comment.

I have the code from your branch running on my computer and I'm developing my understanding of how it works.

ccm_web/tsconfig.json
```diff
@@ -10,6 +14,7 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
```
ssciolla, line 16: The `@nestjs` starter project (https://github.com/nestjs/typescript-starter) uses a library check (using the `skipLibCheck`default of `false`). When I tried building for prod with the default, there were a number of errors thrown, mostly from `webpack`-related types. The idea is that TypeScript checks the consistency of all the declaration files used by your project. TypeScript recommends `true` (I believe for performance reasons), but we should explore this a little more. I will open a new issue.

ccm_web/package.json
```diff
@@ -24,9 +33,14 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-router-dom": "^5.2.0",
```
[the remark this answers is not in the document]
ssciolla: NestJS is the proposed framework for the NodeJS backend. React is the framework we are using in the frontend (see `client`). Because it's all JavaScript, we can use a single `package.json` to collect dependencies for both the server and client. I think there is some debate about whether to use one or multiple `package.json` files, but this is where we're at for now.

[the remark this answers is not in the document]
ssciolla: @lsloan, if you want to put together a list of questions about how all the pieces fit together, I am happy to meet with you individually to explain/discuss.

ccm_web/tsconfig.json
ssciolla: See #76.

ccm_web/package.json
ssciolla: Is the `package.json` looking better to you now? We could remove `@nestjs/cli` entirely from the `devDependencies` if you want. I think the documentation recommends you install it globally any way. https://docs.nestjs.com/cli/overview#installation

pushyamig: I feel we should remove `@nestjs/cli` even as dev dependency since it not used in building the project. It is CLI which make the life easier if want to create a new service or module. I think That info can go in the README more like FYI..

pushyamig:
> We could probably move the work of LTIService into the middleware.

I agree on that we should add the LTI as middleware where create a folder called `middleware` so things get organized and a correct pattern is established.

ssciolla:
> I agree on that we should add the LTI as middleware where create a folder called `middleware` so things get organized and a correct pattern is established.

@pushyamig, I'm not sure I follow what you're proposing, or maybe I don't understand how a separate folder called `middleware` would be a better place for this (unless maybe you're saying put it in `lti/middleware/lti.middleware.ts`). The major organizational unit in NestJS is the Module, so I figured the LTI stuff should have its own module (which we can try to load dynamically in the future, thus dynamically loading the middleware). I could see eliminating `LTIService`, and let the middleware class include all of the `litjs` code; that was what I was suggesting. Keeping middleware in the various modules they are most relevant to makes sense to me (though this is a bit different from what we're used to with Django). Happy to discuss further.

pushyamig:
> @pushyamig, I'm not sure I follow what you're proposing, or maybe I don't understand how a separate folder called middleware would be a better place for this (unless maybe you're saying put it in lti/middleware/lti.middleware.ts). The major organizational unit in NestJS is the Module, so I figured the LTI stuff should have its own module (which we can try to load dynamically in the future, thus dynamically loading the middleware). I could see eliminating LTIService, and let the middleware class include all of the litjs code; that was what I was suggesting. Keeping middleware in the various modules they are most relevant to makes sense to me (though this is a bit different from what we're used to with Django). Happy to discuss further.


@ssciolla some of the tutorials I was following along I saw this pattern so where a middleware was  module with it's own folder and `.middleware.ts` file also in `middleware` folder. Here is the [repo](https://github.com/adrien2p/nest-js-sequelize-jwt/) for it

Screen Shot 2021-05-07 at 3 29 41 PM

But I am not in a best position to access yet if your structure is a best practice or not. But again I will differ to you. I am ok either way. But I will do more research on this.

pushyamig:
Optional: @ssciolla can we create a directory called `module` and put all respective modules into it like `lti`, 'user`, etc`. I saw that way of organizing pretty common around nestJs projects

pushyamig:
Optional: any reason you went with `base` instead of `app`? If you look at the react project the starting point files also made `app`

pushyamig:
> I could see eliminating LTIService, and let the middleware class include all of the ltijs code; 

Yeah from what I understood from NestJS architecture design style, you create a service if you want to inject it to other modules/controllers. But I don't see this as injecting into other module. So the code inside could very well be inside the Middleware file. 

It's ok if you want to deal with this as separate issue

pushyamig:
What are your thoughts on using built-in NestJS logger vs Winston? 

NestJs [documentation](https://docs.nestjs.com/techniques/logger#use-external-logger) write for more robust prod logging tends to use external libraries. I don't have an opinion on this neither asking for change. The setup you have seems to be good.

ccm_web/package.json
```diff
@@ -3,19 +3,28 @@
...
     "dev:client": "webpack --watch --config webpack/webpack.dev.ts",
     "dev": "npm run dev:server & npm run dev:client",
-    "build-client": "webpack --config webpack/webpack.prod.ts",
-    "build-server": "tsc",
-    "preprod": "npm run build-client && npm run build-server",
-    "prod": "NODE_ENV=production node dist/server/app.js"
+    "build:client": "webpack --config webpack/webpack.prod.ts",
+    "build:server": "tsc --project tsconfig.build.json",
+    "preprod": "npm run build:client && npm run build:server",
+    "prod": "NODE_ENV=production node dist/server/src/main.js",
```
pushyamig, line 12: Running in Prod mode with `docker-compose-prod.yml` is throwing me this error. I don't see this issue in Dev mode

```
ccm_web_prod | internal/modules/cjs/loader.js:883
ccm_web_prod |   throw err;
ccm_web_prod |   ^
ccm_web_prod |
ccm_web_prod | Error: Cannot find module '/app/server/app.js'
ccm_web_prod |     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
ccm_web_prod |     at Function.Module._load (internal/modules/cjs/loader.js:725:27)
ccm_web_prod |     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
ccm_web_prod |     at internal/main/run_main_module.js:17:47 {
ccm_web_prod |   code: 'MODULE_NOT_FOUND',
ccm_web_prod |   requireStack: []
ccm_web_prod | }
```

pushyamig:
In Dev Mode:
1. Startup fine, LTI launches happens. It's a different issue that if your launching the App for first time you might see timeout related issue. This is fine. could be addressed with `wait-port`. 
2. Nodeman restarts when server changes happens, VSCode debugger is attached nicely
3. Webpack main_*.js dump file is nicely created and webpack watch is working as expected
In Prod Mode:
1. some issue with application launching

Over all things seems to be working

ccm_web/package.json
ssciolla: Ah, yeah, I need to update the entrypoint in `Dockerfile`.

ssciolla: This should be okay now.

ssciolla:
> Optional: @ssciolla can we create a directory called `module` and put all respective modules into it like `lti`, 'user`, etc`. I saw that way of organizing pretty common around nestJs projects

Thanks for the suggestion. I'm not convinced yet, but maybe it would be nice if/when we have lots of modules. The creator seems to suggest [here](https://github.com/nestjs/nest/issues/2249#issuecomment-494734673) not to use a `modules` directory, but I imagine it's personal preference.

ssciolla:
> What are your thoughts on using built-in NestJS logger vs Winston?
> 
> NestJs [documentation](https://docs.nestjs.com/techniques/logger#use-external-logger) write for more robust prod logging tends to use external libraries. I don't have an opinion on this either asking for change. The setup you have seems to be good.

I kinda figured I'd look into this later; it may be nice to get logging messages to look consistent, but it didn't seem like a top priority. We'll need to set up request logging, I can think about this then.

ssciolla:
> > I could see eliminating LTIService, and let the middleware class include all of the ltijs code;
> 
> Yeah from what I understood from NestJS architecture design style, you create a service if you want to inject it to other modules/controllers. But I don't see this as injecting into other module. So the code inside could very well be inside the Middleware file.
> 
> It's ok if you want to deal with this as separate issue

With some of the refactoring I did on Friday/today, I'm not sure how to handle this. I will start taking a look at it, but I think there are some advantages to using an `LTIService`, particularly because the underlying `ltijs` implementation has a database underlying it and *feels* like a service. It may be beneficial later on to have access to the `Provider` instance programmatically, too. I will continue to fiddle with this, but I think I'd rather not make any other major changes after this PR is reviewed. While I think the `middleware` directory at the root seems to be a common pattern, this `lti.middleware.ts` is applied at the `LtiModule` level, so it feels like it should live here (at least I think so).

ccm_web/package.json
ssciolla: Looking at this now. May be also able to remove `schematics`

ccm_web/server/test/app.e2e-spec.ts
```diff
@@ -0,0 +1,37 @@
+import { Test, TestingModule } from '@nestjs/testing'
+import { INestApplication } from '@nestjs/common'
+import request from 'supertest'
+import { AppModule } from './../src/app.module'
```
pushyamig, line 4: There is no AppModule, this code seems to be unused but the import should be corrected atleast for confusion later

pushyamig: OK, I got confused. What is the purpose of the BASE module? to me right now it serves the purpose for API routing. Can we rename it to API?

ccm_web/server/test/app.e2e-spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+
+  it('/api/globals (GET)', async () => {
+    return await request(app.getHttpServer())
+      .get('/api/globals')
+      .expect(401)
+      .expect(
+        {
+          status: 401,
+          error: 'Unauthorized',
```
ssciolla, line 25: This is kind of a silly test, but I wanted to demonstrate the use of these tools, and testing whether the LTI middleware is blocking unauthenticated users has some benefit. I (or someone) probably needs to do more research on how we'd actually be able to make authenticated requests.

ccm_web/server/test/app.e2e-spec.ts
ssciolla: I can re-name it to API if you'd like. Calling it an API when we'll likely have other API modules seemed kinda strange, but I'll change it.

ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/68/commits/dbd80aedc2698eb5e67a1b70d0762e372c9cf2e9

lsloan:
_@ssciolla [wrote](#issuecomment-833687995):_
> > Midddleware → Middleware with extra middle. 😉
> 
> @lsloan, yep, do you have a suggestion about this? We could probably move the work of `LTIService` into the middleware.

I didn't have that in mind, but that sounds like a good idea.  Modularizing the setup code for LTI support would make it reusable for other projects.  Would it have any immediate benefit for this project?

lsloan:
I was just catching up on comments left here Friday afternoon.  I may have misunderstood your earlier question for me about LTI middleware.

So, you're suggesting removing the `LTIService` code here and let `ltijs` serve that purpose?

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    this.provider = provider
+  }
+
+  getMiddleware (): Express | undefined {
+    return this.provider?.app
+  }
+
+  beforeApplicationShutdown (): void {
+    this.provider?.close()
```
pushyamig, line 73: Can you explain what is going on here?

ccm_web/server/test/app.e2e-spec.ts
pushyamig: OK, I see what you mean. Can we have all the other API modules like merge section, create bulk user under this?

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import baseLogger from '../logger'
+import { DatabaseConfig, LTIConfig } from '../config'
+
+const logger = baseLogger.child({ filePath: __filename })
+
+// ltijs docs: https://cvmcosta.me/ltijs/#/
+@Injectable()
+export class LTIService implements BeforeApplicationShutdown {
+  provider: LTIProvider | undefined
```
pushyamig, line 15: LTI Service and LTI Middleware seems to be more or less have a Inheritance pattern. LTI middleware you seems to define the function and LTI Service is filling in the details of it. Can you direct me to a documentation about your approach? I don't understand it fully why are you doing it this way.

ccm_web/server/src/lti/lti.module.ts
```diff
@@ -0,0 +1,28 @@
...
+import { LTIMiddleware } from './lti.middleware'
+import { LTIService } from './lti.service'
+
+@Module({
+  imports: [ConfigModule],
+  providers: [
+  // https://docs.nestjs.com/fundamentals/custom-providers
+    {
+      provide: LTIService,
+      inject: [ConfigService],
```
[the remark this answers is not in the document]
ssciolla: @pushyamig, take a look here: https://docs.nestjs.com/techniques/configuration#use-module-globally 

It sounds like, once you have configured it `AppModule`, you can import `ConfigModule` as needed to any modules **or** you can make it global. I guess I wasn't sure if all the modules would actually need configuration, so I opted not to go the global route. I can change that though if you want.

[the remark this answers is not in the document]
ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/68/commits/cb3a28b2de223e9d7cb5735a78f42fdd7de44903

It's not a huge change. You will still need to import `ConfigService` and inject it as a dependency through the constructor of whatever class you're using it in (which I think is what you were saying). I believe you only need to configure the `ConfigModule` once, and then it's good to go across the application.

ccm_web/server/src/lti/lti.service.ts
ssciolla: Hm, okay. I still think you and @lsloan had a question about whether we could coalesce `LTIService` and `LTIMiddleware`. Short answer is I don't know. We need to run an `async` method on instance creation, and I'm not sure if this can be done with middleware. Most of what I found online suggests using a dynamic service (through a factory method).  See https://stackoverflow.com/questions/56239773/nestjs-how-to-configure-middleware-with-async-await This is essentially what I'm doing now in `lti.module.ts` with the custom provider there.

There is that other project from Texas that uses Nest and `litjs`, and they are using `onModuleInit` (another lifecycle hook) to set up the instance when the module is loaded. However, when I tried that, I couldn't also use the `ConfigService` (which I need to set up the `ltijs` `Provider`. https://github.com/txstate-etc/attendance-node/blob/main/api/src/lti/lti.middleware.ts

I think the current arrangement seems to work, so my inclination would be not to change it too much -- unless there is some shortcoming of this I'm not seeing.

ccm_web/server/src/lti/lti.module.ts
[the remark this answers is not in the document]
pushyamig: The config Stuff should be available to entire project and accessible when needed. That's my idea. The global config fulfills that way

ccm_web/server/src/lti/lti.service.ts
pushyamig: thanks for more context. I will think about it more

pushyamig (approved):
This seems to be functional and working. The way the LTIJS middleware is working is serving 2 purposes. 
1. Doing the LTI launch stuff
2. the JWT LTIJS created validating if request has it or not.
Ideally after the launch LTIMiddleware should do much more and we could handle the JWT token authorization ourself. This would be one reason to do this way since we are not doing any LTI advantage feature.

Now every request passes through LTI middleware which is ok as it does impose security restriction but could worked otherwise. This is not something NestJS imposed but it definitely showed us ways how we can do this better

Nice work Sam and quickly putting this all together!

PR #83, opened 2021-05-18
Add OpenAPI/Swagger (#82)

### LINKED ISSUE #82: Add OpenAPI/Swagger implementation
OpenAPI and Swagger are common tools for documenting and allowing users to try out custom-built server API endpoints. Nest provides a library and patterns for adding this easily to an existing Nest application. We would need to also figure out how to access the UI created by Swagger while using the `ltijs` authentication.

- https://docs.nestjs.com/openapi/introduction
- https://swagger.io/

ssciolla:
This PR adds an OpenAPI API specification generator and a Swagger UI interface to the project, following the patterns laid out [here](https://docs.nestjs.com/openapi/introduction). Currently, it is only supported in an LTI context (same as the application). The code is only invoked in development. The PR resolves #82.

Note(s): 
* I looked into loading the necessary modules conditionally (so they can only be `devDependencies`), but because decorators are/will be used throughout to generate the spec, this seemed difficult. See https://github.com/nestjs/swagger/issues/184

### DIFF
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -92,6 +92,10 @@ function App (props: AppProps): JSX.Element {
           </Switch>
         </Router>
         <ConsumerTest ltiKey={props.ltiKey} />
+        {
+          globals?.environment === 'development' && props.ltiKey !== undefined &&
+            <Link href={`/swagger?token=${props.ltiKey}`}>Swagger UI</Link>
+        }
       </SnackbarProvider>
     </div>
   )
--- a/ccm_web/server/src/api/api.controller.ts
+++ b/ccm_web/server/src/api/api.controller.ts
@@ -1,8 +1,10 @@
 import { Controller, Get } from '@nestjs/common'
+import { ApiBearerAuth } from '@nestjs/swagger'
 
 import { HelloData, Globals } from './api.interfaces'
 import { APIService } from './api.service'
 
+@ApiBearerAuth()
 @Controller('api')
 export class APIController {
   constructor (private readonly apiService: APIService) {}
--- a/README.md
+++ b/README.md
@@ -190,6 +190,24 @@ for generating module and other code stubs and performing other tasks.
 As the necessary libraries are not included in `package.json`, follow the documentation's suggestion
 and install `@nestjs/cli` globally on your machine if you want to use the tool.
 
+#### Using OpenAPI and Swagger
+
+When in development mode (i.e. the `NODE_ENV` environment variable's value is `development`),
+the application generates [OpenAPI](https://www.openapis.org/) documentation and an explorer for
+the API endpoints using recommended [Nest patterns](https://docs.nestjs.com/openapi/introduction) and
+[Swagger](https://swagger.io/).
+
+To access and use the Swagger UI, launch the application from a course in Canvas,
+then click on the "Swagger UI" link in the application interface's footer.
+This will take you to the Swagger page where you can view the documented API endpoints.
+
+However, to execute requests using Swagger, you need to authenticate using a Bearer token.
+The token is currently made available as a URL parameter called `token`, which you can see and copy
+by using a browser tool to view the frame's source.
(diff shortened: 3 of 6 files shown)


ccm_web/server/src/api/api.controller.ts
```diff
@@ -1,8 +1,10 @@
 import { Controller, Get } from '@nestjs/common'
+import { ApiBearerAuth } from '@nestjs/swagger'
 
 import { HelloData, Globals } from './api.interfaces'
 import { APIService } from './api.service'
 
+@ApiBearerAuth()
```
pushyamig, line 7: This seems to be doing the magic of recognizing if the path is from swagger and checking for token and passing or un passing. 

I commented out that line and try to do an api call from swagger throws error token required nicely! But uncommenting that line and making the api call gave me error as well. I tried to do an LTI launch again, picked the new  token etc. I had to restart the server so that controller recognize  `ApiBearerAuth`. This is ok, Not an issue. Just stating what I noticed.

ccm_web/client/src/App.tsx
```diff
@@ -92,6 +92,10 @@ function App (props: AppProps): JSX.Element {
           </Switch>
         </Router>
         <ConsumerTest ltiKey={props.ltiKey} />
+        {
+          globals?.environment === 'development' && props.ltiKey !== undefined &&
+            <Link href={`/swagger?token=${props.ltiKey}`}>Swagger UI</Link>
```
pushyamig, line 97: This is smart. Somewhere down the lane we need to have a footer and place this URL in it. @chrisrrowland what do you think?

pushyamig (approved):

ccm_web/server/src/api/api.controller.ts
ssciolla: Thanks for looking closely at all this. My understanding is that the creation of the OpenAPI document relies a lot decorators, both those that are part of Nest and additional ones created for OpenAPI purposes. See [Nest - OpenAPI - Types and Parameters](https://docs.nestjs.com/openapi/types-and-parameters) and [Nest - OpenAPI - Decorators](https://docs.nestjs.com/openapi/decorators).

ccm_web/client/src/App.tsx
chrisrrowland: Are you hinting at something different than what @ssciolla  has already done here?  I guess you mean a more formal footer?

We could, especially if we want a footer for other purposes.  Otherwise I also think it's fine just tagging it on like this in development.

ssciolla: We will probably want a footer to put a copyright message and maybe version information, and this link could live there. It may also just be a good UI design thing, lets people know they're at the end of the page and prevents important content from crowding up near the window edge. Separate issue though.

ccm_web/server/src/api/api.controller.ts
```diff
@@ -1,8 +1,10 @@
...
 @Controller('api')
```
chrisrrowland, line 8: Should the ConsumerTest component be using this?

ssciolla: I'm not sure what you mean, can you explain? Do you think `api/hello` shouldn't be behind the authentication? Since we're installing `ltijs` as global middleware, I'm not sure we have that level of control. Also, this decorator is just telling the Swagger UI what authentication is needed to use the endpoints, it doesn't actually control it.

chrisrrowland: I was thinking that all API calls would go through the API Controller.  That's not the case?

I didn't think about it deeply, that is just how I have  done things in the past in my own code, where nobody is working with me to tell me I'm doing it wrong :-D

ssciolla: Hm, well, the `ConsumerTest` API calls the `hello` endpoint that is defined here, right? So it is going through this.

chrisrrowland: You're right, the way I was searching the code I missed the `fetch('/api/hello', params)`

PR #94, opened 2021-06-03
Add `user` tableName to decorator (#93)

### LINKED ISSUE #93: Failing to insert user records when using capitalized `User` table name in some cases
@chrisrrowland reported an issue where the application failed to insert user records upon LTI launch using the table name `User` (the actual table name is `user`). @pushyamig and I could not reproduce the issue. However, it seems specifying `tableName: 'name'` in `user.model.ts` will resolve this for him (and doesn't affect me). I'll create an PR.

We should keep an eye on this problem, as we still don't know why we experienced different things. Using `tableName` in the `@Table` decorator is probably something we should do in all model definitions.

ssciolla:
This PR aims to resolve #93.

### DIFF
--- a/ccm_web/server/src/user/user.model.ts
+++ b/ccm_web/server/src/user/user.model.ts
@@ -18,6 +18,7 @@ interface UserCreationAttributes extends Optional<UserAttributes, 'id'> {
 
 // Sequelize Datatypes mapping with Mysql https://sequelize.org/master/manual/model-basics.html#data-types
 @Table({
+  tableName: 'user',
   freezeTableName: true
 })
 export class User extends Model<UserAttributes, UserCreationAttributes> {


ccm_web/server/src/user/user.model.ts
```diff
@@ -18,6 +18,7 @@ interface UserCreationAttributes extends Optional<UserAttributes, 'id'> {
 
 // Sequelize Datatypes mapping with Mysql https://sequelize.org/master/manual/model-basics.html#data-types
 @Table({
+  tableName: 'user',
```
pushyamig, line 21: I agree this is better way to do it. But @chrisrrowland should validate it and approve

chrisrrowland: This works for me.

chrisrrowland: Should put the properties in alphabetical order.


Hahahah just kidding.  I don't care.

chrisrrowland (approved):

pushyamig:
test passes. since now all tables name are added in the respective models file for user and canvas_tokens

Issue #100, opened 2021-06-08
Refresh token workflow when token expires after 1 hr


pushyamig:

pushyamig:
@melindakraft Can I mark this as done as well?

Issue #104, opened 2021-06-09
Handle case of when scopes for Developer Key change, but users already have token


ssciolla:
As the project evolves, we may need to change the scopes associated with the API Developer Key used to create in the OAuth workflow (implemented in pending PR #92). However, the tokens we've stored in the database for users will still be associated with an older set of scopes. How can we implement a change management strategy such that if the scopes change, tokens are invalidated and/or the OAuth flow is re-initiated?

Idea(s):
- Add management process to remove all Canvas tokens when scopes change
- Re-initiate the OAuth flow for a user if they encounter an error related to scopes not matching.

pushyamig:
This is the last issue remaining on OAuth process. We can either bulk delete token manually during deployment when scopes do changes. But if the scopes were not selected by mistake during the configuration ( by mistake) then we will endup deleting all the scope if want to handle it programmatically. 

I am fine with manual approach for  first release. We can discuss on this

pushyamig:
Testing is needed.

pushyamig:
As Prep before, 
1. I removed one of the API like `'url:GET|/api/v1/accounts', from developer console api scopes and canvas.scope.ts.
2. I deleted the DB that was created before
3. Did the LTI launch, Launch worked, token generated.
After step
1. Enabled  `'url:GET|/api/v1/accounts',` API from the developer Key console
2. added these API’s tp canvas.scopes.ts file.
3. Restarted the server

Results: LTI launch happened, when accessing that specific API in merge section, I saw the authorization page came up and new token is generated. So things worked as expected. 

I consider this as Pass. @ssciolla

ssciolla:
Question about the first three steps: So when you remove a scope, but don't touch the database or existing tokens, what happens? Does the old token still work?

pushyamig:
Resolved.

ssciolla:
I'll QA this.

ssciolla:
I think this issue is resolved by the handling of the `InvalidTokenInterceptor`.

Here is how I tested.

Scenario 1:
1) Removed the PUT `courses` scope from both the `canvas.scopes.ts` file and the API Developer Key.
2) Tried changing the course name with the old token still in the database
3) Observed error and re-authorization flow was triggered. (This doesn't help the user fix this problem, since we're not granting this scope anymore through the OAuth, but we also shouldn't be leaving API calls that require the scope we removed).

Scenario 2 (picking up after S1):
[...]

PR #121, opened 2021-06-14
Implement Canvas requestor pattern, course name backend API routes (#99)

### LINKED ISSUE #99: Make the canvas API call with the generated  token

ssciolla:
This PR integrates a Canvas requestor library from KTH and introduces the first use of the Canavs OAuth scoped tokens (see #92) in order to support course name fetching and re-naming. It also incorporates types for the KTH library, NestJS API endpoint validation, and some Canvas API  request error handling. The PR aims to resolve #99 and works toward the resolution of #26 (we'll still need to integrate with the client and introduce some role handling).

Notes:
* **You will need to add the `GET` course scope to your API Developer Key.**
* Because we have not yet implemented a refresh method yet (I'm working on it, see #100), you may need to remove the `canvasToken` record for your user manually from the database an hour after initially launching.
* The `got` library (which the `@kth/canvas-api` thinly wraps) includes retrying (which is pretty neat!). Default seems to be `2` attempts.
* The `@kth/canvas-api` types do not declare all defined methods, only the public ones that comprise the API surface.

Resources:
* `got`: https://github.com/sindresorhus/got
* `@kth/canvas-api`: https://www.npmjs.com/package/@kth/canvas-api
* NestJS Controllers: https://docs.nestjs.com/controllers
* NestJS Validation - https://docs.nestjs.com/techniques/validation#validation

@jonespm, not asking for your review explicitly, but feel free to take a look if you'd like. Thanks for the research and getting started on the KTH  stuff.


ccm_web/server/src/api/api.interfaces.ts
```diff
@@ -10,3 +10,8 @@ export interface Globals {
     roles: string[]
   }
 }
+
+export interface Course {
+  id: number
+  name: string
+}
```
ssciolla, line 17: This is just a minimal version of the Canvas Course entity: https://canvas.instructure.com/doc/api/courses.html

I could maybe see moving this to the `canvas` module.

pushyamig:
@ssciolla I see this error when I run you code `{"statusCode":401,"message":"You are not authorized to access this resource.","error":"Unauthorized"}` otherwise the version of code in the `main`branch work fine. I double checked for any config changes I don't see you have any new.

ssciolla:
> @ssciolla I see this error when I run you code `{"statusCode":401,"message":"You are not authorized to access this resource.","error":"Unauthorized"}` otherwise the version of code in the `main`branch work fine. I double checked for any config changes I don't see you have any new.

I talked with @pushyamig, the API Developer Key scopes needed to be updated. I've put a note in bold in the PR description.

ccm_web/server/src/api/api.controller.ts
```diff
@@ -19,4 +22,24 @@ export class APIController {
...
+  async getCourseName (
+    @Param('id', ParseIntPipe) courseId: number, @Session() session: SessionData
+  ): Promise<string> {
+    const { userLoginId } = session.data
+    const result = await this.apiService.getCourseName(userLoginId, courseId)
+    if (result === null) throw new InternalServerErrorException('Error occurred while communicating with Canvas')
+    return result
+  }
+
+  @Post('course/:id/name')
```
pushyamig, line 36: Updating a courseName is actually a PUT call, can we change it with same HTTP method?

ssciolla: Sure, `PUT` does seem more correct

ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/2ee56331c71ff1d82bad4548ddf9a4f73fd9626c

ccm_web/server/src/canvas/canvas.scopes.ts
```diff
@@ -7,6 +7,7 @@ can be found in the API Developer Key interface in Canvas.
 
 const privilegeLevelOneScopes = [
   // Courses
+  'url:GET|/api/v1/courses/:id',
```
pushyamig, line 10: I missed look at the API format of scopes and I ended up checking this `url:GET|/api/v1/courses/` scope from the API key console

I see that is there is an error due to miss matched scope then `query` object result is as below.
```
{
  error: "invalid_scope",
  error_description: "A requested scope is invalid, unknown, malformed, or exceeds the scope granted by the resource owner. The following scopes were requested, but not granted: url:GET|/api/v1/courses/:id",
}
```

I feel we need to check for `query.error` scenario and error out that will quickly fixes the configuration error which are initial to development and deployment. Currently the error in the log is this which is miss leading and make me think something wrong with session again.

`2021-06-15T18:42:48.277Z - canvas.controller.ts - warn - State variable returned from Canvas did not match session ID; throwing unauthorized exception...`

Add a case some thing like this in the `canvas.controller.ts`
```
if (query.error) {
      logger.info(`${error} ${error_description} `)
      throw new UnauthorizedException('Please verify your Canvas API configuration')
    }
```

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,44 @@ export class APIService {
...
+    }
+    return name
+  }
+
+  async putCourseName (userLoginId: string, courseId: number, newName: string): Promise<string | null> {
+    const requestor = await this.canvasService.createRequestorForUser(userLoginId)
+    let name = null
+    try {
+      const response = await requestor.requestUrl<Course>(
+        `courses/${courseId}`, 'PUT', { course: { name: newName } }
```
pushyamig, line 57: One of things I heard from Stakeholders is when they change a course name they also change the course_code as well. So please include update of `course[course_code]` in the API call when updating. Course_code is same as course name. You can verify it with any Canvas lecture courses course_name and course_code is the same. I think backend can simply update the course_code when they change the course_name. 
I don't see a point to have course_code variable in the swagger API but I might be wrong. So I will leave it up to you.

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,44 @@ export class APIService {
       }
     }
   }
+
+  async getCourseName (userLoginId: string, courseId: number): Promise<string | null> {
```
pushyamig, line 33: What is the reason for having an API call for fetching the course name?

ccm_web/server/src/canvas/canvas.scopes.ts
ssciolla: I'm a little confused. Can you provide steps to reproduce this?

Could this be a separate issue?

ccm_web/server/src/api/api.service.ts
ssciolla: We need to know the course name so it can be displayed in the UI. Pulling the course name from the launch variables is not ideal because once it's changed, we'd need to keep it updated separately in the session. I suppose we could do that, but this seemed okay as well.

ccm_web/server/src/api/api.service.ts
ssciolla: Is this request documented in the requirements somewhere? Did I miss it, or we adding it now?

pushyamig: I did document this in the issue https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/26.

ccm_web/server/src/canvas/canvas.scopes.ts
pushyamig: Try checking these 2 scopes in CANVAS API KEY console and launch CCM, you will notice this error. You can deal this as separate issue. 
Screen Shot 2021-06-15 at 3 13 03 PM

ccm_web/server/src/api/api.service.ts
ssciolla: You are correct, I will add that.

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,44 @@ export class APIService {
...
+
+  async putCourseName (userLoginId: string, courseId: number, newName: string): Promise<string | null> {
+    const requestor = await this.canvasService.createRequestorForUser(userLoginId)
+    let name = null
+    try {
+      const response = await requestor.requestUrl<Course>(
+        `courses/${courseId}`, 'PUT', { course: { name: newName } }
+      )
+      const course = response.body
+      name = course.name
```
pushyamig, line 60: @chrisrrowland what would you like to see in the response when course rename is successful? Currently this is sending the rename course name?

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,44 @@ export class APIService {
       }
     }
   }
+
+  async getCourseName (userLoginId: string, courseId: number): Promise<string | null> {
+    const requestor = await this.canvasService.createRequestorForUser(userLoginId)
+    let name = null
+    try {
+      const response = await requestor.get<Course>(`courses/${courseId}`)
```
pushyamig, line 37: is the library is throwing an error if the error is other than `2xx`?

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,44 @@ export class APIService {
...
+      const course = response.body
+      name = course.name
+    } catch (error) {
+      if (error instanceof HTTPError && error.response.statusCode !== HttpStatus.OK) {
+        logger.error(`Received unusual status code ${String(error.response.statusCode)}`)
+        logger.error(`Response body: ${JSON.stringify(error.response.body)}`)
+      } else {
+        logger.error(`An error occurred while making a request to Canvas: ${JSON.stringify(error)}`)
+      }
+      throw new Error(error)
```
pushyamig, line 47: when a Non `2xx` case do occur then should you be sending the error status code and message to FE instead of generalizing it. Since it would a value piece of info why the failure has occurred to user. When I tested this from Swagger this is what I see, I don't error is return back to controller.
**Error message which is valuable to send**
Screen Shot 2021-06-15 at 4 49 38 PM

**Swagger console getting the error**
Screen Shot 2021-06-15 at 4 52 57 PM

@chrisrrowland what do you preference,  since you might want to deal with API failures and message. So you have better vision on how this should be done

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,44 @@ export class APIService {
...
```
pushyamig, line 68: You have a throw statement here and that will raises `api.controller.ts` and I feel that should  have try-catch block as well. I don't think `result==null` will be hit when you throw an error. 


```
 const result = await this.apiService.putCourseName(userLoginId, courseId, courseNameDto.newName)
    if (result === null) throw new InternalServerErrorException('Error occurred while communicating with Canvas')
    return result
```

I really couldn't test this from swagger UI since from the UI I received different error message

Screen Shot 2021-06-15 at 4 52 57 PM

I really don't see the message you are wording to send is the response. So I don't know what going on here

ssciolla: Yeah, I think you're right, we don't want to throw the error here, just return null. I'll fix that.

pushyamig:
@chrisrrowland and @ssciolla This PR is taking care of creating the endpoint and we could quickly test it from Swagger. I think there is an issue needs to be created for using the course rename endpoint and integrating with React. I created the issue here https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/122. 

To make this feature complete that would be last bit of work needs to be done. I think swagger is really a good way to share the API endpoint very useful and saves time instead of FE and BE dev talking Swagger is taking care of this.

ccm_web/server/src/api/api.service.ts
ssciolla: I could do this, I guess I was trying to keep the error handling minimal for this first pass. But I could see returning an error object instead of `null` from the `APIService` functions, something like this:

```
{
  statusCode: 401,
  message: 'You are not authorized to modify this resource."
}

File: ccm_web/server/src/api/api.service.ts (hunk 9 above)
Comment (ssciolla) [reply ccm_web/server/src/api/api.service.ts:68]: Sorry, the error handling has been tricky, it's changed a few times.

File: ccm_web/server/src/canvas/canvas.scopes.ts (hunk 3 above)
Comment (ssciolla) [reply ccm_web/server/src/canvas/canvas.scopes.ts:10]: So is this covered by #104?

File: ccm_web/server/src/api/api.service.ts (hunk 5 above)
Comment (pushyamig) [reply ccm_web/server/src/api/api.service.ts:33]: ok

File: ccm_web/server/src/api/api.service.ts (hunk 9 above)
Comment (ssciolla) [reply ccm_web/server/src/api/api.service.ts]: I removed the error throwing here (https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/7ccd58833f005014ad2a5be9af13543491eca96b), but I'm going to see why the specific error string isn't being used that I passed to the `Exception` in the controller.

Comment (ssciolla) [reply ccm_web/server/src/api/api.service.ts]: Ah, the message is now showing up now that I'm not throwing an error. This should be resolved. We can discuss more specific error messages in this thread: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121#discussion_r652144801

File: ccm_web/server/src/api/api.service.ts (hunk 4 above)
Comment (ssciolla) [reply ccm_web/server/src/api/api.service.ts]: Done. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/f54ba63ca5b22ee4d40c3d900dc1f91d71febc93

File: ccm_web/server/src/api/api.service.ts (hunk 7 above)
Comment (ssciolla) [reply ccm_web/server/src/api/api.service.ts:37]: Yes, but also `3xx` I guess. See https://github.com/sindresorhus/got#gothttperror

We can customize this if we want, using the `throwHTTPErrors` option: https://github.com/sindresorhus/got#throwhttperrors

Maybe I need to modify the unusual code check. Right now I'm using `error instanceof HTTPError && error.response.statusCode !== HttpStatus.OK`, but maybe just `error instanceof HTTPError` is sufficient.

File: ccm_web/server/src/canvas/canvas.service.ts [hunk 10]
```diff
@@ -108,4 +109,12 @@ export class CanvasService {
     const token = user.canvasToken === undefined ? null : user.canvasToken
     return token
   }
+
+  async createRequestorForUser (userLoginId: string): Promise<CanvasRequestor> {
+    const token = await this.findToken(userLoginId)
+    if (token === null) throw new Error(`User ${userLoginId} does not have a token!`)
+
+    const requestor = new CanvasRequestor(this.url + '/api/v1/', token.accessToken)
```
ssciolla, line 117: Anyone see a reason to not hardcode `/api/v1`, make it configurable?

ccm_web/server/src/api/api.service.ts
ssciolla: I decided to remove the check for the non-`200` status code; see here https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/d684eeb2fca3bd59040f59df42dcfda583dbd18d.

This might be a moot point, since I don't think we'll encounter `2XX` or `3XX` other than `200`, but I guess those are still technically `OK` response codes.

ccm_web/server/src/canvas/canvas.service.ts
jonespm: Currently all apis are on` /api/v1 ` **except** graphql which is just on `/api/graphql`. Canvas might eventually have some future API's on `/api/v2`, `/api/v3` etc if the API ever changed and they maintained the old API. 

You might just want to make that `/api/v1/` as a method parameter with this value as the default.
 
We removed this hardcoding in the canvasapi Python library recently.

ssciolla: Good idea, how is this, @jonespm? https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/39477d2334738f5618a6348a191729adb03d56cf

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,42 @@ export class APIService {
...
+        logger.error(`Received unusual status code ${String(error.response.statusCode)}`)
+        logger.error(`Response body: ${JSON.stringify(error.response.body)}`)
+      } else {
+        logger.error(`An error occurred while making a request to Canvas: ${JSON.stringify(error)}`)
+      }
+    }
+    return name
+  }
+
+  async putCourseName (userLoginId: string, courseId: number, newName: string): Promise<string | null> {
```
chrisrrowland, line 51: I think something useful to consider is incorporating a request id.
For example, `putCourseName `in `api.service.ts` can accept a [guid ](https://www.npmjs.com/package/guid-typescript)and `putCourseName` in `api.controller.ts` can generate one and forward it on to be used in logging.  You can use this info when trying to isolate a single request 'thread' when there is a lot of activity going on that would otherwise make it difficult to tell what is going on.

You can do stuff like 
`logger.log('putCourseName[ '+ requestId +'] userLoginId: ' + userLoginId +' courseId: ' + ... )`

Honestly I have no idea if there is a design pattern or best practice around this, I just started doing it in my own projects because it seemed useful.  Maybe there is a better way to implement it.

ccm_web/server/src/api/api.service.ts
ssciolla: Okay, what object do you want? The course object? Something like this? 
```
{
   id: XXXXX,
  name: "Some Course Name"
}
```

I'm returning the name because that provides some guarantee that the change was successful, and prevents the client from needing to fetch the updated name again using a `GET` call. Using what the client sent as the new name could obscure unforeseen transformations done by Canvas.

ccm_web/server/src/api/api.service.ts
chrisrrowland: The status code might be sufficient?  I'm not sure exactly what use cases there will be for telling the user what they should do when an error occurs.  If there is any error we can predict happening, it seems like we should have a better way to handle it than telling the user to act in some way.

However the actual translation of 401 to 'You are not authorized to modify this resource.' does seem like something that would come from the front end, as that is where any localizing would happen.. I think?  (Not that we are doing that)
I haven't actually worked on a web application with localized text so I'm not sure.

ccm_web/server/src/api/api.service.ts
chrisrrowland: Yeah I see that you're returning the name as returned by the request, I was just saying I couldn't see how the result of that transaction would be any different than the input, unless it was manipulating the input text in some way.   But I think it makes sense to return the value the way you are, since that would be safe either way.

I have gotten in the habit of returning things like this (I haven't done it in typescript so not exactly but you get the gist)
```
interface PutCourseNameResponse {
   name: string | undefined
   status: number
}
```
Frankly I've done a lot of development in isolation so I don't know how many of my personal design patterns are good or not.
I end up with a lot of different response type objects (which do accurately reflect the quantity of unique responses) which can add a bit of code..

pushyamig: This is my thinking
1. If canvas is sending error messages/status code, I feel the Backend should be a passthrough in those cases
2. Some internal error (say making DB insert) failed after Canvas call, in that case the BE should wrap around the error message and status code = 500 and some custom message
3. with 200 Message I agree with @ssciolla that we don't want to send entire Object but pick what we want, and I definitely agree with @chrisrrowland that it should always be sent as Json Object. We can define case by case bases when to send the success object as is or tailor it by not sending too much information

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,42 @@ export class APIService {
...
+    const requestor = await this.canvasService.createRequestorForUser(userLoginId)
+    let name = null
+    try {
+      const response = await requestor.get<Course>(`courses/${courseId}`)
+      const course = response.body
+      name = course.name
+    } catch (error) {
+      if (error instanceof HTTPError) {
+        logger.error(`Received unusual status code ${String(error.response.statusCode)}`)
+        logger.error(`Response body: ${JSON.stringify(error.response.body)}`)
```
pushyamig, line 43: `error.response.body` it is redundant to log this since it has HTML content.  `error.response.statusCode` and `error.response.statusMessage` what you want to log and send the user message

Screen Shot 2021-06-16 at 3 44 21 PM

You need to have return statement in the `catch` block to return to controller the message/status_code you want to send in
`{statusCode: 422, statusMessage: 'Unprocessable Entity' }`

currently still null is return and it endup as 500 error

Event with 200 case you want to send message_object/status code 
`{statusCode: 200, statusMessage: {couseName: 'course_update'}}`

Similar strategy should be applied to put/get calls for coursenames

ccm_web/server/src/api/api.service.ts
```diff
@@ -21,4 +29,42 @@ export class APIService {
...
+      } else {
+        logger.error(`An error occurred while making a request to Canvas: ${JSON.stringify(error)}`)
```
pushyamig, line 45: I feel you should log `error.message` and `error.code` if not `HTTPError`. Incase of undefined status code make it 500
Screen Shot 2021-06-16 at 3 51 05 PM

ccm_web/server/src/api/api.controller.ts
```diff
@@ -19,4 +22,24 @@ export class APIController {
...
+  }
+
+  @Put('course/:id/name')
+  async putCourseName (
+    @Param('id', ParseIntPipe) courseId: number, @Body() courseNameDto: CourseNameDto, @Session() session: SessionData
+  ): Promise<string> {
+    const { userLoginId } = session.data
+    const result = await this.apiService.putCourseName(userLoginId, courseId, courseNameDto.newName)
+    if (result === null) throw new InternalServerErrorException('Error occurred while communicating with Canvas')
+    return result
```
pushyamig, line 43: based on my comments [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/files?file-filters%5B%5D=.ts#r652996449)

I feel you should send a response object something like this 
` res.status(result.statusCode).json(result.statusMessage)`

This way the FE will identify error state and normal state. Otherwise it will end up in 200 even there is an error

ssciolla:
> I feel like it would make more sense to not have a default value for endpoint here.
> Make the caller provide that value.
> If a new version of an endpoint is published it's likely not compatible with the old one, so the calling code would have to change anyways. I'm having trouble envisioning a scenario where you don't either end up changing the default, and changing the code that is calling it anyways, or changing all the calling code and the default ends up being this defunct value.

We could change the default if/when the main API endpoint changes. I guess I would rather avoid repeating `/api/v1/` in ten different places? I could follow arguments either way.

ccm_web/server/src/api/api.controller.ts
ssciolla: Okay, I will try to refactor a bit, add more return information. NestJS supplies some abstraction so you don't need to Express pattern you suggested in these type of endpoints, but I'll see if it makes sense to override them.

ccm_web/server/src/api/api.service.ts
ssciolla: Yeah, so the error object is `any` if it's not an `HTTPError`, so I'd have to check whether `message` and `code` are defined (and they may not be). This is why I'm just stringifying and logging the object. We can return a `500` here to the controller as well, but I guess I'd prefer this more generic logging than a bunch of conditionals checking for defined properties.

chrisrrowland:
> > I feel like it would make more sense to not have a default value for endpoint here.
> > Make the caller provide that value.
> > If a new version of an endpoint is published it's likely not compatible with the old one, so the calling code would have to change anyways. I'm having trouble envisioning a scenario where you don't either end up changing the default, and changing the code that is calling it anyways, or changing all the calling code and the default ends up being this defunct value.
> 
> We could change the default if/when the main API endpoint changes. I guess I would rather avoid repeating `/api/v1/` in ten different places? I could follow arguments either way.

I'm basing this on my (limited) experience with 3rd party API.  Not every endpoint will exist on each version.  A new version may simply add new endpoints, or it might improves an old one, but it doesn't have to replace the old version.  So our call to /v1/DoTheThing might not exist in /v2 and then might have a new version /v3/DoTheThing that is incompatible with v1, while some new feature wanted to take advantage of was added with v2.  In any of those cases you can't just change the version globally without having side effects.

ccm_web/server/src/api/api.service.ts
ssciolla: Let me think about this some more. Do we really need the status code if it's `200`? 

I'm also a little confused why the body is redundant. That's where Canvas puts its specific error messages. `statusMessage` seems like a separate piece of information, probably giving a generic human-readable description for the error code, right?

ccm_web/server/src/api/api.service.ts
pushyamig: Not just for logging I also want same think response object that we send to the user as well. We should always send the error responses sent from the upstream source.

ccm_web/server/src/api/api.service.ts
pushyamig: Probably not, I think it will defaults to 200. If an API call return other than 200 like 201 then I feel we should return the same response code, but in this case it 200 so no need

ccm_web/server/src/api/api.service.ts
ssciolla: So if it's not an `HTTPError`, it's likely one of these other ones (https://github.com/sindresorhus/got#gotrequesterror), or an error thrown by our application. Those won't necessarily have status codes associated with them (as they aren't HTTP errors) and may not be related to Canvas.

ssciolla:
> I'm basing this on my (limited) experience with 3rd party API. Not every endpoint will exist on each version. A new version may simply add new endpoints, or it might improve upon an old one, but it doesn't have to replace the old version. So our call to /v1/DoTheThing might not exist in /v2 and then might have a new version /v3/DoTheThing that is incompatible with v1, while some new feature wanted to take advantage of was added with v2. In any of those cases you can't just change the version globally without having side effects.

@chrisrrowland, okay, you make valid points, I will remove the default.

ccm_web/server/src/api/api.interfaces.ts
ssciolla: I did end up doing this. I don't really know if it was correct, but I'm trying to keep very-Canvas specific stuff in that module.

ccm_web/server/src/api/api.service.ts
ssciolla: Okay, see latest commits, I think I have made changes incorporating all the feedback.

ccm_web/server/src/api/api.service.ts
ssciolla: Okay, as requested, I'm passing through Canvas error codes and messages now. I've added some parsing code for the Canvas error body shape (which doesn't seem to be well documented anywhere :( )

ccm_web/server/src/api/api.service.ts
ssciolla: In the latest commits, I decided not to include the status code in successful responses. It should already be present and correct.

ccm_web/server/src/api/api.service.ts
ssciolla: Please see latest commits. I am returning a 500 error if it's not an `HTTPError` with a generic message. I don't anticipate this coming up very often, and we can assess the reporting strategy if/when we start seeing those errors.

ccm_web/server/src/api/api.controller.ts
ssciolla: I decided to use the NestJS `HTTPException` approach. See latest commits, and https://docs.nestjs.com/exception-filters#throwing-standard-exceptions

ccm_web/server/src/api/api.service.ts
ssciolla: I like this idea, it does seem useful. I don't really like adding more parameters that will need to be present in all API service methods, though maybe it's unavoidable if you want to trace a connection between the method calls. @chrisrrowland, are you alright with me looking into this separately? I can create an issue based on your comment. In the meantime, I'm going to look into whether I can add any more `debug` level logging that might be useful.

ccm_web/server/src/api/api.interfaces.ts
```diff
@@ -10,3 +10,13 @@ export interface Globals {
     roles: string[]
   }
 }
+
+export interface APIErrorData {
+  statusCode: number
+  message: string
+}
+
+export function isAPIErrorData (v: unknown): v is APIErrorData {
```
pushyamig, line 19: `v` can have better name like `response` or `httpStatus`

ssciolla: I could change it to `value`? I don't want to change it to one of the ones you suggested because we don't yet know what the value is.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -15,3 +17,33 @@ export interface TokenResponseBody {
...
+export interface CanvasErrorBody {
+  errors: CanvasError[]
+}
+
+export function isCanvasErrorBody (v: unknown): v is CanvasErrorBody {
+  if (!hasKey(v, 'errors') || !Array.isArray(v.errors)) {
+    return false
+  } else {
+    const result = v.errors.map(e => isCanvasError(e)).every(e => e)
+    return result
```
pushyamig, line 47: `v` can have better name

ccm_web/server/src/api/api.interfaces.ts
pushyamig: Ok `value` would be better name than `v`

ccm_web/server/src/canvas/canvas.interfaces.ts
ssciolla: See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/048dd661b3e7e57dea83b320c8bce535374af213

ccm_web/server/src/api/api.interfaces.ts
ssciolla: See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/048dd661b3e7e57dea83b320c8bce535374af213

ccm_web/server/src/api/api.service.ts
ssciolla: See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/121/commits/866b34e1dd01989f7d085a5adb3a59067ef549d2 for some debug logging.

ccm_web/server/src/api/api.interfaces.ts
```diff
@@ -10,3 +12,12 @@ export interface Globals {
...
   }
 }
+
+export interface APIErrorData {
+  statusCode: number
+  message: string
+}
+
+export function isAPIErrorData (value: unknown): value is APIErrorData {
+  return hasKey(value, 'statusCode') && hasKey(value, 'message')
```
ssciolla, line 22: There's some redundancy here because `hasKey` will check twice whether it's not null and an object, but seemed somewhat harmless and better to use the custom defined method than do it another idiosyncratic way. Maybe that's wrong though, IDK

ccm_web/server/src/api/api.service.ts
chrisrrowland: That's fine

ssciolla: Great, see https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/128

pushyamig:
I am seeing some linting errors, I don't know if this is just me. I did pulled latest code from your branch

Screen Shot 2021-06-17 at 4 13 23 PM
Screen Shot 2021-06-17 at 4 13 31 PM
Screen Shot 2021-06-17 at 4 13 49 PM

pushyamig:
1. Tested the course rename feature by becoming a **Subaccount admin, teacher, and admin** in a course 484127

	  **Subaccount/Root admin**: They could rename the courses
	  **Teacher**: In reality the feature would be disabled from UI but I wanted to verify the scoped token phenomenon and it **did not allow** for course rename but canvas did not error out either with `4xx`. It simply send 200 OK returning the course object and in the object we can say that course was not renamed. This is strange. I wanted to verified this API call using Postman then I confirmed that this is Canvas Phenomenon not KTH library or CCM backend code.

2. Tested the get course info API call with **Teacher, Subaccountadmin, admin, TA, Assistant** roles all seems to fetch the course info. 
	  I believe that course info a kind of public info to people added to course or admin this API will work for all. I did not want to be exhaustive and going by my believe. May be I am thinking too much here. 

3. I felt swagger was really helpful when testing all the API calls before the UI wired up. Only annoying thing is i need to copy/paste the token. I wish we had automatic way of doing this. Can we whitelist the swagger API route?

Overall I felt good, I did not pay close attention to KTH library details but I will know more while I am working on adding a create sections BE endpoint

pushyamig (approved):

ccm_web/server/src/canvas/canvas.scopes.ts
ssciolla: This is related to #104, but not the same. @pushyamig created a new issue (#129) and has an open PR.

pushyamig:
@ssciolla Is there any thing left to do for the issue?

ssciolla:
>     1. Tested the course rename feature by becoming a **Subaccount admin, teacher, and admin** in a course 484127
>        **Subaccount/Root admin**: They could rename the courses
>        **Teacher**: In reality the feature would be disabled from UI but I wanted to verify the scoped token phenomenon and it **did not allow** for course rename but canvas did not error out either with `4xx`. It simply send 200 OK returning the course object and in the object we can say that course was not renamed. This is strange. I wanted to verified this API call using Postman then I confirmed that this is Canvas Phenomenon not KTH library or CCM backend code.

That is strange, I'm glad you tried testing it a bit. Have you been able to find any documentation about this?

>     3. I felt swagger was really helpful when testing all the API calls before the UI wired up. Only annoying thing is i need to copy/paste the token. I wish we had automatic way of doing this. Can we whitelist the swagger API route?

I don't think the `ltijs` "whitelist" would solve the issue here, since it's the requests Swagger makes we need to authenticate, not the `swagger` route (the route may not be protected currently). There might be a way to further automate this, but my initial search didn't turn up anything simple. I can return to this in the future, but it feels a little lower priority to me right now.

pushyamig:
> There might be a way to further automate this, but my initial search didn't turn up anything simple. I can return to this in the future, but it feels a little lower priority to me right now.

I know this an Icebox issue i.e nice to have. I am not asking for it shared my experience.

pushyamig:
> That is strange, I'm glad you tried testing it a bit. Have you been able to find any documentation about this?

I did not since this is not a real issue to solve.  I think UI could handle it with an error message if saw the course name change in the request and response is not same.

lsloan (approved):
I like this.  It's interesting to see how implementing this feature fits into the framework.  Looks like it will be a good model for implementing other API functions.

ssciolla:
Thanks everyone for the reviews! Very possible I'm missing some important detail, but we can continue to iterate as more endpoints are added. Merging now...

PR #125, opened 2021-06-16
#124 waiting on DB container before Web container

### LINKED ISSUE #124: Strategy for waiting on docker dependent containers
The Docker configuration doesn't seems to wait on dependency container to come up, and it is throwing error as below. This won't happen if a DB volume is already created

`ccm_web     | Error during deployment:  ConnectionRefusedError [SequelizeConnectionRefusedError]: connect ECONNREFUSED 172.18.0.2:3306`

Currently this is handled in the `docker-compose.yml` run using `wait-ports`, But the `docker-compose-prod.yml` build is still throwing the error.

I looked at [wait-for](https://github.com/eficode/wait-for), [wait-for-it](https://github.com/vishnubob/wait-for-it) and [wait-port](https://www.npmjs.com/package/wait-port) library and settled on using wait-port since it is a NPM install and was easy to update than having to download the file to source code or doing wget operation.

pushyamig:
Fixes #124


jonespm:
This is probably fine for a node based project. I had to switch from wait-port in netcat in MyLA, I don't remember exactly why. I think it was just because I didn't want to have node available just for this library. You could just add the same pattern we use in MyLA to the start script and have it check there. 

https://github.com/tl-its-umich-edu/my-learning-analytics/pull/590/files

pushyamig:
> you could just add the same pattern we use in MyLA to the start script and have it check there.
tl-its-umich-edu/my-learning-analytics#590 (files)

Wow @jonespm I did not realize that you were handling waiting on DB container before the Web. 

The thing is the `start.sh` is only for Prod use and not for Local development since we don't transpile typescript to java script, Probably I need to get all these into start.sh and based on the `NODE_ENV=production` in the docker-compose I need to change it. 

```
"dev:server": "nodemon --inspect=0.0.0.0:9229 --config nmon_config.json server/src/main.ts",
"predev": "wait-port ccm_db:3306",
"dev": "node -r ts-node/register server/src/migrator up && npm run dev:server & npm run dev:client",
```

pushyamig:
@jonespm I like your idea and my linux/docker skills are rudimentary I feel  I don't want to allocate more time for this since this is only for local development. I also feel that with Local and prod setup are different there might me some funkiness raising with shifting all the logic to `start.sh`.

ccm_web/package.json
```diff
@@ -10,7 +10,6 @@
     "build:client": "webpack --config webpack/webpack.prod.ts",
     "build:server": "tsc --project tsconfig.build.json",
     "preprod": "npm run build:client && npm run build:server",
-    "prod": "NODE_ENV=production node dist/server/src/main.js",
```
ssciolla, line 13: I think there's a line in the README about this. We should remove it if you want to wipe this out (which I think is fine).

ccm_web/package.json
```diff
@@ -10,7 +10,6 @@
     "build:client": "webpack --config webpack/webpack.prod.ts",
     "build:server": "tsc --project tsconfig.build.json",
     "preprod": "npm run build:client && npm run build:server",
```
ssciolla, line 12: We could change the name of the `preprod` command, maybe just to `build`. The `pre*` construction makes less sense if we don't have a `prod` (it already didn't make too much sense, since we don't use `prod` in the Docker process).

docker-compose-prod.yml
```diff
@@ -11,6 +11,7 @@ services:
       - PORT=4000
       - NODE_ENV=production
     env_file: .env
+    command: sh -c './node_modules/.bin/wait-port ccm_db_prod:3306 && ./start.sh'
```
ssciolla, line 14: This looks pretty good! I wish we could simplify the `wait-port` command, but seems that's the documented approach. I know I suggested it, but the `start.sh` file feels borderline unnecessary (the alternative being sticking to `npm` commands). However, I'm imagining that we may want to additional commands for prod at some point, so I'm good with this. The ability to do some logging is also nice.

ccm_web/Dockerfile
```diff
@@ -23,13 +23,13 @@ FROM node:14-slim AS prod
 ENV NODE_ENV=production
 WORKDIR /app
 
-COPY --from=base /base/package.json ./
+COPY --from=base /base/package.json /base/start.sh ./
```
ssciolla, line 26: Wonder if we should be also adding `package-lock.json` (i.e. `package*.json`)...

ssciolla (approved):
Looks good! We could use @jonespm's MyLA approach, but we'd either need to use an environment variable flag or create a separate `.sh` file, and we'd need to install `netcat`, and do some `apt-utils`ing. I've been pretty happy with `wait-port`, and I'm happy with what you've done here (at least for now). Thanks! Feel free to handle my suggestions/comments as you see fit. No need to re-request review.

jonespm:
Yeah, I'm fine with wait-port for a project that's already node and npm. I feel like I had to switch it in MyLA because we don't have node available in the backend stage.

pushyamig:
Created a issue for [rework](https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/131) in future

ccm_web/package.json
pushyamig: Done [`bfa4929` (#125)](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/125/commits/bfa4929fc93af2504b3d05abe82d631dc6a602f4)

ccm_web/package.json
pushyamig: Done [`bfa4929` (#125)](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/125/commits/bfa4929fc93af2504b3d05abe82d631dc6a602f4)

ccm_web/Dockerfile
pushyamig: Done [`bfa4929` (#125)](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/125/commits/bfa4929fc93af2504b3d05abe82d631dc6a602f4)

docker-compose-prod.yml
pushyamig: Yeah.. I like putting the additional logic in `start.sh` which we don't have now, but I time open up this would be nice [improvement](https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/131)

pushyamig:
I resolved all the issue requested and tested the docker build after the changes

PR #127, opened 2021-06-16
Switch to using underscored table and field names (#126)

### LINKED ISSUE #126: Migration error possibly related to camelCase table names
While working on issue #100, I was playing around with migrations, and I encountered a strange error. After I started up the application from an empty database, stepped back once using `down`, I saw this when attempting to reapply the migration.

```
docker exec -it ccm_web node -r ts-node/register server/src/migrator up
2021-06-16T20:03:32.729Z - config.ts - info - Value undefined for SESSION_SECRET is invalid; using fallback SESSIONSECRET
2021-06-16T20:03:32.731Z - config.ts - info - Value undefined for LTI_ENCRYPTION_KEY is invalid; using fallback LTIKEY
2021-06-16T20:03:32.811Z - migrator.ts - info - Running migrations CLI...
Executing (default): CREATE TABLE IF NOT EXISTS `SequelizeMeta` (`name` VARCHAR(255) NOT NULL UNIQUE , PRIMARY KEY (`name`)) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE utf8_unicode_ci;
Executing (default): SHOW INDEX FROM `SequelizeMeta`
Executing (default): SELECT `name` FROM `SequelizeMeta` AS `SequelizeMeta` ORDER BY `SequelizeMeta`.`name` ASC;
2021-06-16T20:03:32.887Z - migrator.ts - info - {
  "event": "migrating",
  "name": "2021.06.08T13.39.45.create-canvas-token.ts"
}
Executing (default): CREATE TABLE IF NOT EXISTS `canvasToken` (`id` BIGINT NOT NULL auto_increment , `userId` BIGINT NOT NULL, `accessToken` VARCHAR(255) NOT NULL, `refreshToken` VARCHAR(255) NOT NULL, PRIMARY KEY (`id`), FOREIGN KEY (`userId`) REFERENCES `user` (`id`)) ENGINE=InnoDB;

Error: Table 'ccm.canvastoken' doesn't exist
2021-06-16T20:03:33.098Z - migrator.ts - info - Migration tasks ran!
```

We have seen some other strange errors related to casing in table names, which may or may not be related. I think we should investigate whether to use underscored table names, as these are all lowercase and may avoid some of these issues ( a s well as maintain consistency with other projects and follow general MySQL naming conventions). Sequelize supports using an `underscored` option so we can still use camelCase with model methods. We may have to modify or re-create existing migrations (which is not ideal), but should theoretically be okay as we have not deployed to production.

ssciolla:
This PR modifies existing migrations and model definitions so table name and field names use `snake_case`, with the goal of avoiding case-related issues. Modifying migrations was deemed safe by the team, because we have no deployments; timestamps have been updated. Model and attribute names in TS-code still use `[cC]amelCase`, to be consistent with JS best practices. The PR aims to resolve #126.

**Note: you will need to delete your database volume (`.mysql_data`) before running this locally, and if/when it is merged.**

### DIFF
--- a/ccm_web/server/src/app.module.ts
+++ b/ccm_web/server/src/app.module.ts
@@ -35,7 +35,8 @@ const logger = baseLogger.child({ filePath: __filename})
         password: configService.get('db.password'),
         database: configService.get('db.name'),
         models: [CanvasToken, User],
-        logging: (message: string) => logger.debug(message)
+        logging: (message: string) => logger.debug(message),
+        define: { underscored: true }
       })
     }),
     UserModule,
--- a/ccm_web/server/src/app.module.ts
+++ b/ccm_web/server/src/app.module.ts
@@ -36,7 +36,7 @@ const logger = baseLogger.child({ filePath: __filename})
         database: configService.get('db.name'),
         models: [CanvasToken, User],
         logging: (message: string) => logger.debug(message),
-        define: { underscored: true }
+        define: { underscored: true } // Included here to ensure session table uses snake_case
       })
     }),
     UserModule,
--- a/ccm_web/server/src/migrations/2021.06.16T19.36.19.create-user.ts
+++ b/ccm_web/server/src/migrations/2021.06.16T19.36.19.create-user.ts
@@ -11,11 +11,13 @@ export const up: Migration = async ({ context: sequelize }) => {
     },
     firstName: {
       allowNull: true,
-      type: DataTypes.STRING
+      type: DataTypes.STRING,
+      field: 'first_name'
     },
     lastName: {
       allowNull: true,
-      type: DataTypes.STRING
+      type: DataTypes.STRING,
+      field: 'last_name'
     },
     email: {
       allowNull: true,
@@ -24,15 +26,18 @@ export const up: Migration = async ({ context: sequelize }) => {
     loginId: {
       allowNull: false,
       unique: true,
-      type: DataTypes.STRING
+      type: DataTypes.STRING,
+      field: 'login_id'
     },
     createdAt: {
       allowNull: false,
-      type: DataTypes.DATE
+      type: DataTypes.DATE,
+      field: 'created_at'
     },
     updatedAt: {
       allowNull: false,
-      type: DataTypes.DATE
+      type: DataTypes.DATE,
+      field: 'updated_at'
     }
   })
 }
--- a/ccm_web/server/src/migrator.ts
+++ b/ccm_web/server/src/migrator.ts
@@ -25,7 +25,7 @@ export const umzug = new Umzug({
     }
   },
   context: sequelize,
-  storage: new SequelizeStorage({ sequelize, tableName: 'app_sequelize_meta' }),
+  storage: new SequelizeStorage({ sequelize, tableName: 'ccm_sequelize_meta' }),
   logger: logger
 })
 
--- a/ccm_web/server/src/migrator.ts
+++ b/ccm_web/server/src/migrator.ts
@@ -25,7 +25,7 @@ export const umzug = new Umzug({
     }
   },
   context: sequelize,
-  storage: new SequelizeStorage({ sequelize }),
+  storage: new SequelizeStorage({ sequelize, tableName: 'app_sequelize_meta' }),
   logger: logger
 })
 
--- a/ccm_web/server/src/user/user.model.ts
+++ b/ccm_web/server/src/user/user.model.ts
@@ -20,8 +20,7 @@ interface UserCreationAttributes extends Optional<UserAttributes, 'id'> {
 // Sequelize Datatypes mapping with Mysql https://sequelize.org/master/manual/model-basics.html#data-types
 @Table({
   tableName: 'user',
-  freezeTableName: true,
-  underscored: true
+  freezeTableName: true
 })
 export class User extends Model<UserAttributes, UserCreationAttributes> {
   @Column({
--- a/ccm_web/server/src/user/user.model.ts
+++ b/ccm_web/server/src/user/user.model.ts
@@ -20,7 +20,8 @@ interface UserCreationAttributes extends Optional<UserAttributes, 'id'> {
 // Sequelize Datatypes mapping with Mysql https://sequelize.org/master/manual/model-basics.html#data-types
 @Table({
   tableName: 'user',
-  freezeTableName: true
+  freezeTableName: true,
+  underscored: true
 })
 export class User extends Model<UserAttributes, UserCreationAttributes> {
   @Column({
--- a/ccm_web/server/src/canvas/canvas.model.ts
+++ b/ccm_web/server/src/canvas/canvas.model.ts
@@ -15,7 +15,6 @@ interface CanvasTokenCreationAttributes extends Optional<CanvasTokenAttributes, 
 @Table({
   tableName: 'canvas_token',
   freezeTableName: true,
-  underscored: true,
   timestamps: false
 })
 export class CanvasToken extends Model<CanvasTokenAttributes, CanvasTokenCreationAttributes> {
--- a/ccm_web/server/src/canvas/canvas.model.ts
+++ b/ccm_web/server/src/canvas/canvas.model.ts
@@ -13,8 +13,9 @@ interface CanvasTokenAttributes {
 interface CanvasTokenCreationAttributes extends Optional<CanvasTokenAttributes, 'id'> {}
 
 @Table({
-  tableName: 'canvasToken',
+  tableName: 'canvas_token',
   freezeTableName: true,
+  underscored: true,
   timestamps: false
 })
 export class CanvasToken extends Model<CanvasTokenAttributes, CanvasTokenCreationAttributes> {
--- a/ccm_web/server/src/migrations/2021.06.16T19.36.43.create-canvas-token.ts
+++ b/ccm_web/server/src/migrations/2021.06.16T19.36.43.create-canvas-token.ts
@@ -2,7 +2,7 @@ import { DataTypes } from 'sequelize'
 import { Migration } from '../migrator'
 
 export const up: Migration = async ({ context: sequelize }) => {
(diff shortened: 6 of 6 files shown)


ccm_web/server/src/app.module.ts
```diff
@@ -31,7 +31,8 @@ import { validateConfig } from './config'
         username: configService.get('db.user'),
         password: configService.get('db.password'),
         database: configService.get('db.name'),
-        models: [CanvasToken, User]
+        models: [CanvasToken, User],
+        define: { underscored: true }
```
ssciolla, line 35: This may not be necessary, but seems pretty harmless, may prevent future issues.

ccm_web/server/src/migrations/2021.06.16T19.36.19.create-user.ts
```diff
@@ -11,11 +11,13 @@ export const up: Migration = async ({ context: sequelize }) => {
     },
```
ssciolla, line 11: Should we also specify `field` for single-word fields? It isn't necessary, but could make things clearer.

lsloan:
I've already been in the habit of frequently deleting `.mysql_data` while working on this project.

ccm_web/server/src/app.module.ts
lsloan: If/when #140 is merged, you'll need to pull those changes so you get the `logging` setting here, too.

ccm_web/server/src/migrations/2021.06.16T19.36.19.create-user.ts
lsloan: I'd say no.

Django also has the ability to specify different names for objects in the DB, but I only use them when necessary.  So I would do the same here.

ccm_web/server/src/user/user.model.ts
```diff
@@ -20,7 +20,8 @@ interface UserCreationAttributes extends Optional<UserAttributes, 'id'> {
 // Sequelize Datatypes mapping with Mysql https://sequelize.org/master/manual/model-basics.html#data-types
 @Table({
   tableName: 'user',
-  freezeTableName: true
+  freezeTableName: true,
+  underscored: true
```
lsloan, line 24: Why is `underscored` necessary here?  The table name doesn't include underscores.  Must this be specified if the columns within the table have names with underscores?

lsloan (commented):
This PR will resolve the problem I mentioned a few weeks ago, before #126 was opened.

Does the framework do any automatic translation between mixed-case and underscored symbol names?  I wonder why `underscored` needs to be specified while `tableName` contains a name that contains underscores.  Is that a flag to the framework that some translation needs to occur, rather than it assuming it should translate based on the presence of underscores?  (I know… RTFM, right?)

Aside from some questions, I like this.  I'll give it a try.

ccm_web/server/src/app.module.ts
ssciolla: Whose PR was open first? 🤣

pushyamig (approved):
Sam, This looks Good and thanks for improvement. I did not run this but changes seems to make sense to me.

ccm_web/server/src/migrations/2021.06.16T19.36.19.create-user.ts
ssciolla: K, thanks for the input

ccm_web/server/src/app.module.ts
pushyamig: >  define: { underscored: true }

Can this be added to `migrator.ts` as well?

ccm_web/server/src/user/user.model.ts
ssciolla: You know, maybe it's not. I will try removing it and see what happens. The documentation for Sequelize says that it will convert camelCase attribute names to snake_case ones, but I'm already doing that manually via the `field` in the migrations. I guess I was thinking that this would give the model more information, allow for the correct traversal between snake_case tables/fields and [cC]amelCase models and attributes, but I'm questioning that now.
https://sequelize.org/master/manual/naming-strategies.html

ccm_web/server/src/app.module.ts
ssciolla: I guess I need to determine whether it's needed first, see https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/127#discussion_r658826076

lsloan: I think it matters which PR is merged first.  That is, who can click the "Merge" button faster.

On your mark! Get set! 😆 

SRSLY, it sounds like you may want to investigate some changes to this PR anyway, so the other may get merged first.  Que sera, sera.

ccm_web/server/src/user/user.model.ts
lsloan: I was going to check out your branch to try it, but based on [your comment](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/127#discussion_r658826076), I'll hold off until you've experimented a bit more with combinations of the `underscored` and name properties.

Speaking of branches, I thought **_I_** had a lot of old branches in my fork.  You win. 🏆

ssciolla: Okay, @lsloan, I did some research and experimented. `underscored` seems to have a long history of being buggy in `sequelize` (see [this issue](https://github.com/sequelize/sequelize/issues/11225)). Based on my current understanding, I think we can go without specifying it and prefer the hardcoded approach using `field`. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/127/commits/00327465fb1ec84dd9dbad1561d0e7bba7448831

I do have a question about how to handle all this with `session`, which I'll comment on in the thread off of `app.module.ts`.

ccm_web/server/src/app.module.ts
ssciolla: If that gets merged first, I'll resolve the conflict, claro que si.

ssciolla: There may be some merit to including `underscored` here, as this is what determines the behavior of the `session` model automatically generated by the `connect-session-mysql` library related to `express-session`. What do you think? If we remove it, we'll have some `createdAt` type fields in `session` (the automatically generated fields seem to actually be affected by `underscored`, don't ask me why!).

ccm_web/server/src/migrator.ts
```diff
@@ -25,7 +25,7 @@ export const umzug = new Umzug({
     }
   },
   context: sequelize,
-  storage: new SequelizeStorage({ sequelize }),
+  storage: new SequelizeStorage({ sequelize, tableName: 'app_sequelize_meta' }),
```
ssciolla, line 28: I decided to separate our `migrations` table from the one used by `ltijs`. Let me know what you think.

ccm_web/server/src/user/user.model.ts
ssciolla: Some additional explanation: we could have tried to lean into `underscored` and not use the `field` approach. However, when I did this, I didn't see that the custom-defined fields like `User.lastName` followed the snake_case naming convention. So I fell back to using `field`, which seems more predictable.

ccm_web/server/src/migrator.ts
ssciolla: We could rename to whatever, or not make the change at all. Just an idea. I guess I was trying to avoid more table names with camelCasing. And there might be some benefit to keeping `ltijs` even more separate from the rest of what we're doing.

ssciolla: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/127/commits/3f836f18ba5ac8633a33c2d741143c1bf108dc41

pushyamig:
I have tested this PR with changed convention to underscore to camelCasing. The PR is working against Local and Prod build.

Mysql best practice table name convention seem to be using underscore for tables and fields so hopefully we don't end up in more messy situations with naming convention. 

Having a separate migration table name other than default is working and keeping things simple. I don't know what is the best practice and what will be side effects of this. We can only find out as we go. Library Sequelize did not said they don't recommend it but so I suppose there are use cases out there since it made it as configurable option.

pushyamig (approved):

ccm_web/server/src/app.module.ts
lsloan: I saw in the documentation you linked to earlier that it said `created_at` would be used if `underscored` is set to true.  I think we want that.

lsloan (commented):
So, just to make sure I understand, you've removed the object-level `underscored` definitions because it is already specified at the database level.  Is that right?

ccm_web/server/src/app.module.ts
ssciolla: > So, just to make sure I understand, you've removed the object-level `underscored` definitions because it is already specified at the database level. Is that right?

Kinda. `underscored` in the models wasn't doing anything extra because `field` was specified when needed and that value overrides any other behavior. Besides, the model definition is separate from the migration, they need to be kept in sync in our current setup. If we wanted to try to use `underscored`, I think the important place to set this would be in `migrator.ts`. However, I wasn't seeing the `underscored` naming strategy being used consistently when I tried it there. So I'm relying on `field` instead. I hope that helps explain. This is confusing to me as well. I recommend trying it out yourself.

ssciolla: I'm keeping `underscored` here because, as I said, this `sequelize` instead is responsible for the `session` table via `sync` (see `main.ts`). If you remove this setting, you should see that the session table uses `createdAt`. I am currently keeping this setting so that it uses `created_at`.

pushyamig:
@lsloan and @ssciolla do you still have any concerns that needs to be resolved. I think this PR is blocking this PR #135

lsloan (approved):
No further questions. :shipit:

ssciolla:
Thanks for the reviews, merging now...

PR #130, opened 2021-06-18
#129 error message when needed scopes not selected

### LINKED ISSUE #129: Show appropriate error message when scopes don't match
For the Canvas API key we check the scopes from the Developer key console and also mention the scopes in the CCM [code base](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/main/ccm_web/server/src/canvas/canvas.scopes.ts). 
When you miss checking the required scope currently logs it as `'State variable returned from Canvas did not match session ID; throwing unauthorized exception..` and from UI `You are not authorized to access this resource.` This is message is misleading and make you think something wrong with session data 

Steps to reproduce:
1. Go to Developer console and edit the API key scopes
2. don't chose the scopes that is mentioned in the [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/main/ccm_web/server/src/canvas/canvas.scopes.ts)
3. Delete the your token from the `canvasToken` table and launch CCM
4. This should show you the error

pushyamig:
Fixes #129

### DIFF
--- a/ccm_web/server/src/canvas/canvas.interfaces.ts
+++ b/ccm_web/server/src/canvas/canvas.interfaces.ts
@@ -1,10 +1,13 @@
 import { hasKeys } from '../typeUtils'
 
-export interface OAuthResponseQuery {
+export interface OAuthGoodResponseQuery {
   code: string
   state: string
 }
-
+export interface OAuthErrorResponseQuery {
+  error: string
+  error_description: string
+}
 export interface TokenResponseBody {
   access_token: string
   token_type: 'Bearer'
--- a/ccm_web/server/src/canvas/canvas.interfaces.ts
+++ b/ccm_web/server/src/canvas/canvas.interfaces.ts
@@ -50,3 +50,7 @@ export function isCanvasErrorBody (value: unknown): value is CanvasErrorBody {
     return result
   }
 }
+
+export const isOAuthErrorResponseQuery = (value: unknown): value is OAuthErrorResponseQuery => {
+  return hasKeys(value, ['error']) && hasKeys(value, ['error_description'])
+}
--- a/ccm_web/server/src/canvas/canvas.utils.ts
+++ b/ccm_web/server/src/canvas/canvas.utils.ts
@@ -1,0 +1,5 @@
+import { OAuthErrorResponseQuery, OAuthGoodResponseQuery } from './canvas.interfaces'
+
+export const isOAuthErrorResponseQuery = (data: OAuthErrorResponseQuery | OAuthGoodResponseQuery): data is OAuthErrorResponseQuery => {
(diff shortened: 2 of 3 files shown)


ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -1,8 +1,14 @@
...
   code: string
   state: string
 }
 
+export interface OAuthErrorResponseQuery {
+  error: string
+  error_description: string
+}
+
+export interface OAuthResponseQuery extends OAuthGoodResponseQuery, OAuthErrorResponseQuery {}
```
ssciolla, line 11: So, what's the expected shape here? Is it `OAuthGoodResponseQuery` **or** `OAuthErrorQuery`, or both combined? If it's the former, I think we want to change the type on line 47 of the controller so it's `OAuthGoodResponseQuery | OAuthErrorQuery`. Then we probably want a type guard to check for `error` and `error_description`, something like this?

```
function isOAuthErrorResponseQuery (data: Record<string, unknown>): data is OAuthErrorResponseQuery {
  return 'error' in data && 'error_description' in data
}
```

Then you can change the type check on 53 of the controller, to `isOAuthErrorResponseQuery(query)`.

Maybe this is change is marginal, but I'd like to ensure the types are describing the right objects in the two scenarios.

pushyamig: Thanks! I resolved this as per the suggestion. I also added `canvas.utils.ts` file  to simplify the canvas controller file a bit. LMK what you think about this.

pushyamig:
Let me resolve the conflict first.

pushyamig:
> Let me resolve the conflict first.

Done. This is ready for review

ssciolla (approved):
Nice improvement! Thanks for doing this. I left some code organization/style comments, but this is functional. Handle them as you see fit.

ccm_web/server/src/canvas/canvas.utils.ts
```diff
@@ -0,0 +1,5 @@
+import { OAuthErrorResponseQuery, OAuthGoodResponseQuery } from './canvas.interfaces'
+
+export const isOAuthErrorResponseQuery = (data: OAuthErrorResponseQuery | OAuthGoodResponseQuery): data is OAuthErrorResponseQuery => {
+  return 'error' in data && 'error_description' in data
+}
```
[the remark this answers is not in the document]
pushyamig: Done. [`4fadd87` (#130)](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/130/commits/4fadd87e181e9ec0e7f8248ba2a817b85dfb6053)

pushyamig:
I resolved all the issue as per suggestion so merging it

PR #135, opened 2021-06-23
Add proactive refreshing solution for Canvas tokens (#100, #136)

### LINKED ISSUE #100: Refresh token workflow when token expires after 1 hr

ssciolla:
This PR adds proactive token refreshing to `CanvasService` (i.e. refresh tokens just before they expire). It also fixes handling of HTTP errors originating from `HttpService` calls (wrapper for `axios`); introduces custom error classes (subclasses of `Error`) to make error throwing consistent; and throws errors when unexpected problems occur (API or Database problems, or missing users or tokens when they are expected to exist) rather than returning `null` and leaving the controller to interpret that. The custom errors thrown in the `*Service` levels will be caught by the NestJS exception filter (see [here](https://docs.nestjs.com/exception-filters)), resulting in `500` errors on routes. We can look into customizing this as well. The PR aims to resolve #100 and #136 and will work towards the resolution of #137.

Resources:
- https://canvas.instructure.com/doc/api/file.oauth.html#using-refresh-tokens
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error


ccm_web/server/src/migrations/2021.06.30T19.52.46.add-canvas-token-expires-at.ts
```diff
@@ -0,0 +1,19 @@
+import Sequelize, { DataTypes } from 'sequelize'
+import { Migration } from '../migrator'
+
+export const up: Migration = async ({ context: sequelize }) => {
+  await sequelize.getQueryInterface().addColumn(
+    'canvas_token',
+    'expires_at',
```
ssciolla, line 7: It's kind of unusual to me that, unlike the `CanvasToken` migration, you don't ever actually specify the model's property name in `camelCase` (e.g. `expiresAt`). The mapping still works though, go figure. If you change this first usage to `expiresAt`, it seems to set the database column name to that, even with the `field` specified in options below.

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -47,75 +51,138 @@ export class CanvasService {
...
     /*
     Make a call to the Canvas API to create token
     https://canvas.instructure.com/doc/api/file.oauth_endpoints.html#post-login-oauth2-token
     */
 
+    const user = await this.userService.findUserByLoginId(userLoginId)
+    if (user === null) {
+      logger.error(`User ${userLoginId} is not in the database.`)
+      throw new UserNotFoundError(userLoginId)
+    }
```
ssciolla, line 73: If/when we implement Passport and attach `User` objects to `Request`s, we will be able to throw this error in one place, and I can instead pass a `User` object in the first parameter position to a few `CanvasService` methods. I will tackle that with the Passport work.

I could get even closer to this by moving the `UserService` calls to the controllers, but I wanted to minimize changes at this point.

ccm_web/server/src/migrations/2021.06.30T19.52.46.add-canvas-token-expires-at.ts
```diff
@@ -0,0 +1,19 @@
...
+    {
+      allowNull: false,
+      type: DataTypes.DATE,
+      field: 'expires_at',
+      defaultValue: Sequelize.fn('NOW')
```
ssciolla, line 12: The default value is the current time. This `sequelize` function seems to handle the MySQL conversion. I saw other patterns online using a `sequelize.NOW` function, but that pattern didn't seem to work for me.

ssciolla:
Okay, I've updated this to be merge-able with `main`, and I've incorporated some error handling stuff I was working on separately. I want to continue refactoring the services (as I've indicated a little in the description), but hopefully this is an improvement and adds the important refreshing functionality. I'll address questions when I return next week.

pushyamig:
The initial testing of refresh token strategy seems to be working. 
1. The migration created the new column `expire_at` and populated with current date
2. The expired access token got renewed and updated the DB. I guess the refresh token always remain the same unless until user revokes the authorization or scope change. 

I will conti review of the PR .

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -47,75 +51,136 @@ export class CanvasService {
...
+    } catch (error) {
+      if (axios.isAxiosError(error) && error.response !== undefined) {
+        logger.error(`Received unusual status code ${error.response.status}`)
+        logger.error(`Response body: ${JSON.stringify(error.response.data, null, 2)}`)
+      } else {
+        logger.error(
+          `Error occurred while making request to Canvas for access token: ${JSON.stringify(error, null, 2)}`
+        )
+      }
+      throw new CanvasOAuthAPIError()
```
pushyamig, line 158: When this case occurs it is ending up in with error message in UI as `{"statusCode":500,"message":"Internal server error"}`. I know that in the API service it is configured like that, but couldn't figure out how this case is trigged in the Canvas service Case. 

In Logs this is how it is show
```
ccm_web     | [Nest] 91   - 07/06/2021, 2:33:19 PM   [ExceptionsHandler] An error occurred while requesting or refreshing a Canvas OAuth access token. +100818ms
ccm_web     | CanvasOAuthError: An error occurred while requesting or refreshing a Canvas OAuth access token.
ccm_web     |     at CanvasService.<anonymous> (/base/server/src/canvas/canvas.service.ts:102:13)
ccm_web     |     at Generator.throw (<anonymous>)
ccm_web     |     at rejected (/base/server/src/canvas/canvas.service.ts:18:65)
ccm_web     |     at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

This comment is for clarity purpose, nothing to change

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -47,75 +51,136 @@ export class CanvasService {
...
+      refresh_token: token.refreshToken
+    }
+
+    let data: TokenRefreshResponseBody | undefined
+    const timeOfRequest = new Date()
+
+    try {
+      const response = await this.httpService.post<TokenRefreshResponseBody>(
+        `${this.url}/login/oauth2/token`, params
+      ).toPromise()
```
pushyamig, line 146: what is the purpose of having `toPromise` here? this is something I noticed new.

ccm_web/server/src/user/user.errors.ts
```diff
@@ -0,0 +1,12 @@
...
+  public name = 'UserNotFoundError'
+
+  constructor (userLoginId: string) {
+    super(
+      `The user "${userLoginId}" could not be found; ` +
+      'the database may have been modified or there may be a problem with the application code.'
+    )
+    this.date = new Date()
+  }
+}
```
pushyamig, line 12: I appreciate your effort with structuring the custom error types here, this really make code well organized. Thanks for taking time and doing it!

ccm_web/server/src/canvas/canvas.errors.ts
```diff
@@ -0,0 +1,22 @@
+export class CanvasOAuthAPIError extends Error {
+  date: Date
+  public name = 'CanvasOAuthError'
```
pushyamig, line 3: Minor: Do you mean to name the public error name as `CanvasOAuthAPIError` instead of `CanvasOAuthError`? either way it is fine, but felt class name and public name could be same

pushyamig:
Over all this PR looks good, I have few comments left but nothing major for stopping from getting merged.

pushyamig (approved):

ccm_web/server/src/canvas/canvas.service.ts
ssciolla: The `HttpService` pattern described here (https://docs.nestjs.com/techniques/http-module) returns an `rxjs` `Observable` (see https://rxjs.dev/guide/observable), which I don't really know how to use effectively. So I'm just calling this method to make it a `Promise` for now.

ccm_web/server/src/canvas/canvas.service.ts
ssciolla: When an error is thrown by a controller-initiated process but isn't caught, NestJS catches it using a global exceptiion filter and throws a `500`. Check out the beginning of this section: https://docs.nestjs.com/exception-filters#exception-filters-1

If we want, I think we can do more here so that we can pass more specific error info to the UI, likely by using a custom exception filter. That might involve making these custom errors NestJS `Exception`s rather than `Error`s. I can look into this more when trying to close #137, or do it now if you want me to.

ccm_web/server/src/user/user.errors.ts
ssciolla: 😃

ccm_web/server/src/canvas/canvas.errors.ts
ssciolla: Oh, yeah, I'll fix that, must have missed it.

ccm_web/server/src/canvas/canvas.service.ts
pushyamig: I am ok addressing this as part of #137

ccm_web/server/src/canvas/canvas.errors.ts
ssciolla: Done, see https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/135/commits/9d00fe543c9f4081e6494e63e3d3faf71b54b2de

ccm_web/server/src/canvas/canvas.service.ts
pushyamig: Well, This is definitely unknown territory for me. So what ever works, we should go for it. 

Mostly the Observer pattern is used if 2 parts of the UI has same piece of information to update and these 2 UI modules/parts ( what ever you want to call) might be using 2 different service layers. So in that case these 2 parts will subscribe to same a source of truth and receive updates when the source of truth changes. For example, in the UI there might be 2 place where you have Thumbs up/down count and you want to keep both places in sync you will end up with Observable pattern.

With that said, I really don't know how that situation will comprehend to our scenario to receive token stuff. Probably what you did is correct.

ssciolla:
Thanks for the review, @pushyamig! Merging now, since I want people to have access to this refreshing stuff.

PR #140, opened 2021-06-24
configure Sequelize to log at `debug` level (iss. #91)

### LINKED ISSUE #91: Control verbosity of Sequelize
Currently, the logs include detailed SQL statements for each database transaction. This is useful but probably not desirable in all situations. This should be configurable.

https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/4a9bbe78b4a7c0673d0bfbe01efb5e20b95ddcad/ccm_web/server/src/app.module.ts#L13-L39

The configuration for this should be added using a `logging` property in the `useFactory` function return above.

lsloan:
As was discussed in an earlier meeting, configured Sequelize to use the winston logger at `debug` level for logging its messages.

CCM's logging level is controlled by the `LOG_LEVEL` directive in `.env`.  If it is not set, the default level is `debug`.  If `LOG_LEVEL` is set to any level higher than `debug` (i.e., `info`, `notice`, `warning`, `error`, etc.), Sequelize's messages will not appear in the output.  (See: https://github.com/winstonjs/winston#logging-levels)

Resolves #91.

This took me longer than expected to implement due to Docker poor performance on my computer.  Having to restart the Docker daemon periodically impacts productivity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,14 +81,22 @@ but as necessary, rebuild the image using Step 1.
 
 Explicit steps for setting up CCM in a development environment.
 
+##### Start local ngrok server
+
 1. Start an ngrok instance for the application, which will run on port 4000.
    `ngrok http 4000`
 2. Make note of the hostname of the ngrok instance for use later.  The hostname may be found in the console output or it can be obtained from the ngrok API.  For example, if `jq` is installed, use the command:
    `curl --silent http://127.0.0.1:4040/api/tunnels | jq -r '.tunnels[0].public_url'`
-   ***Note:*** Get only the hostname, not the `http`/`https` scheme prefix.  
+   ***Note:*** Get only the hostname, not the `http`/`https` scheme prefix.
+
+##### Create LTI configuration file
+
 3. Create an LTI key JSON file, based on the sample provided.
    `cp config/lti_dev_key_sample.json lti_dev_key.json`
 4. Edit the new `lti_dev_key.json` file to add or replace settings from the sample.  Replace all occurrences of `{ccm_app_hostname}` with the ngrok hostname copied in the earlier step.
+
+##### Configure Canvas LTI key
+
 5. Go to Canvas "Developer Keys" management page available from the "Admin" page for your account.  Click the button for "+ Developer Key", then "+ LTI Key".
 6. When the "Key Settings" modal appears, select "Paste JSON" from the "Method" menu in the "Configure" section.  Then copy the contents of `config/lti_dev_key.json` and paste it into the "LTI 1.3 Configuration" field.
 7. *Recommended:* Enter a name for the application in the "Key Name" field and an email address in the "Owner Email" field.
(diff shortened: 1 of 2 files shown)


pushyamig:
@ssciolla and @chrisrrowland please review this when ever you get a chance

README.md
```diff
@@ -98,44 +106,53 @@ Explicit steps for setting up CCM in a development environment.
...
 11. Click the "ON" part of the switch in the "State" column of your LTI key, so that it has a green background.
+
+##### Configure Canvas API key
+
 12. Go to Canvas "Developer Keys" management available from the "Admin" page for your account. Click the button for "+ Developer Key", then "+ API Key".
 13. Enter a name in the "Key Name" field, and under "Redirect URIs" enter:
-    `https://{ccm_app_hostname}/canvas/returnFromOAuth`
-    Where `{ccm_app_hostname}` is the ngrok hostname copied earlier (in step 2).
-    ***Note:*** Do ***NOT*** use the "Redirect URI (Legacy)" field.
+     `https://{ccm_app_hostname}/canvas/returnFromOAuth`
```
ssciolla, line 114: So indentation is off according to your linter? Should be 5 spaces? Did you apply this change uniformly? I feel like I see 3 spaces some places, 4 most of the time, and then 5 in these places you added. Something doesn't seem right.

ssciolla (approved):
Looks okay to me. The `README.md` updates are unrelated to this issue and I don't quite understand the indentation changes, but I won't block merging over it.

There are two other places that lead to emitted SQL:
1) the `migrator`: doesn't seem to be using `winston`, I don't know why, I couldn't figure it out in a few minutes. I think we do want to see migration changes in all scenarios, but would be nice if it used the `winston` formatting.
2) the session setup, which is done using `sync` in `main.ts`. That's at the `info` level and uses `winston`; I think like with the migrations, it's good to see that being set up, even in production.

README.md
lsloan: According to the Markdown editor I use (https://typora.io/), yes.  Indentation depends on the number of digits in numbered list.  Single digit items need 4 spaces (room for one digit, a period, and two spaces), double digit items need 5 (room for two digits, a period, and two spaces).  Bullets need only 3 (room for the bullet and two spaces).  I think in Markdown additional lines are supposed to appear to be indented under the bullet/item that contains it.

For example…

* bullet 1
   info that
   goes with
   bullet 1
* bullet 2

9. single-digit item
    info that 
    goes with
    item 9
10. double-digit item
     info that
     goes with
     item 10

ccm_web/server/src/app.module.ts
```diff
@@ -31,7 +34,8 @@ import { validateConfig } from './config'
         username: configService.get('db.user'),
         password: configService.get('db.password'),
         database: configService.get('db.name'),
-        models: [CanvasToken, User]
+        models: [CanvasToken, User],
+        logging: (message: string) => logger.debug(message),
```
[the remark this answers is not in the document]
lsloan: I'm sorry about that. I like the dangling comma, and my IDE didn't mark any problems in the syntax, even with `.eslintrc` in effect.  I'll remove the dangler.

lsloan:
@ssciolla, from [your comments](#pullrequestreview-694425952), I take it that:

1. Migrator code should be changed to use winston, too.
2. Session setup is fine as it is.

You've approved this PR, so I'll make a new issue for item 1.

PR #144, opened 2021-07-01
Hookup course rename w/ backend

### LINKED ISSUE #122: Integrate the FE with the BE endpoints created for course rename
Course re-naming endpoints are now in `main`. `GET` and `PUT` methods are supported at '/course/:id/name'. These will need to be used by and integrated with the existing client code.

chrisrrowland:
Closes #122 

Using usePromise hook here, however I'm not sure how well it works.

The result of the usePromise task is always void, so when passing that task to InlineTextEdit, that component is not aware if the save was successful or not.
Side effect of this is that if you wanted to retry the same name again, you would have to reload the page or attempt to save it with a different name so the component would recognize the text changed again to enable the save button.

Additionally made a slight change to handleErrors, and I'm passing the error through but not doing any sort of validation on the message in Home.  Might need to consider a more comprehensive strategy for this.


ccm_web/client/src/pages/Home.tsx
```diff
@@ -33,7 +36,38 @@ interface HomeProps extends LtiProps {
...
+    setCourseNameAsync,
+    (course: CanvasCourseBase|undefined) => {
+      setCourse(course)
+      enqueueSnackbar('Course name saved', { variant: 'success' })
+    }
+  )
+
+  useEffect(() => {
+    if (setCourseNameError !== undefined) {
+      enqueueSnackbar('Error saving course name: ' + setCourseNameError.message, { variant: 'error' })
```
chrisrrowland, line 68: Error message from backend being spit out here verbatim.  Probably not the best idea.

ccm_web/client/src/utils/handleErrors.ts
```diff
@@ -48,8 +48,8 @@ const handleErrors = async (resp: Response): Promise<void> => {
       console.error(text)
       throw new NotFoundError()
     default:
-      console.error(await resp.text())
-      throw new Error(resp.statusText)
+      text = (JSON.parse(await resp.text()) as Error).message
```
chrisrrowland, line 51: Is there some scenario where this won't work?
resp.statusText was blank in my use case, so i changed it to parse the message from the json object in the text.

pushyamig:
@chrisrrowland you opened it as a draft? is that intentional?

chrisrrowland:
> @chrisrrowland you opened it as a draft? is that intentional?

Yeah there are a couple of bits I commented on that I wasn't sure about.

ccm_web/client/src/pages/Home.tsx
pushyamig: I trigged the error case with unknown canvas endpoint and that will throw 422 error and currently the error message the BE is sending is HTml document. When I changed to get the correct response variable then it will show as below
Screen Shot 2021-07-01 at 1 52 10 PM

I don't know if user want more info here other than `course name was not changed` 

You can do `Error saying course name due to setCourseNameError.message, { variant: 'error' }`, what do you think?

ccm_web/client/src/api.ts
```diff
@@ -1,36 +1,63 @@
...
+  request.method = 'PUT'
+  request.body = body
+  return request
+}
+
+export const getCourse = async (key: string | undefined, courseId: number): Promise<CanvasCourseBase> => {
+  const request = getGet(key)
+  const resp = await fetch(`/api/course/${courseId}/name`, request)
+  await handleErrors(resp)
+  return await resp.json()
```
pushyamig, line 41: Screen Shot 2021-07-01 at 1 10 27 PM

I see this error when getting course name failed. should be leave it as blank if the call failed? or we can pull that info from LTI launch and give it.

I am OK to leave it blank if that's not available

pushyamig: Another thing I noticed is that the `getCourseName` is called every time when user switches between features. To me that is redundant call when user navigating using breadcrumbs

I feel some how the course name should be saved in the React state or props once we get the course name and not make API call every time. since homepage is often visited while navigating between features so getCourseName call should not be made after success. UI should make another getCourseName call if it failed to get during the launch when Homepage is navigated using breadcrumbs

pushyamig:
When I logged in as a Teacher, I see a blank homepage and lot of errors in browser console. Something s not working right for Teacher role since he should not be given access to edit feature

ccm_web/client/src/utils/handleErrors.ts
pushyamig: Backend will always send response text I don't know why it was blank, what was the status code did you receive for a blank statusText

ccm_web/client/src/api.ts
pushyamig: Also whenever the PUT happens for updating the course name, the course name that is holding in the local react code should be updating as well.

ccm_web/client/src/pages/Home.tsx
chrisrrowland: I wasn't sure what to do here.  I think to really handle it well there needs to be more to the Error object.. perhaps some custom errors that extend  the Error class.  I know @ssciolla was trying to provide error feedback to the user in the case of an invalid course name, so in that case it would make sense to give more details, while in general I believe its pretty much always a bad idea to pass errors along verbatim.  I'll see if i can do some custom error type.

ccm_web/client/src/utils/handleErrors.ts
chrisrrowland: I believe it was 400.

chrisrrowland:
> When I logged in as a Teacher, I see a blank homepage and lot of errors in browser console. Something s not working right for Teacher role since he should not be given access to edit feature

Trying to figure out why the getCourse response seems to have no body in that case..

ccm_web/client/src/api.ts
chrisrrowland: @pushyamig get course seems to take about .5 sec or less, testing locally for me.  Not insignificant but not huge.  Do you think putting course info into globals should be spun off into its own issue?

chrisrrowland: What local react code are you referring to.. do you se an issue somewhere?  Not [this](https://github.com/chrisrrowland/canvas-course-manager-next/blob/fe8cd3a0395ea5036181043c385dca31623ede97/ccm_web/client/src/pages/Home.tsx#L61)?

ccm_web/client/src/pages/Home.tsx
chrisrrowland: Yeah I'm not sure what do do here besides say the save failed.  It's not going to give any helpful feedback when its a validation sort of error, though.  More than that and its going to complicate things quite a bit I think.

ccm_web/client/src/api.ts
pushyamig: > What local react code are you referring to.. do you se an issue somewhere? Not this?

I mean to say when a course name get changes/updated then I feel it should be placed in the props. If SetCourse is the method, probably that is the place i guess. This suggestion is based on idea that courseName should be only pulled once during the launch of the tool and not when user navigates back to homepage via breadcrumbs

pushyamig: > @pushyamig get course seems to take about .5 sec or less, testing locally for me. Not insignificant but not huge. Do you think putting course info into globals should be spun off into its own issue?

I want to hear from @ssciolla on this. I am ok it leave it off as part of this issue.

ccm_web/client/src/api.ts
```diff
@@ -1,36 +1,63 @@
...
     return {
       credentials: 'include',
-      headers: { Authorization: 'Bearer ' + key }
+      headers: headers
     }
   }
   return {}
 }
 
+const getGet = (key: string | undefined): RequestInit => {
```
ssciolla, line 20: Nit-pick: `getGet` and `getPut` seem like they are missing a word. I think we're trying to avoid `Headers` because there is a `headers` property in this object we're creating. Maybe `getGetRequestConfig`? I dunno...

ccm_web/client/src/api.ts
```diff
@@ -1,36 +1,63 @@
+import { CanvasCourseBase } from './models/canvas'
 import { Globals, HelloMessageData } from './models/models'
 import handleErrors from './utils/handleErrors'
 
 export interface LtiProps {
   ltiKey: string | undefined
 }
 
-const createAuthHeaders = (key: string | undefined): RequestInit => {
+const initRequest = (key: string | undefined, headers: string[][] = []): RequestInit => {
```
ssciolla, line 9: In Python, using a mutable default parameter is an antipattern (I've been corrected on this before), and the recommended remedy is using a sentinel value. Not sure if it's an issue in JavaScript.

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

ccm_web/client/src/api.ts
```diff
@@ -1,36 +1,63 @@
...
   return {}
 }
 
+const getGet = (key: string | undefined): RequestInit => {
+  const request = initRequest(key)
+  request.method = 'GET'
+  return request
+}
+
+// This currently assumes all put requests have a json payload and recieve a json response
```
ssciolla, line 26:
```suggestion
// This currently assumes all put requests have a JSON payload and receive a JSON response.
```

ccm_web/client/src/api.ts
ssciolla: Hm, some interesting questions raised in this discussion. I also agree that we probably want to avoid spamming the API every time we navigate back to `Home`. One way to solve that would be by moving the hook and initial API call up to `App` and then passing the `set` action down to the `Home` page component.

I don't know if I like putting the course object in `globals`, because if the course name changes, then we either need to re-fetch globals or change the `globals` object. I'd like to not modify or fetch globals a bunch. Keeping the course fetching separate seems better to me, but I could probably be persuaded if someone feels strongly.

pushyamig:
There is a color difference in the course name title with and without edit roles. We should stick with grey color
Screen Shot 2021-07-06 at 11 47 43 AM
Screen Shot 2021-07-06 at 11 47 57 AM

ccm_web/client/src/pages/Home.tsx
```diff
@@ -61,18 +95,14 @@ function Home (props: HomeProps): JSX.Element {
...
-          reject(new Error('Rejected as requested'))
-        } else {
-          return resolve()
-        }
-      })
-      return (<InlineTextEdit {...{ text: 'Course 123ABC', save: saveCourseName, placeholderText: 'Course name', successMessage: 'Saved', failureMessage: 'Error saving course name' }}/>)
+    if (isLoadingCourse) {
+      return (<div className={classes.courseName}>Loading...</div>)
+    } else if (getCourseError !== undefined || course === undefined) {
+      return (<div className={classes.courseName}>Error getting course name</div>)
```
ssciolla, line 101: I think this is okay for now, but we probably want to look at more robust and re-usable loading (spinner?) and error displays. Seems like a separate issue.

ssciolla (commented):
Some initial comments so we can begin discussing. I think moving the hook up to `App` and passing the actions down to `Home` and `InlineTextEdit` seems like the way to go -- to avoid making unnecessary requests. Let me know what you think and/or if you have questions.

I'll do some testing this afternoon.

ccm_web/client/src/api.ts
```diff
@@ -1,36 +1,63 @@
...
+  headers.push(['Accept', 'application/json'])
+  const request = initRequest(key, headers)
+  request.method = 'PUT'
+  request.body = body
+  return request
+}
+
+export const getCourse = async (key: string | undefined, courseId: number): Promise<CanvasCourseBase> => {
+  const request = getGet(key)
+  const resp = await fetch(`/api/course/${courseId}/name`, request)
```
pushyamig, line 39: What should be the UI response when we fail getting the course? It is currently displaying as error getting the course name. Should we leave it blank?

ccm_web/client/src/models/FeatureUIData.tsx
```diff
@@ -80,11 +80,11 @@ const allFeatures: FeatureUIGroup[] = [
...
-  if (rolesAllowingAccess.length > 0) {
-    console.debug('Feature ' + featureDescriptionForLogging + ' allowed by ' + rolesAllowingAccess.join(','))
-  } else {
-    console.debug('Feature ' + featureDescriptionForLogging + ' forbid for ' + userRoles.join(','))
-  }
+  // if (rolesAllowingAccess.length > 0) {
+  //   console.debug('Feature ' + featureDescriptionForLogging + ' allowed by ' + rolesAllowingAccess.join(','))
+  // } else {
+  //   console.debug('Feature ' + featureDescriptionForLogging + ' forbid for ' + userRoles.join(','))
+  // }
```
pushyamig, line 87: These comments come be removed if not using it

ccm_web/client/src/api.ts
ssciolla: I'd think about rendering an error alert (some red box or other layout element) in the place of the title with a message about what happened.

ccm_web/client/src/pages/Home.tsx
```diff
@@ -33,7 +36,38 @@ interface HomeProps extends LtiProps {
...
+    void doLoadCourse()
+  }, [])
+
+  const setCourseNameAsync = async (
+    newCourseName: string
+  ): Promise<CanvasCourseBase|undefined> => {
+    return await apiSetCourseName(props.ltiKey, props.globals.course.id, newCourseName)
+  }
+
+  const [doSetCourseName, setCourseNameLoading, setCourseNameError] = usePromise<CanvasCourseBase|undefined, typeof setCourseNameAsync>(
```
pushyamig, line 58: `setCourseNameLoading` is not being used and linting error is shown? can be be used in the code?

ccm_web/client/src/api.ts
ssciolla: Hm, forgot about the snackbar, which may make such an alert repetitive. Some placeholder ("Unknown") might be okay.

ccm_web/client/src/api.ts
pushyamig: > One way to solve that would be by moving the hook and initial API call up to App and then passing the set action down to the Home page component.
I like  this idea!

> I don't know if I like putting the course object in globals, because if the course name changes, then we either need to re-fetch globals or change the globals object.

React [state-lifting](https://reactjs.org/docs/lifting-state-up.html) can be used with out refetching the globals

ccm_web/client/src/api.ts
chrisrrowland: Hmm, I don't think that is how javascript works.. in python that creates some sort of  global variable?

ccm_web/client/src/pages/Home.tsx
chrisrrowland: Agreed, the loading text is gone before I can even see it, so it might be better off just not showing anything rather than worrying about a spinner.

ccm_web/client/src/api.ts
chrisrrowland: I'm not a UX person but I feel like using the snackbar should be more for interactive things, rather than for an error happening on page load.. does that sound reasonable?  Just seems like something I have not come across in the wild.

ccm_web/client/src/pages/Home.tsx
chrisrrowland: It can, i don't imagine it will provide much functionality though :-p

ssciolla: Something we did in ROHQ was to have an aggregate `isLoading` variable that combined all the loading booleans like `setCourseNameLoading | getCourseLoading` and then we'd use that to determine whether to show a spinner in the bottom right hand corner. From a UX perspective, it's nice to know the site is doing some work. The problem of when it happens too quickly (which @chrisrrowland referenced elsewhere) is tricky, but I think this aggregate approach is sound.

ccm_web/client/src/api.ts
ssciolla: Yeah, I didn't see anything about this being a problem in JS, but I thought I'd flag it in case you knew something about it. I don't think it would be wrong to use `null` here as the default, but it may not be necessary.

ccm_web/client/src/api.ts
ssciolla: Ah, yeah, that seems valid. So you think think the in-place alert for automatic processes is okay?

pushyamig: >  does that sound reasonable?

Agreed

With that said we can discuss this with Janel and melinda. 

My 2 cents would be leave it blank for instructor user, and Admin user we can default it some name Course name ( who can edit). 

We can chat with Janel and Melinda once they come back

ccm_web/client/src/pages/Home.tsx
pushyamig: I feel @ssciolla suggested for a spinner and @chrisrrowland don't want that but leave it blank. I think there is some disagreement, unless I did not understood the comments.

I feel the spinner is there currently until get the course name

ssciolla: I think the spinner question can be discussed/handled in a separate issue. I'm just saying, instead of having separate (or separately styled) loading or error displays, we probably want to adopt a generic approach for that. It'll become more relevant once we're making other API calls. I'm fine with what @chrisrrowland has here for the time being.

ccm_web/client/src/api.ts
chrisrrowland: I moved the getCourse call into App

chrisrrowland:
> There is a color difference in the course name title with and without edit roles. We should stick with grey color
> Screen Shot 2021-07-06 at 11 47 43 AM
> Screen Shot 2021-07-06 at 11 47 57 AM

@janelilr I think the grey here was supposed to make it stand out as being editable.  I think I stole that style from an example, but I can't recall where.  Is there a standard practice around this?

ccm_web/client/src/pages/Home.tsx
chrisrrowland: This is no longer relevant after refactoring

ccm_web/client/src/utils/handleErrors.ts
pushyamig: This `resp.text()` is not holding any info about the cause of the error that BE sends  but instead this is more useful `resp.json()`

@ssciolla `resp.text() and resp.statusText()` is the default message from JS library have based on the status code. Why did you prefer this string over the BE sent message.

pushyamig: @ssciolla never mind I got confused.

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -78,20 +83,29 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
...
+    if (!props.isSaving) {
+      return (<EditIcon className={classes.editIcon} fontSize='inherit' onClick={toggleEdit}/>)
+    } else {
+      return (<EditIcon className={classes.editIconDisabled} fontSize='inherit'/>)
+    }
+  }
+
   return (
     <form className={classes.root} noValidate autoComplete="off">
+      Is Saving &quot;{props.isSaving.toString()}&quot;
```
pushyamig, line 96: Screen Shot 2021-07-07 at 3 59 24 PM

Something here is messing it up in the UI

ccm_web/client/src/api.ts
ssciolla: Thanks for making the change. In my head, I was also thinking that it might make sense to move the set API call up and pass that down, but you did exactly what I said and it works. I'm fine with what you did. However, let's watch out to see if it makes sense to move that other API action (`usePromise` hook for `setCourseName`) up in the future (it doesn't seem to yet).

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -85,20 +83,29 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
...
       <Grid container className={classes.inputArea}>
         <Grid item xs={8} sm={8}>
           <TextField className={classes.inputRow} aria-readonly={false} onClick={toggleEdit} inputProps={{ style: { fontSize: 24 } }} ref={textInput} id="standard-basic" placeholder={props.placeholderText} value={tempTextValue} onKeyDown={(e) => keyPress(e.code)} onChange={(e) => setTempTextValue(e.target.value)} disabled={!isEditing}/>
```
ssciolla, line 99: Should we also disable the `TextField` when `isSaving` is `true`?

ccm_web/client/src/pages/Home.tsx
```diff
@@ -30,11 +33,35 @@ const useStyles = makeStyles((theme) => ({
...
+    (course: CanvasCourseBase|undefined) => {
+      props.setCourse(course)
+      enqueueSnackbar('Course name saved', { variant: 'success' })
+    }
+  )
+
+  useEffect(() => {
+    if (setCourseNameError !== undefined) {
+      enqueueSnackbar('Error saving course name', { variant: 'error' })
+    }
```
ssciolla, line 62: At some point, we will need to think about how validation will work with the snackbar. The `error` object probably contains the messages about why it failed. Could these be exposed to the user, so they know how to respond (i.e. make it shorter, or more than 0 characters)? Should this be in the snackbar or somewhere more permanent? Do we want to add client-side validation? I'm open to different solutions; I think the first might be the simplest for now, but I could see a future requirement for on-type validation, which would mean some kind of FE code. I've done something similar before, so I'm happy to discuss the pros and cons, or share code.

It's okay with me pushing this off to a future issue/PR.

ccm_web/client/src/App.tsx
```diff
@@ -84,7 +107,7 @@ function App (props: AppProps): JSX.Element {
             </Route>
           </div>
           <Switch>
-            <Route exact={true} path="/" render={() => (<Home globals={globals} ltiKey={props.ltiKey} />)} />
+            <Route exact={true} path="/" render={() => (<Home globals={globals} ltiKey={props.ltiKey} course={course} setCourse={setCourse} />)} />
```
pushyamig: Can you please explain how is `setCourse` working here? I never passed around a useState set unction as another component. Your approach is working as expected with out having to make an getCourse API call. I couldn't wrap my head around it so I am asking here.

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -85,20 +83,29 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
     }
   }
 
+  const renderEditButton = (): JSX.Element => {
+    if (!props.isSaving) {
+      return (<EditIcon className={classes.editIcon} fontSize='inherit' onClick={toggleEdit}/>)
+    } else {
+      return (<EditIcon className={classes.editIconDisabled} fontSize='inherit'/>)
+    }
+  }
```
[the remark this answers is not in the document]
ssciolla: Another way to write this would be:

```
  const renderEditButton = (): JSX.Element => {
    return (
      <EditIcon
        className={props.isSaving ? classes.editIconDisabled : classes.editIcon}
        fontSize='inherit'
        onClick={props.isSaving ? undefined : toggleEdit}
      />
    )
  }
```

Not necessarily better, just thought I'd throw it out there.

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -39,31 +37,31 @@ const useStyles = makeStyles((theme) => ({
...
+    fontSize: '24px',
+    opacity: '75%'
+  },
   inputArea: {
     width: '500px'
   }
 }))
 
 function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
+  console.log('InlineTextEdit')
```
pushyamig, line 50:
```suggestion
  
```

ccm_web/client/src/api.ts
pushyamig: What is conclusion of what to display when fetching an getCourseName has an error? If we are waiting to here from @melindakraft and @janelilr on this, then we need to create an issue to deal with that later

ccm_web/client/src/App.tsx
```diff
@@ -32,8 +35,23 @@ function App (props: AppProps): JSX.Element {
...
+      return await getCourse(ltiKey, courseId)
+    },
+    (value: CanvasCourseBase|undefined) => setCourse(value)
+  )
+
+  useEffect(() => {
+    if (globals !== undefined) {
+      void doLoadCourse(props.ltiKey, globals.course.id)
+    }
+  }, [isGlobalsLoading])
```
ssciolla, line 52: I think we'll get the same effect but trigger this hook less often if we subscribe to changes to `globals` instead of `isGlobalsLoading`. With the latter we get three triggers (on mount, going to `true`, going to `false`), where as with the former, we just get two (on mount, going from `undefined` to `{...}`). Minor detail, but it occurred to me.

ssciolla (approved):
Nice work, @chrisrrowland! I think this is functional and could be merged. While I commented on a few significant things that are missing, I think these will all require some discussion and tackling them all in one PR would be a lot to ask. If you want help creating separate issues, or have questions about my comments, let me know.

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -85,20 +83,29 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
...
+    if (!props.isSaving) {
+      return (<EditIcon className={classes.editIcon} fontSize='inherit' onClick={toggleEdit}/>)
+    } else {
+      return (<EditIcon className={classes.editIconDisabled} fontSize='inherit'/>)
+    }
+  }
+
   return (
     <form className={classes.root} noValidate autoComplete="off">
+      Is Saving &quot;{props.isSaving.toString()}&quot;
```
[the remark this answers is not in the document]
chrisrrowland: I forgot to delete this after debugging

ccm_web/client/src/pages/Home.tsx
chrisrrowland: I gave it a little thought and wasn't sure how to distinguish between an error that contained a message that we would want to  show the user and one that we didn't.

My thought was that we'd have to return custom errors from the API calls and that would clash with the universal error handling in at least some small way.  If we can figure out a strategy  to handle that I feel like  it's best.  Secondary client side validation seems like an ok idea too, just would have to be manually synced up with the validation rules applied by canvas, and would require some additional enhancement of the InlineTextEdit component I assume.

ccm_web/client/src/App.tsx
chrisrrowland: I'm interested in this, I don't know the lifecycle well.  When I put a debugging statement in it seemed like this only got called once, and when I switched it to filter on changes to globals it got called  twice.  Was this a theoretical thing or did you observe something?

ccm_web/client/src/App.tsx
chrisrrowland: doLoadCourse gets called when globals are loaded, and then when the course name is changed from within the home component, it uses the setCourse handle passed in to update the course value in App.  The new value is coming from the value returned by the call to  set the course name.

ccm_web/client/src/App.tsx
ssciolla: I tested this, but could have missed something. Try putting a log statement inside `useEffect` but before the `globals !== undefined` check, and count how many times it's logged when you launch. I saw the 2 versus 3 difference, I thought.

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -85,20 +83,29 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
...
       <Grid container className={classes.inputArea}>
         <Grid item xs={8} sm={8}>
           <TextField className={classes.inputRow} aria-readonly={false} onClick={toggleEdit} inputProps={{ style: { fontSize: 24 } }} ref={textInput} id="standard-basic" placeholder={props.placeholderText} value={tempTextValue} onKeyDown={(e) => keyPress(e.code)} onChange={(e) => setTempTextValue(e.target.value)} disabled={!isEditing}/>
         </Grid>
```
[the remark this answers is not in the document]
chrisrrowland: More like so?

image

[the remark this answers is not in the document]
chrisrrowland: image

ccm_web/client/src/App.tsx
chrisrrowland: You're right, it was 3 vs 2, when it was 3 chrome grouped them up and put a 3 next to it that I missed.  I can switch it to use globals.

ccm_web/client/src/components/InlineTextEdit.tsx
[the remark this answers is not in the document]
ssciolla: Yeah, something like that. Can we fix it so that the combined length of the field and the buttons stays mostly even with the feature tiles below? Or alternatively, is there a way to make it adjust in size as you type (with a minimum and maximum)? Latter seems more complicated, unless it's built into `TextField`.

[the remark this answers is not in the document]
chrisrrowland: That was my thought too, the issue here is there are 2 buttons for each collection of features, but each row there is sized to fit 3, so if a 3rd feature gets added it will fill that area to the right and it will line up with the input.  I suppose that makes my argument more of a philosophical one than a real one since we're not adding a 3rd feature to any section.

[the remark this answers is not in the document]
ssciolla: There might be something you can do with `Grid` to use 2/3's of the space. I think even if we added a third feature, we wouldn't want it to be that long.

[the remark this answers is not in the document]
chrisrrowland: So it's using 2/3 of the space for the input, and 1/3 for the buttons.  
When you're editing seems like maybe when horizontal space is most important, and that's when there is more to display (the buttons)
Not sure what is ideal.
The original size I think was what the mockup looked like, but I agree that when I'm just messing around I often have names longer than the input, so not sure what is best..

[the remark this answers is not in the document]
ssciolla: Maybe we could just make it a little larger than what it is now, using a fixed length, and make sure it looks okay in different screen sizes, and then push off the discussion for later? Or we could leave it as is for now. I don't think this is the last time we'll think about text field styling.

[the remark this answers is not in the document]
chrisrrowland: I'll just leave it for now I guess.  Not really related to hooking up the backend anyways.

The only change required to make it behave as above is to add the attribute `fullWidth={true}` to TextField

ccm_web/client/src/App.tsx
janelilr: In regards to the above screenshot about the color difference in the course name title with and without edit roles, the color of text should be reversed. If a user is unable to edit, the text should be displayed in gray (indicates to users this cannot be changed). If a user can edit, the text would be black.

pushyamig: Thanks @janelilr. I will create a new issue with your suggestions.

PR #147, opened 2021-07-08
implement get sections API function (iss. #141)

### LINKED ISSUE #141: get course section API function
Provide an API function to get course sections. This will be used by features like:

* [Adding UMich users](https://github.com/tl-its-umich-edu/UX-Work/blob/master/ccm/wireframes/add_user_feature/add_um_user_sc.pdf)
* [Adding non-UMich users](https://github.com/tl-its-umich-edu/UX-Work/blob/master/ccm/wireframes/add_user_feature/add_nonUM_user_sc.pdf)
* [External tool Gradebook converter](https://github.com/tl-its-umich-edu/UX-Work/blob/master/ccm/wireframes/gradebook_feature/third_party_gradebook.pdf) 

The implementation must:

1. GET course sections from Canvas
  ~https​://umich.test.instructure.com/api/v1/courses/200?include[]=sections&include[]=total_students~
  The sample request given above **_WILL NOT_** give the number of students for each section.  It will give only the number of students in the **_COURSE_**.  Most likely, additional queries will need to be run to get the number of students in each section.
1. If Canvas' response is broken into multiple pages, assemble all the pages to be returned as a single response
1. Return for each section its name, ID, and number of enrollments

lsloan:
Resolves #141.

The implementation didn't need the course function of the Canvas API that was first suggested in the issue, with two `include[]` arguments.  That one only returned the student count for the entire course, not for each section.  Instead, the course-sections function is used, which only needs one `include[]` argument.

So, that's good-news/bad-news.  The good news is that we now know how to assign multiple values to parameters with the same name when using this request library.  The bad news, which is only very slightly bad, is that time was wasted figuring out the multiple values problem, when we ended up not needing it for this issue.

I'm not very satisfied with the solution for reducing the section data at [ccm_web/server/src/api/api.service.ts, lines 48–52](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/147/files#diff-43c0f2fb96f9f59e316e973ab22778a0aea7e79b4605246aca1fdb56dae06e66R48-R52).  I was hoping for something in which I only needed to specify the names of the keys to be kept.  There's a function in lodash for this, but there aren't any other cases in this project that need that module (yet).  I was thinking along the lines of a Python dictionary comprehension, but JavaScript doesn't have an equivalent.  The solution I thought of was…

```javascript
map(s => Object.fromEntries(Object.entries(s).filter(e => ['id', 'name', 'total_students'].includes(e[0]))))
```

But that's rather complex and not easy to understand.  So, although the solution I ended up with isn't as elegant as I'd like, it's easy to maintain.


lsloan:
Note that I've updated the requirements for #141:

> 1. GET course sections from Canvas
  ~https​://umich.test.instructure.com/api/v1/courses/423008?include[]=sections&include[]=total_students~
  The sample request given above **_WILL NOT_** give the number of students for each section.  It will give only the number of students in the **_COURSE_**.  Most likely, additional queries will need to be run to get the number of students in each section.

That API request will not the number of students for each section as shown in the mock-ups.

ccm_web/server/src/api/api.service.ts
```diff
@@ -40,16 +40,20 @@ export class APIService {
...
+      const courseResponse = await requestor.get<CanvasCourse>(courseEndpoint)
+      logger.debug(`Received response with status code ${courseResponse.statusCode}`)
+      const course = courseResponse.body
+
+      const sectionsEndpoint = `courses/${courseId}/sections`
+      const queryParams = {'include': ['total_students']}
+      logger.debug(`Sending request to Canvas - Endpoint: ${sectionsEndpoint}; Method: GET`)
+      const sectionsResponse = await requestor.get<object>(sectionsEndpoint, queryParams)
+      logger.debug(`Received response with status code ${sectionsResponse.statusCode}`)
+      const courseSections = sectionsResponse.body
```
pushyamig, line 54: This API itself is good enough and don't need the `courses/${courseId}`

lsloan: OK.  If you think so.  I thought the course name was also required.

If this is all that's needed, that will simplify things a little.

pushyamig: No, Section name and id, student count in each section is Good. 

Ideally we want all the users (teachers, students, TA etc) count in the section roster, but that info is not something easily available with this API `/course/:courseID/sections` with out making extra enrollment API call which we want to avoid.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -40,6 +40,10 @@ export interface CanvasCourse extends CanvasCourseBase {
   course_code: string
 }
 
+export interface CanvasCourseSection extends CanvasCourseBase {
```
lsloan, line 43: Technically, `CanvasCourseBase` should be renamed `CanvasBaseObject`.  Then, most references to `CanvasCourseBase` should be replaced with `CanvasCourse`.

What do you think, @ssciolla?

lsloan: `Entity` is a pretty good choice.

I suggested this because it didn't make sense to me that `CanvasCourseSection` extends `CanvasCourseBase`, because section objects don't inherit from course objects.  Now that we're discussing it, I guess there's no reason that extension needs to be declared.  I can declare `CanvasCourseSection` as a new, independent type.

I guess if we want to pursue the type cleanup, it should be a new issue.

lsloan:
So, this API function now returns (using my test course)…

```json
[
  {
    "id": 399389,
    "course_id": 374350,
    "name": "Section 2",
    "start_at": null,
    "end_at": null,
    "created_at": "2020-01-29T14:29:14Z",
    "restrict_enrollments_to_section_dates": false,
    "nonxlist_course_id": null,
    "sis_section_id": "lsloan-1111",
    "sis_course_id": "ccmC1599662699396-25",
    "integration_id": null,
    "sis_import_id": null,
    "total_students": 4
  },
  {
    "id": 518499,
    "course_id": 374350,
    "name": "Section V",
    "start_at": null,
    "end_at": null,
    "created_at": "2021-07-13T20:33:24Z",
    "restrict_enrollments_to_section_dates": false,
    "nonxlist_course_id": null,
    "sis_section_id": "lsloan-2222",
    "sis_course_id": "ccmC1599662699396-25",
    "integration_id": null,
    "sis_import_id": null,
    "total_students": 2
  }
]
```

Is this too much info?  Should I pare it down to `id`, `name`, and `total_students`, as specified in the requirements for #141 or would that be a waste of time?  I imagine I can use a some JS mapping function to remove the undesired properties.

lsloan:
After @ssciolla and I discussed my comment (https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/147#issuecomment-880002098) in Slack (https://app.slack.com/client/T69CPPGNP/G0130F688N6/thread/G0130F688N6-1626277548.215900), I've reduced the properties returned by the API to the following (again using my test course as an example)…

```json
[
  {
    "id": 399389,
    "name": "Section 2",
    "total_students": 4
  },
  {
    "id": 518499,
    "name": "Section V",
    "total_students": 2
  }
]
```

ccm_web/server/src/api/api.service.ts
```diff
@@ -37,6 +37,26 @@ export class APIService {
     }
   }
 
+  async getCourseSections(userLoginId: string, courseId: number): Promise<CanvasCourseSection[] | APIErrorData> {
+    const requestor = await this.canvasService.createRequestorForUser(userLoginId, '/api/v1/')
+    try {
+      const endpoint = `courses/${courseId}/sections`
+      const queryParams = {'include': ['total_students']} // use list for "include" values
+      logger.debug(`Sending request to Canvas - Endpoint: ${endpoint}; Method: GET`)
+      const response = await requestor.get<CanvasCourseSection[]>(endpoint, queryParams)
```
ssciolla, line 46: Pagination? Seems to have been requested on the issue. I haven't tried this, but maybe `list` method will make it easy? https://github.com/KTH/canvas-api/blob/master/docs/API.md#listendpoint-queryparams-- 

I started trying this out, seems to work, but I think we need to modify the types for `@kth/canvas-api` (my initial types based on the auto-generated ones seem to be wrong). I'll get back to you, may open a supporting PR to the main repo or to your branch.

```
      const something = requestor.list(endpoint, queryParams)
      console.log(something)
      const allSections = await something.toArray() as CanvasCourseSection[]
      console.log(allSections)
      const sections = allSections.map(s => ({
        id: s.id,
        name: s.name,
        total_students: s.total_students
      }))
```

Canvas page default is 10 I believe, but we can set `per_page` to be something more reasonable, maybe 50? I'd make a global variable for that in `APIService`.

ccm_web/server/src/api/api.service.ts
```diff
@@ -37,6 +37,26 @@ export class APIService {
...
+      logger.debug(`Received response with status code ${response.statusCode}`)
+      const sections = response.body.map(s => ({
+        'id': s.id,
+        'name': s.name,
+        'total_students': s.total_students
+      }))
+
```
ssciolla, line 53:
```suggestion
```

ssciolla (changes requested):
Looks pretty good so far, but we're missing a few things: scope and pagination. I responded to your thread in Slack about specifying the scope. It might make sense to create a small section in the `README.md` about modifying the scopes? I dunno. Feel free to reach out to me tomorrow if you want to talk about scopes, pagination, linting, etc.

ccm_web/server/src/api/api.service.ts
pushyamig: Yeah we want to add pagination support to it. Since FE will just ask for section in single call and BE pages throughs all sections and send it. We discussed this in the slack channel as well. 
I think we should set it to `per_page = 100`, since that is the Max we can expect from canvas and for getting section info it should be fast.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -40,6 +40,12 @@ export interface CanvasCourse extends CanvasCourseBase {
   course_code: string
 }
 
+export interface CanvasCourseSection {
+  id: number
+  name: string
+  total_students: number
+}
```
pushyamig, line 47: I  think since My PR https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/154/files is also needs sections object, this interface need to extend the section interface

How about this?
```
export interface CanvasCourseSection {
 id: number
 name: string
}

export interface CanvasCourseSectionWithStudents extends CanvasCourseSection {
  total_students: number
}
```
If you agree that I will make changes to my PR

ccm_web/server/src/api/api.service.ts
chrisrrowland: I might suggest setting per_page to 50% or 100% more than what you might expect to be the max, since it sounds like we're effectively eliminating paging altogether by setting the limit to the expected max.

Setting it higher would have no negative effect on performance and avoid the occasional pagination if there were 101 sections or something.

lsloan: I'm guilty of assuming the pagination was being handled by KTH.  I didn't look into it.  I'll do so.

Any recommendations of Canvas courses that have enough sections to trigger pagination in the response from the API?

ccm_web/server/src/api/api.service.ts
```diff
@@ -37,6 +37,26 @@ export class APIService {
...
+      const endpoint = `courses/${courseId}/sections`
+      const queryParams = {'include': ['total_students']} // use list for "include" values
+      logger.debug(`Sending request to Canvas - Endpoint: ${endpoint}; Method: GET`)
+      const response = await requestor.get<CanvasCourseSection[]>(endpoint, queryParams)
+      logger.debug(`Received response with status code ${response.statusCode}`)
+      const sections = response.body.map(s => ({
+        'id': s.id,
+        'name': s.name,
+        'total_students': s.total_students
+      }))
```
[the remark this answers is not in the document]
lsloan: I didn't see them.  Sure, we can look at my linter settings.  Do we have a CLI way of running the linter?

ccm_web/server/src/canvas/canvas.interfaces.ts
lsloan: No, I think maybe we should make `total_students` optional, then both functions can use the same type.

ccm_web/server/src/api/api.service.ts
ssciolla: Keep the `per_page` default, add 11 sections to your course?

ccm_web/server/src/api/api.service.ts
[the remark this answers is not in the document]
ssciolla: Not sure I've done it before, but there is a CLI for `eslint`

ccm_web/server/src/api/api.service.ts
ssciolla: @lsloan, I opened a PR against your branch with the type changes and I think a work-able pagination solution. I also included the scope change I requested, since I needed it for testing. See https://github.com/lsloan/ccm/pull/1/files

Sorry for maybe stepping on your toes here, I just felt responsible for the wrong types and wanted to fix them. I will create an issue about that this afternoon so that you can link it to your PR (assuming these changes are OK, get merged).

lsloan: OK…  I thought there might be an existing course.

ssciolla: As promised, I've created an issue for updating the types (#158), and linked it to this PR.

ssciolla (commented):
I contributed a commit to this, so I will let others finish approving and testing. Looks like there may be a few other minor outstanding things to address or tidy up.

ssciolla (approved):

ccm_web/server/src/api/api.service.ts
lsloan: I've accepted @ssciolla's PR to my branch that included code for handling Canvas API pagination.

ccm_web/server/src/api/api.service.ts
```diff
@@ -25,6 +25,29 @@ export class APIService {
...
+    const requestor = await this.canvasService.createRequestorForUser(userLoginId, '/api/v1/')
+    try {
+      const endpoint = `courses/${courseId}/sections`
+      const queryParams = { 'include': ['total_students'] } // use list for "include" values
+      logger.debug(`Sending request to Canvas (get all pages) - Endpoint: ${endpoint}; Method: GET`)
+      // FIXME: list() should return promise, toArray() should be callable later
+      const sectionsFull = await requestor.list<CanvasCourseSection>(endpoint, queryParams).toArray()
+      // FIXME: no access to got Response; statusCode, etc. not available!
+      // logger.debug(`Received response with status code ${sectionsFull.statusCode}`) // broken
+      const sections = sectionsFull.map(s => ({
```
lsloan, line 38: Although my TypeScript experience is limited, it seems that the declaration of `list()` and its `toArray()` method could be better.

The way this Canvas API function was called before, without pagination handling via `get()`, the got `Response` object that was returned had a `statusCode` property.  It seems that this `list()` method doesn't return that type of object, so that property (and several others that would be useful) isn't available.

Also, it seems that the `toArray()` call here is required, because that is actually returning the promise.  It seems to me that `list()` should return a promise for a got `Response` object **_and_** the `toArray()` method should be callable at any time on that object.  That is, maybe in some case the code may want to use the `Response` object for some operations, then its `toArray()` method could be called later to get the response contents as an array.

For example, I'd expect it to be used this way…

```suggestion
      // FIXME: list() should return promise, toArray() should be callable later
      const response = await requestor.list<CanvasCourseSection>(endpoint, queryParams)
      // FIXME: no access to got Response; statusCode, etc. not available!
      logger.debug(`Received response with status code ${response.statusCode}`)
      const sections = response.toArray().map(s => ({
```

I will open a new issue for this.

lsloan (commented):
Commenting on changes that will go into a new issue.

ccm_web/server/src/api/api.service.ts
lsloan: Opened issue #159.

lsloan:
:shipit:

PR #148, opened 2021-07-09
Manage user authz using NestJS and Passport, log in through `ltijs` (#139, #166)

### LINKED ISSUE #139: Use NestJS to handle user authentication, instead of relying solely on LTIJS
Currently the NestJs backend knows the user/launch info via Express session approach and LTIJS also taking care of this. It won't allow any call if it don't have the token (ltik) attached to request to backend. If you make Backend request with ltik from postman it know this is not user session probably this is the cookie approach it is taking. LTIJS being the gatekeeper than a middle man.

This issue  will set the baseline for authentication mechanism for the project by permeating user/launch info in the Backend.

ssciolla:
This PR represents a refactor of authentication in the application -- essentially taking away route protection responsibilities from `ltijs` and only using that library for the log in routine. 

The new `AuthModule` makes use of Passport and Nest libraries (see the NestJS Authentication chapter under resources below, especially the JWT sections). The JWT lives in a cookie that is only readable by the backend (`httpOnly`) and is extracted and checked by a route guard. This approach also allows us to attach the `User` model instance for the user making the request to each `Request` object, allowing for backend pattern improvements.

The `AuthModule` is 'hooked' together with the `ltijs` process by injecting `AuthService` into the `LTIService` and then calling a `AuthService.loginLTI` method that mints and sets the JWT cookie before `LTIService` redirects to the main application.

CSRF is also implemented using the `csurf` library and the ~~"double submit cookie pattern".~~ documented session approach. It works according to my current (perhaps inaccurate) understanding as follows: The `csurf` middleware adds a `csrfSecret` property to the session data and then we set a non-`httpOnly` cookie (see `/auth/csrf` in `AuthController`) containing another secret (made available by the middleware at `req.csrfToken()`). The client JS code then grabs the CSRF token from the cookie and then submits with `POST` and `PUT` requests in the `CSRF-Token` header. The middleware then compares (using some algorithm) the secret in the session with the token from the header. See the CSRF resources section below for more info.

All requests will still work in the Swagger UI, but you will need to grab the CSRF Token from the `csrfToken` URL parameter -- as we did with `ltiKey` before -- and add it using the "Authorize" button in order to make `POST` or `PUT` requests.

The PR aims to resolve #139 and #166.

Resource(s):
- NestJS/Passport
  - https://docs.nestjs.com/security/authentication
  - http://www.passportjs.org/packages/passport-jwt/
  - https://docs.nestjs.com/custom-decorators
- CSRF
  - https://www.npmjs.com/package/csurf
  - https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#token-based-mitigation
  - https://stackoverflow.com/a/59549909

Tasks
- [x] Implement `passport` with JWT strategy and login through `ltijs`
- [x] Remove `ltiKey` stuff from the frontend
- [x] Implement CSRF protection (`csurf`) for `POST` and `PUT` calls
- [x] Figure out how CSRF works with Swagger (included hack similar to previous `ltiKey` solution so far; we may want to look into [`swagger-ui-react`](https://www.npmjs.com/package/swagger-ui-react))
- [x] Throw `ForbiddenError` when missing CSRF token (currently `500`)
- [x] Hide some `/canvas` and `/auth` routes from Swagger
- [x] Update README as necessary
- [x] Fix end-to-end test
- [x] Improve typing in `jwt.strategy.ts` (didn't quite resolve this; will comment to that effect)
- [x] Add `helmet` to non-`ltijs` routes 
- [x] Set cookie and token expiration (create `MAX_AGE_IN_SEC` environment variable).
- [ ] Try to break it


ccm_web/client/src/hooks/useGlobals.ts
```diff
@@ -4,32 +4,44 @@ import usePromise from './usePromise'
...
-  if (ltiKey !== undefined && globals !== undefined) {
+  if (globals !== undefined) {
     isAuthenticated = true
   } else if (getGlobalsError !== undefined) {
     isAuthenticated = false
   }
 
-  return [globals, isAuthenticated, getGlobalsLoading, getGlobalsError]
+  const loading = getGlobalsLoading || setCSRFTokenCookieLoading
+  return [globals, isAuthenticated, loading, getGlobalsError, setCSRFTokenCookieError]
```
ssciolla, line 44: Does it seem reasonable to handle the CSRF cookie setting in `useGlobals`? Or should it be somewhere else?

ccm_web/server/src/auth/jwt.strategy.ts
```diff
@@ -0,0 +1,39 @@
...
+    private readonly configService: ConfigService,
+    private readonly userService: UserService
+  ) {
+    super({
+      jwtFromRequest: (req: Request): string | null => {
+        if (req.signedCookies.jwt !== undefined) {
+          const token = req.signedCookies.jwt as string
+          return token
+        }
+        return null
```
ssciolla, line 24: This extracts the JWT token from the signed cookie. I tried to improve the types here (i.e. make TS aware that `request.signedCookies` is an object with certain keys), but I couldn't find a good way to do so.

pushyamig:
I will review this first thing in the morning

ccm_web/server/src/main.ts
```diff
@@ -60,13 +62,17 @@ async function bootstrap (): Promise<void> {
...
+        'CSRF-Token', {
+          type: 'apiKey',
+          name: 'CSRF-Token',
+          in: 'header',
+          description: (
+            'POST and PUT requests need to include a CSRF-Token header. ' +
+            'The token can be found in the "csrfToken" URL parameter ' +
+            '(use a browser tool to view the URL of the frame).'
+          )
+        })
```
ssciolla, line 75: I did a fair amount of research on how to set this up so we don't have to add the token manually, but I hit several deadends. I'm inclined to keep it like this for now and open a separate issue (I will do that anyway and add the links that I found related to this). One possible solution is to use `swagger-react-ui` (since it's on the client side, it can access the token more easily), but we'd need to serve the Swagger document somehow, which is an extra step I haven't thought through.

ssciolla: See https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/167

pushyamig:
The PR seems to be functional from UI testing perspective. The Get/PUT call are fetching and changing data as intended. The Swagger setup is pretty functional. I will start looking into code now.

But one weirdness I have seen is with Swagger/CSRF token authorize. When I  grab an any CRSF token(Like past token) that was generated over each LTI launch and try to authorize and make a API call like PUT and Get ( from Swagger UI) it is fetching result. I suppose it shouldn't since past CSRF token should not work. But if you look at the same API call request from the Browser console it is attaching a newly/correct generated CSRF token. This is not necessarily an issue, some how swagger Authorize button taking in any CSRF token generated in past ( won't take when I punch in some random token) but when making PUT call will attach correct CSRF token ~**No changes needed here, since this only affects local and I'm sharing my experience.**

ccm_web/server/src/auth/jwt.strategy.ts
```diff
@@ -0,0 +1,39 @@
...
+export class JwtStrategy extends PassportStrategy(PassportJwtStrategy) {
+  constructor (
+    private readonly configService: ConfigService,
+    private readonly userService: UserService
+  ) {
+    super({
+      jwtFromRequest: (req: Request): string | null => {
+        if (req.signedCookies.jwt !== undefined) {
+          const token = req.signedCookies.jwt as string
+          return token
```
pushyamig, line 22: I looked some of the Req/Res for Course API calls from the Browser console, and I don't see that you are passing this JWT token as bearer token with request from client, but more attached this as a Cookie to requests. By reading through the [documentation](https://docs.nestjs.com/security/authentication) from nestJS they shared bearer approach, there is no right or wrong here. But what was you thinking here, if I understood  this correctly 

here is some context from what I mean from documentation....
```
$ # GET /profile
$ curl http://localhost:3000/profile
$ # result -> {"statusCode":401,"error":"Unauthorized"}

$ # POST /auth/login
$ curl -X POST http://localhost:3000/auth/login -d '{"username": "john", "password": "changeme"}' -H "Content-Type: application/json"
$ # result -> {"access_token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2Vybm... }

$ # GET /profile using access_token returned from previous step as bearer code
$ curl http://localhost:3000/profile -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2Vybm..."
$ # result -> {"userId":1,"username":"john"}
```

ccm_web/server/src/auth/auth.module.ts
```diff
@@ -0,0 +1,42 @@
...
+  imports: [
+    UserModule,
+    PassportModule,
+    JwtModule.registerAsync({
+      imports: [ConfigModule],
+      inject: [ConfigService],
+      useFactory: async (configService: ConfigService) => ({
+        secret: configService.get<string>('server.tokenSecret'),
+        // JWT tokens will expire after 24 hours.
+        signOptions: { expiresIn: 60 * 60 * 24 }
```
pushyamig, line 23: This is nice the you set an expiration, 24 hours expiration seems ok. we could also do 12 hrs. I don't know if that is big of a deal though. 24 vs 12

ccm_web/server/src/auth/jwt.strategy.ts
ssciolla: Based on the research I did, I was operating with the understanding that it was more secure to store the JWT in a cookie than any other way, but that it did open you up to CSRF attacks -- hence the need for the `csurf` solution. I open to changing this, but we need to be careful regardless to ensure we're covering our bases. @jonespm, any thoughts?

Here's a couple places I read about the preference for storing JWTs in cookies (including some I just found that echo similar sentiments):
- https://dev.to/gkoniaris/how-to-securely-store-jwt-tokens-51cf
- https://medium.com/@ryanchenkie_40935/react-authentication-how-to-store-jwt-in-a-cookie-346519310e81
- https://stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage
- https://indepth.dev/posts/1382/localstorage-vs-cookies

ccm_web/server/src/app.module.ts
```diff
@@ -39,9 +42,20 @@ const logger = baseLogger.child({ filePath: __filename})
...
+    consumer.apply(helmet())
+    // Exclude ltijs routes, which are already protected by helmet
+      .exclude(
+        { path: '/lti', method: RequestMethod.POST },
+        { path: '/lti', method: RequestMethod.GET },
+        '/lti/(.*)'
+      )
+      .forRoutes('*')
+  }
+}
```
ssciolla, line 61: I just added this, which should ensure all route are covered by `helmet` without providing redundancy. It made sense to do it in this PR since the decoupling caused the lack of protection. I will link the other issue to this PR as well.

ssciolla: You can see which routes this is being applied to by replacing the `.apply(helmet()` with something like the following:
```
      .apply((req: Request, res: Response, next: NextFunction) => {
        console.log(req.method)
        console.log(req.baseUrl)
        helmet()(req, res, next)
      })

```

ssciolla:
> But one weirdness I have seen is with Swagger/CSRF token authorize. When I grab an any CRSF token(Like past token) that was generated over each LTI launch and try to authorize and make a API call like PUT and Get ( from Swagger UI) it is fetching result. I suppose it shouldn't since past CSRF token should not work. But if you look at the same API call request from the Browser console it is attaching a newly/correct generated CSRF token. This is not necessarily an issue, some how swagger Authorize button taking in any CSRF token generated in past ( won't take when I punch in some random token) but when making PUT call will attach correct CSRF token

@pushyamig, that's an interesting observation that I hadn't thought about. This seems to be the way `csurf` is implemented, and I didn't see any options about invalidating old tokens or whatnot. These links might be relevant to the question.
- https://stackoverflow.com/questions/57812757/how-csurf-middleware-validates-tokens
- https://stackoverflow.com/questions/41120046/should-i-be-able-to-re-use-csrf-tokens-within-the-same-session-when-using-npm-cs

ccm_web/server/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+export class AuthController {
+  constructor (private readonly authService: AuthService) {}
+
+  @ApiExcludeEndpoint()
+  @Get('csrfToken')
+  async setCSRFTokenCookie (
+    @Req() req: Request, @Res({ passthrough: true }) res: Response
+  ): Promise<void> {
+    // Cookie options deliberately include defaults of httpOnly false and signed false.
+    res.cookie('CSRF-Token', req.csrfToken(), this.authService.commonCookieOptions)
```
ssciolla, line 19: We could also just send the token in the response here. I'm seeing people talk about doing it either way.

lsloan: What does it mean to send it in the response?  It's a value that the client must add to subsequent requests itself?  Setting a cookie is like the automatic way of specifying values that must be included in requests.  I think that's a good way to go.

ccm_web/server/src/auth/auth.interfaces.ts
```diff
@@ -0,0 +1,10 @@
+export interface JwtPayload {
+  username: string
+  sub: bigint
+  iat: number
+  exp: number
+}
+
+export interface MaybeCSRFError extends Error {
```
lsloan, line 8: "`Maybe`"‽ 😆

ccm_web/server/src/auth/auth.module.ts
lsloan: 24 hours is good.  This could be set via an application configuration option if you think this would ever need to be changed.  I'm thinking ahead to organizations other than U-M that may want to use this app and may have different auth policies.

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -43,17 +44,15 @@ export class LTIService implements BeforeApplicationShutdown {
       { plugin: dbPlugin },
       {
         // Reserved routes
-        appRoute: '/lti',
-        loginRoute: '/lti/login',
-        keysetRoute: '/lti/keys',
+        appRoute: '/',
+        loginRoute: '/login',
+        keysetRoute: '/keys',
```
lsloan, line 49: Do these changes mean the URLs in the LTI config used for the LMS will need to change?  Or are these new routes relative to the `/lti` route?

ccm_web/server/src/main.ts
lsloan: If that makes using the Swagger UI easier, it will be nice to see.

ccm_web/server/test/app.e2e-spec.ts
```diff
@@ -12,25 +13,17 @@ describe('APIController (e2e)', () => {
     }).compile()
 
     app = moduleFixture.createNestApplication()
+    app.use(cookieParser('SOME_COOKIE_SECRET'))
```
lsloan, line 16: Why this specific value, `'SOME_COOKIE_SECRET'`?  Should this be gotten from the application configuration?

ccm_web/server/src/api/api.controller.ts
```diff
@@ -1,60 +1,61 @@
...
+    @Param('id', ParseIntPipe) courseId: number, @UserDec() user: User
   ): Promise<CanvasCourseBase> {
-    const { userLoginId } = session.data
-    const result = await this.apiService.getCourseName(userLoginId, courseId)
+    const result = await this.apiService.getCourseName(user, courseId)
     if (isAPIErrorData(result)) throw new HttpException(result, result.statusCode)
     return result
   }
 
+  @ApiSecurity('CSRF-Token')
```
lsloan, line 44: Why don't the various `GET` API entries also use CSRF security?

lsloan (commented):
Fairly complex.  I can tell it was no easy task to understand the interactions of CSRF, LTI auth, and getting it to work with the framework.  I like that it makes the parts of the code that's most likely to change (e.g., API entries) a little simpler.

ccm_web/server/src/lti/lti.service.ts
ssciolla: Nothing should need to change in the configuration. As you said, these should be relative to `/lti/`. You can see where I have applied this as middleware to that route root in `lti.module.ts`. (It is working, but maybe there's a more correct way to specify this in `forRoutes`; I'll look into it.)

ccm_web/server/test/app.e2e-spec.ts
ssciolla: I mean, this is a test, so I don't think it really needs the actual configuration value. I think I'll probably re-work some of this eventually, but this makes it work for now, since the test module has access to the cookie properties off of `express.Request`.

ccm_web/server/src/api/api.controller.ts
ssciolla: My understanding is that `GET` requests are not vulnerable in the same way to CSRF attacks. I think these attacks basically try to embed mutating requests to your application in a form on another site, so that it can take advantage of the JWT you have in a cookie that will get sent automatically. However, because of how that works, the other site doesn't have easy access to the response. 

"CSRF attacks target functionality that causes a state change on the server, such as changing the victim’s email address or password, or purchasing something. Forcing the victim to retrieve data doesn’t benefit an attacker because the attacker doesn’t receive the response, the victim does. As such, CSRF attacks target state-changing requests."
- From https://owasp.org/www-community/attacks/csrf

That said, maybe the `GET` calls for `/canvas/*` could benefit from CSRF protection (at least `canvas/returnFromOAuth` could be considered mutating), but I think there is additional security baked in there with the `state` variable. I'll have to think about that some more.

ccm_web/server/src/auth/auth.module.ts
ssciolla: I read somewhere that the cookie should also use the same expiration. I'm gonna look into how to set that.

ccm_web/server/src/auth/auth.controller.ts
ssciolla: Basically, I need to get this token to the client somehow. I'm not sure it matters particularly whether it's a cookie or in the response, because ultimately it needs to be added by client JS code to the `CSRF-Token` header for requests. Why? That's how CSRF protection seems to work. If the token were just read automatically from a cookie by the backend (like currently we're doing with the JWT), then it would also be vulnerable to CSRF attacks.

ccm_web/server/src/lti/lti.service.ts
lsloan: If it's working, you don't need to change it on my account.  I just needed a little clarification to be sure I understood how it was working.  If anything is needed, just adding a comment to say these paths are relative to `/lti` would be sufficient.

ccm_web/server/test/app.e2e-spec.ts
lsloan: I saw in some documentation that `cookieParser()` is called without arguments.  Does that mean a value is optional here?  If so, would no secret make the cookies less secure?

ccm_web/server/src/auth/auth.controller.ts
lsloan: Sure, I understand it needs to get to the client.

I think what you were saying in [your earlier comment](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/148#discussion_r679589493) is that the token can be sent in a cookie as you're doing now, or "in the response".  I'm not certain what the latter entails.  However, it sounds as though it would require the client to do something special with it when it needs to make a request.  Using a cookie avoids those special steps.

Also, I think some of the other TL apps use cookies for CSRF and it works well.

ccm_web/server/src/auth/auth.interfaces.ts
ssciolla: Well, it's only possibly due to a CSRF error. The CSRF middleware throws an error/exception with a `code` property, so I wanted to type it with that property being optional.

ccm_web/server/src/auth/auth.controller.ts
ssciolla: So it's either do what I'm doing now, in the client, pull the CSRF token from the non-`httpOnly` cookie (as I'm doing now in `client/src/api.ts`. Or return the csrfToken in `response.body` as we would any other data from a `GET` request. This would require a bit more code on the client-side, though we wouldn't need the `js-cookie` library I'm adding. Basically we'd do something like what the author of [this article](https://medium.com/@ryanchenkie_40935/react-authentication-how-to-store-jwt-in-a-cookie-346519310e81) is doing towards the end.

I'm honestly not sure what the security implications are for either approach. This [SO answer](https://stackoverflow.com/a/59549909) seems to suggest doing it as I'm doing it now, via cookie readable by the client.

lsloan: I looked at the code again and I see that I made a faulty assumption.  I was thinking that since the token was being set in a cookie, then that cookie would automatically be included when the UI makes a request to the API.  I see now where it is being set in the `CSRF-Token` header for each API request.

Now I understand better that the UI code **_must do something_**  to add the token to the request, regardless of the way it gets the token. So, the answer to your question whether the token should be set in a cookie is a little less obvious to me now.  I guess being able to simply get the token using `Cookies.get('CSRF-Token')` is pretty convenient.  Putting the token in a cookie might make it a little easier for an attacker to find it, so putting it in the response might provide more "security through obscurity".  I'm not sure, but I suspect if the token were sent as a part of the response, it would be a little more work to get it.

So, in the article by Chenkie you referenced, is the part you referred to the "[Getting and Setting the CSRF Token](https://medium.com/@ryanchenkie_40935/react-authentication-how-to-store-jwt-in-a-cookie-346519310e81#:~:text=getting%20and%20setting%20the%20csrf%20token)" section?  In there, it looks like the code is using a `csrf-token` service or API entry to get the token.  Would CCM need to do that as well?  Is that service or API entry provided by one of the modules you've installed or would that need to be added to CCM's code itself?

ssciolla: Cool, sounds like we're more on the same page now. Yes, that is the section I'm referencing, although there is code above that that's relevant.

As far as I understand, `csurf` makes the token needed by the client available on an `express.Request` object, so it's up to the application programmer to get that token to the client (either through a cookie or in the response body). Yes, to use the response body, we'd have to call that endpoint and probably pass the CSRF token around the React components, instead of treating it more like a global (which I'm doing now). I can do that, but my current approach is probably less lines of code. The author gets around some of this by using an `axios` default header, but we're using `fetch`...

> Putting the token in a cookie might make it a little easier for an attacker to find it, so putting it in the response might provide more "security through obscurity". I'm not sure, but I suspect if the token were sent as a part of the response, it would be a little more work to get it.

This kinda makes sense; I'd be interested to see references supporting your argument. I'm guessing a sophisticated enough attacker could find it either way.

pushyamig:
The Approach adopted here with Auth, JWT and CSRF token, I don't see any concerns. Lot of these is new to the team and everybody is learning as we progressing. @ssciolla Really nice work and pulling it all together fairly quickly. We will be doing penetration test in September and we will find out more.

I tested the PR with both local and Prod build and i couldn't break it. I have few comment left nothing major that need stop this PR from merging. I will leave upto @lsloan, @chrisrrowland and @jonespm for giving a formal approval.

jonespm (commented):
It looks like you've done a lot of research on this and after walking through it I don't have anything I'd suggest to change here.

I think my only question is if we still even need CSRF. It sounds like if you just want to [prevent CSRF in modern browsers](https://www.sjoerdlangkemper.nl/2019/02/27/prevent-csrf-with-the-origin-http-request-header/) we can just use that now. It really just still exists for the 1-2% of browsers out there that people might use that are older? I tested on Firefox/Safari and Chrome from that 2019 site and all of he bugs mentioned were resolved. Did you come across this at all in your research?

[Implementing in CORS Origin Express looks much easier than CSRF](https://www.section.io/engineering-education/how-to-use-cors-in-nodejs-with-express/)

ccm_web/server/test/app.e2e-spec.ts
ssciolla: I believe so? I'm not sure I can omit the secret in the end-to-end test, because the extractor function in `jwt.strategy.ts` checks `signedCookies`, not `cookies...

ccm_web/server/src/auth/auth.controller.ts
ssciolla: @lsloan, I think pulling a CSRF token -- subsequently submitted via a custom header -- from a cookie using client code seems to be a standard approach, used by Angular for instance (https://angular.io/guide/http#security-xsrf-protection). See also https://en.wikipedia.org/wiki/Cross-site_request_forgery#Cookie-to-header_token

Because of the greater FE simplicity, I'm leaning towards not changing this. Please let me know if you have strong feelings that this is wrong or a significant security concern.

ssciolla:
> It looks like you've done a lot of research on this and after walking through it I don't have anything I'd suggest to change here.

Thank you very much, @jonespm, for taking the time to look at this on Friday, I do appreciate it.

> I think my only question is if we still even need CSRF. It sounds like if you just want to [prevent CSRF in modern browsers](https://www.sjoerdlangkemper.nl/2019/02/27/prevent-csrf-with-the-origin-http-request-header/) we can just use that now. It really just still exists for the 1-2% of browsers out there that people might use that are older? I tested on Firefox/Safari and Chrome from that 2019 site and all of he bugs mentioned were resolved. Did you come across this at all in your research?
> 
> [Implementing in CORS Origin Express looks much easier than CSRF](https://www.section.io/engineering-education/how-to-use-cors-in-nodejs-with-express/)

This comment sent me down a rabbit-hole this morning, with stops on several SO posts and other resources. I hadn't thought about this much before, so I'm glad you did bring it up, for my education. (For what it's worth) OWASP doesn't really flag a proper CORS setup as a replacement for CSRF tokens, just as [a mitigation strategy](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md#introduction). They may just be behind the times, but it does make sense when you're talking about security to try to cover all your bases. I'm still a little confused about some of the details, such as a) are "simple" requests prevented by a strict CORS policy see [SO answer and its comment thread](https://security.stackexchange.com/a/97938) and b) whether a SOP (Same-origin policy) is set by default in browsers.

I guess my inclination would be to err on the side of caution, until we get a clear message from the industry that this is an antiquated way of doing things. The fact that Django, Angular, Nest, etc. are all talking about CSRF token functionality and implementation signals to me that we're not quite there yet. Let me know if that thinking makes sense to you.

I do think I will implement an explicit CORS policy allowing only requests from certain Canvas URLs, as a mitigation strategy.

lsloan (approved):
After much helpful discussion, I think the code is ready to go.

:shipit:

jonespm:
I think I might have went down the same rabbit hole. I know that many browser security issues that had hacky workarounds in the past have been mitigated directly by the browsers over time. For instance iframe sandbox and cookie issues like SameSite. I saw something mention it just as a mitigation strategy but I did remember seeing on page said it was a replacement strategy. 

In legacy apps many left this code in there since it doesn't hurt anything, but in newer apps it seems like all it would give you is compatibility with an increasing smaller number or legacy browsers, none of which are supported anymore by Canvas. The risk if you don't have CSRF seemed to be that if someone was somehow using those old browsers and were directly attacked they wouldn't be protected. Though that's super low risk in the case of CCM app.

I know for MyLA I did have to add a CSP header mostly to replace the deprecated X_FRAME_OPTIONS but weren't using any origin policy.

ccm_web/server/src/api/api.controller.ts
ssciolla: I looked into this a bit more, and the `state` parameter seems to take the place of the CSRF token to prevent attacks here. I will separately research and vet the use of the `state` parameter to ensure we're covered here.
See https://auth0.com/docs/protocols/state-parameters#set-and-compare-state-parameter-values and https://security.stackexchange.com/questions/20187/oauth2-cross-site-request-forgery-and-state-parameter

chrisrrowland (approved):
So much going on here, I have only begun to scrape the surface from an understanding perspective.  Functionally outside of an initial blip which I think was just  stale data in my browser things have  worked  perfectly as far as I could tell.

ccm_web/server/src/auth/auth.module.ts
ssciolla: Okay, I've added a `MAX_AGE_IN_SEC` environment variable that I'm using for the expiration here, as well as for the cookie contaning the JWT and the `CSRF-Token`cookie.

There are some repercussions to adding a max age to the session cookie, such as the data no longer being available when needed (from `/api/globals` for instance). I'm going to tackle this separately; I've opened issue #171 to keep track of session-related questions I have.

ssciolla:
FYI, especially @jonespm and @pushyamig since they asked about this, I tried changing the `csurf` setup so that it relies on the session instead of a separate cookie, and it worked flawlessly (despite my hesitancy before). There is now a `csrfSecret` in the session data, and one less cookie (no `_csrfToken`). This seemed better to me, since the secret needed to verify the other one used in the header is only available in the database. I've made the change.

ssciolla:
Thank you all for the reviews, and for the approvals, @lsloan and @chrisrrowland. I don't think this is 100 percent the optimal setup; I've opened #167, #170, and #171 as follow-up issues, and I wouldn't be surprised if more things to fix popup. However, I do think it's mostly what we want, and I think it will help me (and us generally) keep moving forward to merge.

PR #150, opened 2021-07-09
Correct `User.canvasToken` type to be `CanvasToken | null` (#149)

### LINKED ISSUE #149: User.canvasToken has wrong type for when it doesn't exist
While working on #139, I discovered that I had incorrectly typed the `canvasToken` field on the `User` model. I thought that when the relation didn't exist, `User.canvasToken` would be `undefined`. However, it actually is `null`. This hasn't posed a problem yet because of a fluke in the `CanvasService.findToken` method where I was accidentally sometimes returning `null` when I thought it was truthy. See below on line 128:
https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/c3076b6f0bb2c469060638da59cd485b0cb860f0/ccm_web/server/src/canvas/canvas.service.ts#L122-L130

The type on `User.canvasToken` should instead be changed to `canvasToken!: CanvasToken | null`. This shouldn't require any migration changes. You can confirm that `null` is the expected type by logging any `User` instance before the OAuth flow is initiated.

Side note: I plan to get rid of `findToken` in other refactoring, but I feel should be fixed first.

ssciolla:
This PR fixes the type for `canvasToken` in `user.model.ts` so that it uses the correct `null` value for when it doesn't exist (previously `undefined`). I also simplified some code in `CanvasService.findToken` to return the token (`null` or not) directly. See the issue for more info on the bug. The PR aims to resolve #149.

### DIFF
--- a/ccm_web/server/src/canvas/canvas.service.ts
+++ b/ccm_web/server/src/canvas/canvas.service.ts
@@ -125,8 +125,7 @@ export class CanvasService {
       logger.error(`User ${userLoginId} is not in the database.`)
       throw new UserNotFoundError(userLoginId)
     }
-    const token = user.canvasToken === undefined ? null : user.canvasToken
-    return token
+    return user.canvasToken
   }
 
   async refreshToken (token: CanvasToken): Promise<CanvasToken> {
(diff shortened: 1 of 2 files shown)


pushyamig (approved):

lsloan (approved):

ssciolla:
Thanks for the reviews, folks!

Issue #88, opened 2021-06-01
Integrate the FE with BE for create sections


pushyamig:
Sections creation (~60) takes about 2 min even if all failed and retry it takes about ~5sec. So user gets an immediate response about the status of the request.

1. Once section creating is finished, the BE is sending what sections failed and what is the reasons for failure. The FE needs to display the section/reason for failure in a tabular form. This was something discussed with Janel/Melinda
2. When GetSection call failed, the user needs to be presented with the reason for failure of that section that is coming from the BE. "Failed to get the exiting section, please try again later" He should not be taken to the creation section upload screen. 
3. For all success case we tell user that with a generic message that "All sections are successfully created". not need to present with all the list of section that was successful

Response object from Canvas
```
{
  "statusCode": 422,
  "errors": [
    {
      "failedInput": "sec1",
      "message": "Error(s) from Canvas:  Unprocessable Entity"
      "canvasStatusCode": 422
    }
    
  ]
}
```
4. NestJs does some basic validation to check if the string is not empty and should not exceed 250 in length. Below is the Response structure from NestJS error. Some thought needs to be given in `handleErrors` how to handle different error object or FE probably could also do the validation for empty and length > 250. Doing both would be redundant. 
Response object from NestJs
```
{
  "statusCode": 400,
  "message": [
    "each value in sections must be shorter than or equal to 250 characters",
    "each value in sections should not be empty"
  ],
  "error": "Bad Request"
}
```

Issue #17, opened 2021-02-23
Feature: Bulk Section creation


pushyamig:
This is Admin/Subaccount/Service center only Feature. Note: Service Center people have Support Consultant role which is a level of Root account privilege. Sub account admins have highest privileges in a sub account and or not allowed access to other sub accounts in canvas

For workflow guideline follow [Wireframes](https://github.com/tl-its-umich-edu/UX-Work/blob/master/ccm/wireframes/add_section_feature/add_section_sc.pdf)

**Description:**
1. This process needs to take a simple CSV, with a header. SIS id creation is not essential for this. For consistence user experience we will stick with CSV file format when importing the file. 
3. The CSV file should have one column saying `section_name`. This is to enforce the users across with consistent naming convention. This only matter when showing a sample CSV file(shows as pop up in wireframes) should look like.
4. CSV validation involves 
   * Will be going with a simple file parsing instead of using CSV library( which is an over kill for this case) and probably don't validate the header name. Simply look at the data from row 2. 
   * making sure no 2 sections should not be same in the `.csv` file, 
   * section names should not be same with exiting section names in the course.
   * If there are exiting section name in the CSV file that are in the current Course, FE should list out what are those sections and show then a link to the Canvas course section page as an FYI. Here is the link to course section (Eg., https://canvas-test.it.umich.edu/courses/403334/settings#tab-sections)
3. This process can give an instantaneous response to the users about success/failure. **Note**: while implementing this needs to be  tested if that is the case otherwise we might need to run this as background process if that is the case an email should sent to the user updating the status.
4. The Max section names given in a csv file will be ~60

**Api Call**:
https://umich.test.instructure.com:443/api/v1/courses/232623/sections 
1. No need to create SIS id for sections and courses during this process since that will eliminate the self enrolling option.

**Note**: when creating sections in canvas via API, you can create  a section with same section names but both of the section id will be different. In order to save the user from creating same sections we do extra validation in during the CSV upload process

Here is the detailed Description on [API](https://docs.google.com/spreadsheets/d/1pm5y9FX0zrDX7Zy3mOyDLxmA-iKoWfmlxvbtNkWg6Zw/edit#gid=578161201) Doc

pushyamig:
This is truly feature complete now. All issues resolved

pushyamig:
[...]

PR #154, opened 2021-07-12
#152, #88 Create sections backend endpoint

### LINKED ISSUE #152: Create sections an BE endpoint

pushyamig:
Fixes #152 as part Epic #17

This PR is creating an endpoint for creating sections. This Endpoint should be able to handle single section and Multiple sections. The Endpoint creates multiple sections concurrently

**Add the new Scope in the API Developer Key** `url:POST|/api/v1/courses/:course_id/sections`

For creating 60 section it takes about ~2 sec and with retry as well about ~5 sec. I did not see any rating limiting error while working on the issue. So for the max section we could create (60) with CCM shouldn't hit the rate limiting phenomenon.

I have used Promise.all() approach for triggering concurrent creation of sections.

Some of the articles that I explored
coreycleary.me/awaiting-multiple-requests-to-finish-using-promise-all
davidroyer.me/blog/optimized-async-await
https://github.com/sindresorhus/got#retry

### DIFF
--- a/ccm_web/server/src/api/api.create.section.handler.ts
+++ b/ccm_web/server/src/api/api.create.section.handler.ts
@@ -1,52 +1,64 @@
 import CanvasRequestor from '@kth/canvas-api'
 import baseLogger from '../logger'
-import { CreateSectionResponseData, CreateSectionReturnResponse, CanvasSectionBase, CreateSectionsResponseObject, APIErrorData, handleAPIError } from './api.interfaces'
+import { CreateSectionTempDataStore, CanvasSectionBase, CreateSectionsAPIErrorData, APIErrorData, isAPICreateSectionErrorData } from './api.interfaces'
+import { handleAPIError } from './api.utils'
 
 const logger = baseLogger.child({ filePath: __filename })
 
 export class CreateSectionApiHandler {
-  constructor (private readonly requestor: CanvasRequestor,
-    private readonly sections: string[],
-    private readonly courseId: number) {}
+  requestor: CanvasRequestor
+  sections: string[]
+  courseId: number
 
-  async apiCreateSection (sectionName: string): Promise<CreateSectionsResponseObject> {
+  constructor (requestor: CanvasRequestor, sections: string[], courseId: number) {
+    this.requestor = requestor
+    this.sections = sections
+    this.courseId = courseId
+  }
+
+  async apiCreateSection (sectionName: string): Promise<CanvasSectionBase | CreateSectionsAPIErrorData> {
     try {
-      const endpoint = `courses/${this.courseId}/sections`
+      const fake = `courses/${this.courseId}/sections/dig/`
+      const real = `courses/${this.courseId}/sections`
+      const endpoint = Math.random() < 0.5 ? fake : real
       const method = 'POST'
       const requestBody = { course_section: { name: sectionName } }
-      logger.debug(`Sending request to Canvas - Endpoint: ${endpoint}; Method: ${method}; Body: ${JSON.stringify(requestBody)}`)
+      logger.debug(`Sendings request to Canvas - Endpoint: ${endpoint}; Method: ${method}; Body: ${JSON.stringify(requestBody)}`)
       const response = await this.requestor.requestUrl<CanvasSectionBase>(endpoint, method, requestBody)
-      return { statusCode: response.statusCode, message: response.statusMessage as string, sectionName: response.body.name }
+      const { id, name } = response.body
+      return { id, name }
     } catch (error) {
       const errResponse: APIErrorData = handleAPIError(error)
-      logger.error(`Request to create section ${sectionName} failed with StatusCode: ${errResponse.statusCode} and StatusMessage: ${errResponse.message} `)
-      return { statusCode: errResponse.statusCode, message: errResponse.message, sectionName: sectionName }
+      return { statusCode: errResponse.statusCode, errors: [{ message: errResponse.message, sectionName: sectionName }] }
     }
   }
 
-  makeReturnResponseCreateSections (sectionsReturnRes: CreateSectionsResponseObject[]): CreateSectionReturnResponse {
-    const sectionsDataStore: CreateSectionResponseData = { createdSections: 0, givenSections: this.sections.length, statusCode: [], error: {} }
+  makeReturnResponseCreateSections (sectionsReturnRes: Array<CreateSectionsAPIErrorData | CanvasSectionBase>): CanvasSectionBase[] | CreateSectionsAPIErrorData {
+    const sectionsDataStore: CreateSectionTempDataStore = { createdSections: 0, givenSections: this.sections.length, allSuccess: [], statusCode: [], errors: [] }
     for (const section of sectionsReturnRes) {
-      const { statusCode, message, sectionName } = section
-      if (statusCode === 200 || statusCode === 201) {
-        sectionsDataStore.createdSections++
-      } else {
-        sectionsDataStore.error[sectionName] = message
+      if (isAPICreateSectionErrorData(section)) {
+        const { statusCode, errors } = section
+        sectionsDataStore.errors.push(errors[0])
         sectionsDataStore.statusCode.push(statusCode)
+      } else {
+        sectionsDataStore.createdSections++
+        sectionsDataStore.allSuccess.push(section)
       }
     }
 
     if (sectionsDataStore.createdSections === sectionsDataStore.givenSections) {
-      return { statusCode: 201, message: { section: { success: true } } }
+      return sectionsDataStore.allSuccess
     } else {
+      const statusCodes = [...new Set(sectionsDataStore.statusCode)]
+      const statusCode = statusCodes.length > 1 ? 400 : statusCodes[0]
       return {
-        statusCode: Math.max(...[...new Set(sectionsDataStore.statusCode)]),
-        message: { section: { success: false, error: sectionsDataStore.error } }
+        statusCode: statusCode,
+        errors: sectionsDataStore.errors
       }
     }
   }
 
-  async createSectionBase (): Promise<CreateSectionReturnResponse> {
+  async createSectionBase (): Promise<CanvasSectionBase[] | CreateSectionsAPIErrorData> {
     const start = process.hrtime()
     const apiPromises = this.sections.map(async (section) => await this.apiCreateSection(section))
     const sectionsOrErrorDataObjs = await Promise.all(apiPromises)
--- a/ccm_web/server/src/api/api.create.section.handler.ts
+++ b/ccm_web/server/src/api/api.create.section.handler.ts
@@ -23,7 +23,7 @@ export class CreateSectionApiHandler {
       const endpoint = `courses/${this.courseId}/sections`
       const method = 'POST'
       const requestBody = { course_section: { name: sectionName } }
-      logger.debug(`Sendings request to Canvas - Endpoint: ${endpoint}; Method: ${method}; Body: ${JSON.stringify(requestBody)}`)
+      logger.debug(`Sending request to Canvas - Endpoint: ${endpoint}; Method: ${method}; Body: ${JSON.stringify(requestBody)}`)
       const response = await this.requestor.requestUrl<CanvasSectionBase>(endpoint, method, requestBody)
       const { id, name } = response.body
       return { id, name }
--- a/ccm_web/server/src/api/api.create.section.handler.ts
+++ b/ccm_web/server/src/api/api.create.section.handler.ts
@@ -1,6 +1,6 @@
 import CanvasRequestor from '@kth/canvas-api'
 
-import { CanvasSectionBase, CreateSectionsAPIErrorData, APIErrorData, isAPICreateSectionErrorData } from './api.interfaces'
+import { APIErrorData, CanvasCourseSection, isAPIErrorData } from './api.interfaces'
 import { handleAPIError } from './api.utils'
 
 import baseLogger from '../logger'
@@ -18,25 +18,25 @@ export class CreateSectionApiHandler {
     this.courseId = courseId
   }
 
-  async createSection (sectionName: string): Promise<CanvasSectionBase | CreateSectionsAPIErrorData> {
+  async createSection (sectionName: string): Promise<CanvasCourseSection | APIErrorData> {
     try {
       const endpoint = `courses/${this.courseId}/sections`
       const method = 'POST'
       const requestBody = { course_section: { name: sectionName } }
       logger.debug(`Sending request to Canvas - Endpoint: ${endpoint}; Method: ${method}; Body: ${JSON.stringify(requestBody)}`)
-      const response = await this.requestor.requestUrl<CanvasSectionBase>(endpoint, method, requestBody)
+      const response = await this.requestor.requestUrl<CanvasCourseSection>(endpoint, method, requestBody)
       const { id, name } = response.body
       return { id, name }
     } catch (error) {
-      const errResponse: APIErrorData = handleAPIError(error)
-      return { statusCode: errResponse.statusCode, errors: [{ message: errResponse.message, failedInput: sectionName }] }
+      const errResponse = handleAPIError(error, sectionName)
+      return { statusCode: errResponse.canvasStatusCode, errors: [errResponse] }
     }
   }
 
-  makeReturnResponseCreateSections (sectionsReturnRes: Array<CreateSectionsAPIErrorData | CanvasSectionBase>): CanvasSectionBase[] | CreateSectionsAPIErrorData {
+  makeReturnResponseCreateSections (sectionsReturnRes: Array<APIErrorData | CanvasCourseSection>): CanvasCourseSection[] | APIErrorData {
     const successes = []; const statusCodes = []; const errorsList = []
     for (const section of sectionsReturnRes) {
-      if (isAPICreateSectionErrorData(section)) {
+      if (isAPIErrorData(section)) {
         const { statusCode, errors } = section
         errorsList.push(...errors)
         statusCodes.push(statusCode)
@@ -50,13 +50,13 @@ export class CreateSectionApiHandler {
     } else {
       const uniqueStatusCodes = [...new Set(statusCodes)]
       return {
-        statusCode: uniqueStatusCodes.length > 1 ? 400 : uniqueStatusCodes[0],
+        statusCode: uniqueStatusCodes.length > 1 ? 502 : uniqueStatusCodes[0],
         errors: errorsList
       }
     }
   }
 
-  async createSections (): Promise<CanvasSectionBase[] | CreateSectionsAPIErrorData> {
+  async createSections (): Promise<CanvasCourseSection[] | APIErrorData> {
     const start = process.hrtime()
     const apiPromises = this.sections.map(async (section) => await this.createSection(section))
     const sectionsOrErrorDataObjs = await Promise.all(apiPromises)
--- a/ccm_web/server/src/api/api.create.section.handler.ts
+++ b/ccm_web/server/src/api/api.create.section.handler.ts
@@ -1,7 +1,9 @@
 import CanvasRequestor from '@kth/canvas-api'
-import baseLogger from '../logger'
+
 import { CreateSectionTempDataStore, CanvasSectionBase, CreateSectionsAPIErrorData, APIErrorData, isAPICreateSectionErrorData } from './api.interfaces'
 import { handleAPIError } from './api.utils'
+
+import baseLogger from '../logger'
 
 const logger = baseLogger.child({ filePath: __filename })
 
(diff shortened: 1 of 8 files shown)


ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+  }
+
+  async createSectionBase (): Promise<CreateSectionReturnResponse> {
+    const start = process.hrtime()
+    const apiPromises = this.sections.map(async (section) => await this.apiCreateSection(section))
+    const sectionsOrErrorDataObjs = await Promise.all(apiPromises)
+    const stop = process.hrtime(start)
+    logger.info(`Time Taken to ${this.sections.length} create sections : ${(stop[0] * 1e9 + stop[1]) / 1e9} seconds`)
```
ssciolla, line 54:
```suggestion
    logger.debug(`Time taken to create ${this.sections.length} sections : ${(stop[0] * 1e9 + stop[1]) / 1e9} seconds`)
```

ccm_web/server/src/api/api.interfaces.ts
```diff
@@ -13,7 +19,34 @@ export interface APIErrorData {
...
+  sectionName: string
+}
+
+export interface CanvasSectionBase {
+  name: string
+}
+
+export interface CreateSectionReturnResponse {
+  statusCode: number
+  message: Record<any, unknown>
```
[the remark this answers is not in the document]
ssciolla: After having looked at the handler class more closely, I think we can be more precise with this. I'll talk to you about it off GitHub.

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,57 @@
...
+    private readonly courseId: number) {}
+
+  async apiCreateSection (sectionName: string): Promise<CreateSectionsResponseObject> {
+    try {
+      const endpoint = `courses/${this.courseId}/sections`
+      const method = 'POST'
+      const requestBody = { course_section: { name: sectionName } }
+      logger.debug(`Sending request to Canvas - Endpoint: ${endpoint}; Method: ${method}; Body: ${JSON.stringify(requestBody)}`)
+      const response = await this.requestor.requestUrl<CanvasSectionBase>(endpoint, method, requestBody)
+      return { statusCode: response.statusCode, message: response.statusMessage as string, sectionName: response.body.name }
```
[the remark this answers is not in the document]
ssciolla: Then, in your `makeReturn...` function, you could use a type guard to determine whether it's an error object or not.

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+
+    if (sectionsDataStore.createdSections === sectionsDataStore.givenSections) {
+      return { statusCode: 201, message: { section: { success: true } } }
+    } else {
+      return {
+        statusCode: Math.max(...[...new Set(sectionsDataStore.statusCode)]),
+        message: { section: { success: false, error: sectionsDataStore.error } }
+      }
```
ssciolla, line 45: I'm thinking we should come up with a better structure or way of doing this. It doesn't make sense to me to use the max status code; either we should keep track of each irregular status code and tie them to particular errors, or we should provide a more generic one explicitly.

Storing nested data under `message` feels clunky. If we want to give the client more data, then we might want to consider providing an alternative error data structure for these bulk processes.

ssciolla (commented):
Hi @pushyamig, I haven't tried running this yet, but it represents some a lot of good thinking and work to get it all wired up. That said, I have some thoughts and concerns about the error handling and data serialization that I think we should discuss a bit. I will reach out to you. I've also included some miscellaneous type, organization, and style comments/suggestions that occurred to me.

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -181,7 +182,8 @@ export class CanvasService {
       logger.debug('Token for user has expired; refreshing token...')
       token = await this.refreshToken(token)
     }
-    const requestor = new CanvasRequestor(this.url + endpoint, token.accessToken)
+    const options: GotOptions = { retry: { limit: 2, methods: ['POST', 'GET', 'PUT', 'HEAD', 'DELETE', 'OPTIONS', 'TRACE'] } }
```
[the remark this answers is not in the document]
pushyamig: Do you mean out side of class `CanvasService` but with in the scope of `canvas.service.ts`?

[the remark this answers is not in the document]
ssciolla: Yeah, that seems okay (outside class definition, still in `canvas.service.ts`). I don't think we'll need this in other files.

ccm_web/server/src/api/api.interfaces.ts
```diff
@@ -13,7 +13,29 @@ export interface APIErrorData {
   statusCode: number
   message: string
 }
+export interface CanvasSectionBase {
+  id: number
+  name: string
+}
+export interface createSectionError {
+  failedInput: string
```
pushyamig, line 21: @ssciolla after you are convinced that this structure is good, I will change it with other API endpoints.

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -18,6 +19,7 @@ import baseLogger from '../logger'
 const logger = baseLogger.child({ filePath: __filename })
 
 type SupportedAPIEndpoint = '/api/v1/' | '/api/graphql/'
+const options: GotOptions = { retry: { limit: 2, methods: ['POST', 'GET', 'PUT', 'DELETE'] } }
```
ssciolla, line 22: Nit-pick: `requestorOptions`?

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+      return { statusCode: errResponse.statusCode, errors: [{ message: errResponse.message, failedInput: sectionName }] }
+    }
+  }
+
+  makeReturnResponseCreateSections (sectionsReturnRes: Array<CreateSectionsAPIErrorData | CanvasSectionBase>): CanvasSectionBase[] | CreateSectionsAPIErrorData {
+    const sectionsDataStore: CreateSectionTempDataStore = { allSuccess: [], statusCode: [], errors: [] }
+    for (const section of sectionsReturnRes) {
+      if (isAPICreateSectionErrorData(section)) {
+        const { statusCode, errors } = section
+        sectionsDataStore.errors.push(errors[0])
```
ssciolla, line 39: Do you see any possible issues with using `sectionDataStore.errors.push(...errors)` instead? I guess I feel minorly uncomfortable with the indexing (emphasis minorly).

ssciolla (commented):
Thanks, @pushyamig, for working with me on this. I'm liking it even more. I left a few more comments about minor things.

I think you can go ahead and change the structure of the `APIErrorData` used by the course endpoints to use `errors: [{ message: '', failedInput: '' }]`.

I will get back to you today about the body versus `statusMessage` thing.

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  constructor (requestor: CanvasRequestor, sections: string[], courseId: number) {
+    this.requestor = requestor
+    this.sections = sections
+    this.courseId = courseId
+  }
+
+  async apiCreateSection (sectionName: string): Promise<CanvasSectionBase | CreateSectionsAPIErrorData> {
+    try {
+      const endpoint = `courses/${this.courseId}/sections/ding`
```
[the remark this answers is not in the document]
pushyamig: Sorry, Removed it

ccm_web/server/src/api/api.create.section.handler.ts
pushyamig: No, this works nicely.

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+      } else {
+        sectionsDataStore.allSuccess.push(section)
+      }
+    }
+
+    if (sectionsDataStore.allSuccess.length === this.sections.length) {
+      return sectionsDataStore.allSuccess
+    } else {
+      const statusCodes = [...new Set(sectionsDataStore.statusCode)]
+      const statusCode = [...new Set(sectionsDataStore.statusCode)].length > 1 ? 400 : statusCodes[0]
```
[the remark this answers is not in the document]
pushyamig: Yeah...

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+      const { id, name } = response.body
+      return { id, name }
+    } catch (error) {
+      const errResponse: APIErrorData = handleAPIError(error)
+      return { statusCode: errResponse.statusCode, errors: [{ message: errResponse.message, failedInput: sectionName }] }
+    }
+  }
+
+  makeReturnResponseCreateSections (sectionsReturnRes: Array<CreateSectionsAPIErrorData | CanvasSectionBase>): CanvasSectionBase[] | CreateSectionsAPIErrorData {
+    const sectionsDataStore: CreateSectionTempDataStore = { allSuccess: [], statusCode: [], errors: [] }
```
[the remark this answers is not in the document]
pushyamig: Ok, I removed the data store variable and I ended up declaring something like this in single line. which ever reason linter don't complain and happy with it
	`const successes = []; const statusCodes = []; const errorsList = []`

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,67 @@
...
+        successes.push(section)
+      }
+    }
+
+    if (successes.length === this.sections.length) {
+      return successes
+    } else {
+      const uniqueStatusCodes = [...new Set(statusCodes)]
+      return {
+        statusCode: uniqueStatusCodes.length > 1 ? 400 : uniqueStatusCodes[0],
```
lsloan, line 53: Why a status code of 400 here?  That means "bad request".  What if the problem is not in the request to the CCM API, but is a problem with contacting the Canvas API?  In that case, giving a 400 here will give the impression to the frontend that its request was made incorrectly.  It hides the fact that something may have gone wrong in Canvas.

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+  }
+
+  async createSectionBase (): Promise<CanvasSectionBase[] | CreateSectionsAPIErrorData> {
+    const start = process.hrtime()
+    const apiPromises = this.sections.map(async (section) => await this.apiCreateSection(section))
+    const sectionsOrErrorDataObjs = await Promise.all(apiPromises)
+    const stop = process.hrtime(start)
+    logger.debug(`Time taken to create ${this.sections.length} sections: ${(stop[0] * 1e9 + stop[1]) / 1e9} seconds`)
```
lsloan, line 64: It's probably unnecessary to compute the exact time spent on the process.  It's probably good enough to give begin/end debug log messages.  We can figure out from the timestamps in the log how long it took.

ccm_web/server/src/api/api.utils.ts
```diff
@@ -0,0 +1,18 @@
+import { HTTPError } from 'got'
+import { APIErrorData } from './api.interfaces'
+
+import baseLogger from '../logger'
+
+const logger = baseLogger.child({ filePath: __filename })
+
+export function handleAPIError (error: unknown): APIErrorData {
```
lsloan, line 8: Why has this been moved to a new module?

ccm_web/server/src/api/api.interfaces.ts
```diff
@@ -13,7 +13,25 @@ export interface APIErrorData {
...
+export interface CreateSectionsAPIErrorData {
+  statusCode: number
+  errors: CreateSectionError[]
+}
 
 export function isAPIErrorData (value: unknown): value is APIErrorData {
   return hasKeys(value, ['statusCode', 'message'])
 }
+
+export function isAPICreateSectionErrorData (value: unknown): value is CreateSectionsAPIErrorData {
```
lsloan, line 35: This seems useful for more than just creating sections.  It should have a more general name and work with a more general error data type.

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,67 @@
...
+    } else {
+      const uniqueStatusCodes = [...new Set(statusCodes)]
+      return {
+        statusCode: uniqueStatusCodes.length > 1 ? 400 : uniqueStatusCodes[0],
+        errors: errorsList
+      }
+    }
+  }
+
+  async createSectionBase (): Promise<CanvasSectionBase[] | CreateSectionsAPIErrorData> {
```
lsloan, line 59:
```suggestion
  async createSections (): Promise<CanvasSectionBase[] | CreateSectionsAPIErrorData> {
```

ccm_web/server/src/api/api.create.section.handler.ts
```diff
@@ -0,0 +1,67 @@
...
+  sections: string[]
+  courseId: number
+
+  constructor (requestor: CanvasRequestor, sections: string[], courseId: number) {
+    this.requestor = requestor
+    this.sections = sections
+    this.courseId = courseId
+  }
+
+  async apiCreateSection (sectionName: string): Promise<CanvasSectionBase | CreateSectionsAPIErrorData> {
```
lsloan, line 21:
```suggestion
  async createSection (sectionName: string): Promise<CanvasSectionBase | CreateSectionsAPIErrorData> {
```

lsloan (changes requested):

ccm_web/server/src/api/api.interfaces.ts
ssciolla: @pushyamig is in the process of changing that, we've discussed it separately (may be alluded to in other comments).

ccm_web/server/src/api/api.utils.ts
ssciolla: @pushyamig needed access to it in her handler class.

ccm_web/server/src/api/api.create.section.handler.ts
ssciolla: What status code would you use then? This might need some additional discussion and testing.

ccm_web/server/src/api/api.utils.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (error instanceof HTTPError) {
+    const { statusCode, statusMessage } = error.response
+    logger.error(`Received unusual status code ${String(statusCode)}`)
+    logger.error(`Response body: ${JSON.stringify(statusMessage)}`)
+    return { statusCode, message: `Error(s) from Canvas:  ${statusMessage as string}` }
```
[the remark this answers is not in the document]
ssciolla: Okay, so here's where I'm at on the `body` question currently.

When there are errors reported here, these will be errors that have fallen through the cracks, that we may want to handle explicitly in future versions of the code. Because of that, I don't know if it's necessary to provide this information to the client. Errors that client will need to know lots of detail about should be caught by the validation code (see the `dto`). Instead, we can log the body so that someone responding to production incidents can find and use in creating an issue.

Thus, so long as we do log the body somewhere, I am okay with deleting `CanvasService.parseErrorBody` (and the related type guards) and relying on the `statusMessage` as what we deliver to the client. Doing this also has an added benefit of us not having to handle all possible shapes of Canvas error responses (of which I've seen a few, and which could conceivably change).

I'm willing to discuss further as needed, @pushyamig, thanks!

pushyamig:
>Thanks, @pushyamig, for working with me on this. I'm liking it even more. I left a few more comments about minor things.
I think you can go ahead and change the structure of the APIErrorData used by the course endpoints to use errors: [{ message: '', failedInput: '' }].


@ssciolla, I have created a new [PR](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/157) addressing this, Please review it once we merge that I will apply it here

ccm_web/server/src/api/api.interfaces.ts
pushyamig: Yeah, I open a different PR https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/157/files, please review that as well so can merge that first and bring in that changes here

ccm_web/server/src/api/api.utils.ts
pushyamig: Yeah... I did the refactoring this function could be used by all API endpoints

ccm_web/server/src/api/api.interfaces.ts
```diff
@@ -13,7 +13,25 @@ export interface APIErrorData {
   statusCode: number
   message: string
 }
+export interface CanvasSectionBase {
```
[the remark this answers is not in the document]
pushyamig: OK, I don't mind changing it to `CanvasCourseSection`

ccm_web/server/src/api/api.create.section.handler.ts
pushyamig: Yeah.... If there are 6 section needs to be created, probably we should expect that failure can occur for different reason, although we don't know how often we would hit this scenario. We could have a mix of [200, 400, 402, 503, 200, 200]

We can make it 500 making it as internal error, but we are logging all those status code in logs if we need to find more details.

We can discuss this in the meeting

ccm_web/server/src/api/api.create.section.handler.ts
pushyamig: I don't think it hurts to calculate, since I am not using any external logging library but using the Node built in library `process.hrtime`. I would like to keep it this way, But I am open for discussion if you have any concerns from that

ccm_web/server/src/api/api.create.section.handler.ts
pushyamig: Can we use a custom code that is available as per HTTP standards? That way we know that when user reports failure and  give us a status code ( if they are that savvy) we can figure out this is custom error code https://stackoverflow.com/a/62866782

Thoughts???

ccm_web/server/src/api/api.create.section.handler.ts
pushyamig: I renamed it

ccm_web/server/src/api/api.create.section.handler.ts
pushyamig: Renamed it

ccm_web/server/src/api/api.interfaces.ts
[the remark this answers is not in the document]
pushyamig: I renamed to `CanvasCourseSection`

ccm_web/server/src/api/api.create.section.handler.ts
pushyamig: After the discussion on the slack and in meeting, I changed it to 502 as per @lsloan suggestion `Note that the definition of 502 seems a good fit for some problems CCM's API will handle with Canvas' API: https://datatracker.ietf.org/doc/html/rfc7231#section-6.6.3`

pushyamig:
@chrisrrowland can you review/run locally this PR?

chrisrrowland:
> @chrisrrowland can you review/run locally this PR?

Yup, will hit it asap.  Trying to get my course setup again...

ccm_web/server/src/api/api.create.section.handler.ts
chrisrrowland: I agree.  Calculate the elapsed time all of the places!

chrisrrowland:
It would seem that canvas lets you create duplicate sections, is it ok that we let that happen?

pushyamig:
>It would seem that canvas lets you create duplicate sections, is it ok that we let that happen?

@chrisrrowland we control that using Frontend validation. We get the sections first and compare that with section uploaded to existing section and throw error to the user. You have that flow in place or to my knowledge a placeholder is set for that set
 The Swagger interface won't mimic that so you might see that behavior when testing

chrisrrowland:
> > It would seem that canvas lets you create duplicate sections, is it ok that we let that happen?
> 
> @chrisrrowland we control that using Frontend validation. We get the sections first and compare that with section uploaded to existing section and throw error to the user. You have that flow in place or to my knowledge a placeholder is set for that set
> The Swagger interface won't mimic that so you might see that behavior when testing

Ok I was asking if the backend should prevent it.

pushyamig:
>Ok I was asking if the backend should prevent it.

If FE is validating then that should be good enough

chrisrrowland (approved):
I don't have any issues to bring up.

Might have some changes to consider once I wire it up to the front end but looks good.

ccm_web/server/src/api/api.create.section.handler.ts
lsloan: @chrisrrowland…
Calculate ALL the elapsed times!

PR #165, opened 2021-07-27
bulk enrollment (iss. #108)

### LINKED ISSUE #108: create API function for bulk-enrolling users
## Requirements
* Accept POST requests to `sections/:sectionId/enroll`
* The section identified by `sectionId` must exist in Canvas and the user must have proper access to it.  This will not be checked.  If there are any errors from Canvas (e.g., section doesn't exist or the user doesn't have the proper access), they will be returned in the response from this API entry.
* List of users to be enrolled, along with their role, must be submitted as JSON array. The CSV of users, submitted to the UI by the application user, will be reformatted to JSON by the UI code.
  * Request body example:
    ```json  
    {"users":[
      {"loginId": "name1", "role": "StudentEnrollment"},
      {"loginId": "name2", "role": "TeacherEnrollment"}
    ]}
    ```
  * Values for `role` must be one of the following valid Canvas roles:
    * `StudentEnrollment`
    * `TeacherEnrollment`
    * `TaEnrollment`
    * `ObserverEnrollment`
    * `DesignerEnrollment`
  * Invalid roles will be rejected by the Canvas API.  Canvas errors will be relayed back to the CCM API user.
  * Values for `loginId` must be a U-M user uniqname or a U-M Friend account email address.  Use these rules, in this order:
    1. Email addresses ending with `@umich.edu` (this string should be configurable) will have the `@` and everything following it removed.
    2. During processing, email addresses will be converted to Canvas login ID format by replacing the `@` with `+`.
* Limit the number of users to be enrolled to 400 per request.  See Slack conversation (https://app.slack.com/client/T69CPPGNP/G0130F688N6/thread/G0130F688N6-1627054990.263100).  The limit should be configurable.
* Notifications should not be sent to users when they are enrolled in a section.  The request to the Canvas API must include `"notify": false`.
* Users should not be invited to the course, they should be added directly to the course by setting the enrollment state to `"active"`.
* Example request:  
  ```sh
  curl 'https://canvas-test.it.umich.edu/api/v1/sections/543762/enrollments' \
    -H 'authorization: Bearer 1770~u6…' \
    -H 'content-type: application/json' \
    --data-raw '{"enrollment":{"user_id":"sis_login_id:bjensen","type":"StudentEnrollment","enrollment_state":"active","notify":false}}'
  ```  
  See API reference: https://docs.google.com/spreadsheets/d/1pm5y9FX0zrDX7Zy3mOyDLxmA-iKoWfmlxvbtNkWg6Zw/edit#gid=578161201&range=C9
* Users are specified by uniqname or Friend account email address (`@` → `+` transformed), which Canvas refers to as the user `login_id`.  This function must prefix each uniqname/email with the string `sis_login_id:` to refer to users in the Canvas API enrollment request.  (See: https://canvas.instructure.com/doc/api/file.object_ids.html)
* Validations
	* Each section must exist.  Handled by the Canvas API enrollment request.
	* Each user must exist.  Handled by the Canvas API enrollment request.

lsloan:
Resolves #108.

Using the Got retry feature, this implementation is often (usually?), but not always, successful in enrolling a full set of 400 students.  It's ignoring the Canvas API throttling and essentially using brute force.  From team discussions, this is the technique that I was directed to pursue.  It may be important to revisit this to use request throttling before merging this PR.  I don't know, that's up to the reviewers and testers to determine.  I've received conflicting directions on this matter.

**_There are some complaints from eslint about using variables that may possibly be undefined._**  I've tried to correct them, but I've not found the right TypeScript syntax.  **_Maybe the reviewers can make a suggestion._**  Keep in mind that these complaints are about logging statements, so an appropriate suggestion may be to just remove the statements.

---

# TODO

* [x] Fix list of Got default errors for retry feature.  Include `403`. (see comment: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/165/files/c8171f0952361aa058e91de311081435544516d9#r702155573)
* [x] Eslint appeasement.  (see comment: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/165/files/c8171f0952361aa058e91de311081435544516d9#r703611833)
* [x] Use enrollment type enum. (see comment: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/165/files/c8171f0952361aa058e91de311081435544516d9#r702151039)


pushyamig:
I tested the work being done locally, I was able to enroll users to a course as me and as a subaccount admin. I did not test the 400 users case and @lsloan seem already testing this case. 

@chrisrrowland and @ssciolla can you review this PR since this is opened for couple of days now?

ccm_web/server/src/api/api.enroll.section.users.handler.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    if (successes.length === this.users.length) {
+      return successes
+    } else {
+      return {
+        statusCode: statusCodes.size > 1 ? 502 : [...statusCodes][0],
+        errors: failures
+      }
+    }
+  }
```
ssciolla, line 49: I think we could make this a utility function in `api.utils.ts`, since it's basically doing the same thing as what is being done in `api.create.section.handler.ts `. To do that, we probably need to use a [generic](https://www.typescriptlang.org/docs/handbook/2/generics.html). I think the method signature would look something like this:

```
makeResponse<T extends CanvasEntity> (inputCount: number, result: Array<APIErrorData | T>): T[] | APIErrorData { ...}
```

where `CanvasEntity` is defined in `canvas.interfaces.ts` as

```
type CanvasEntity = CanvasEnrollment | CanvasSection | CanvasCourse ...
```

It might need some additional finessing, but seems like a common enough operation that we should extract it, provided we can get the typing to work (which we should be able to). 

Side note: I am thinking of making `api.utils` into another service, so we can inject other services into it if necessary (like `CanvasService`, if we need to modify tokens based on failures for example). In that case, this new function would just be a method.

ccm_web/server/src/api/api.service.ts
```diff
@@ -3,13 +3,15 @@ import { Injectable } from '@nestjs/common'
...
-import { CanvasCourse, CanvasCourseBase, CanvasCourseSection } from '../canvas/canvas.interfaces'
+import { CanvasCourse, CanvasCourseBase, CanvasCourseSection, CanvasEnrollment } from '../canvas/canvas.interfaces'
 import { APIErrorData, Globals } from './api.interfaces'
 import { CreateSectionApiHandler } from './api.create.section.handler'
 import { CanvasService } from '../canvas/canvas.service'
 import { User } from '../user/user.model'
 
 import baseLogger from '../logger'
+import { EnrollSectionUsersApiHandler } from './api.enroll.section.users.handler'
+import { SectionUserDto } from './dtos/api.section.users.dto'
```
ssciolla, line 14: Nit-pick: Group with other `api` imports?

ccm_web/server/src/api/dtos/api.section.users.dto.ts
```diff
@@ -0,0 +1,33 @@
+import { ArrayMaxSize, ArrayMinSize, IsArray, IsIn, IsNotEmpty, ValidateNested } from 'class-validator'
+import { ApiProperty } from '@nestjs/swagger'
+import { Type } from 'class-transformer'
+
+export class SectionUserDto {
```
ssciolla, line 5: Do you think it would it be clearer/more accurate to call this a `SectionEnrollment`? As I scan the PR, I see the variable shorthand often ends up as `user(s)`, which is somewhat confusing given the preexisting presence of `user`. And really, once we add an enrollment type for a context, we've moved past it being a simple Canvas user object. This would require some tweaking through, but might help readability.

ssciolla (commented):
Hi @lsloan, looks solid so far, though I haven't tested this. I know you're still working on the throttling, but I wanted to give you some initial feedback since you have had requested a review. Please re-request a review once you think it's ready, and I will do testing and scan again.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: Yes, I got some warnings from IntelliJ IDEA about duplicated code.  It'd make sense to generalize this and move it into the utils.

ccm_web/server/src/api/api.service.ts
lsloan: So, just move these lines up with lines 7–8?

ccm_web/server/src/api/dtos/api.section.users.dto.ts
lsloan: So, rather than the enrollment code referring to "user(s)", refer to "enrollment(s)"?  I think the problem with that is I was thinking of the input as a "user" because Canvas' API accepts a "user" object and returns an object called "enrollment".

I'm happy to make changes that clarify what the code is doing, but I think that change would make it less clear.

ccm_web/server/src/api/api.service.ts
lsloan: So, just move these lines up with lines 7–8?

ssciolla: Yes. I'm trying not to be too particular about imports, since it's low on the "it matters" scale, but here's how I generally approach ordering them.

1) Library imports, alphabetical by package name
2) NestJS module imports, current directory and subdirectories first, then directories on higher levels, alphabetical by file and directories.
3) Stuff outside of Nest modules (like `logger`, `config`, etc.), similar file and directory orders.

Hard to be consistent always, but grouping stuff from the same directory works out pretty well with Nest module imports.

ccm_web/server/src/api/dtos/api.section.users.dto.ts
ssciolla: I think we could probably debate this for a while. You are creating an enrollment, so you give it enrollment data (not just data about a user -- you really only provide a user identifier -- but also a role in that context, when that enrollment begins and ends, etc.). The endpoint describes the object provided in the body as an enrollment (https://canvas.instructure.com/doc/api/enrollments.html#method.enrollments_api.create).That's consistent with creating a course or a section: you give it some course and section data, it gives you the course or section it created back.

So, while I don't quite follow your logic, if it's going to make it less clear to you, you can keep it as is. It's not terribly confusing to me now, so it's not worth getting into a back-and-forth over. I'll leave it to your discretion.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async enrollUser (user: SectionUserDto): Promise<CanvasEnrollment | APIErrorData> {
+    try {
+      const endpoint = `sections/${this.sectionId}/enrollments`
+      const method = HttpMethod.Post
+      const body = {
+        enrollment: {
```
ssciolla, line 56: ... ahem ...

chrisrrowland:
Just got back from vacation, will review this asap

lsloan:
> Just got back from vacation, will review this asap

Thanks, @chrisrrowland.  Something I was wondering about and I think your opinion would be valuable…

If the enrollment results are a mixture of successes and failures, should this API entry return all results or just the failures?  It is currently based on the other API entry for creating sections, which when there are **_any_** failures, returns only the failures in the `errors` property.  When all is successful, this entry returns a list of enrollment-like objects.  It occurred to me as I'm debugging the rate limit exceeded error that those successful results aren't being returned.  Should I change that?

chrisrrowland:
> > Just got back from vacation, will review this asap
> 
> Thanks, @chrisrrowland. Something I was wondering about and I think your opinion would be valuable…
> 
> If the enrollment results are a mixture of successes and failures, should this API entry return all results or just the failures? It is currently based on the other API entry for creating sections, which when there are **_any_** failures, returns only the failures in the `errors` property. When all is successful, this entry returns a list of enrollment-like objects. It occurred to me as I'm debugging the rate limit exceeded error that those successful results aren't being returned. Should I change that?

There is no wireframe that demonstrates what a partial failure/success looks like, and my answer would probably depend on that.  If we only want to show errors that seems sufficient.  When we have a 100% success we don't even show the successful adds, though I think its reasonable that the  API returns them.  Maybe @melindakraft can chime in.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
ssciolla: I'm sorry, this was not the nicest way to draw attention to this. Just another data point in the `user` versus `enrollment` discussion I suppose.

lsloan:
> There is no wireframe that demonstrates what a partial failure/success looks like, and my answer would probably depend on that. If we only want to show errors that seems sufficient. When we have a 100% success we don't even show the successful adds, though I think its reasonable that the API returns them. Maybe @melindakraft can chime in.

OK.  I see @ssciolla has given a thumbs-up to my question.  Since I was already leaning in that direction to return successes and failures together, I'll go ahead with that.

I'm trying to decide what the structure of the response would look like.  All successful currently looks like:

```json
[{enrollmentObject},{enrollmentObject},…]
```

Any errors looks like:

```json
{"errors": [{errorObject},{errorObject},…]}
```

Could a combined response look like:

```json
{"enrollments":[{enrollmentObject},{enrollmentObject},…],"errors": [{errorObject},{errorObject},…]}
```

chrisrrowland:
>{"enrollments":[{enrollmentObject},{enrollmentObject},…],"errors": [{errorObject},{errorObject},…]}

seems reasonable

ccm_web/server/src/api/api.enroll.section.users.handler.ts
chrisrrowland: It might make more sense to make this a base class for something that implements a partial success response.

If that's possible with generics, anyways.

-

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: My apologies.  I was not intentionally ignoring the discussion about "user" vs. "enrollment".  I'm just intensely focused on handling the Canvas API's "rate limit exceeded" error.  That seems to be my most important task at the moment.

Perhaps I misspoke about the input values required by Canvas API at the time I wrote my comment.  Or maybe I think about it from a different perspective than you do.  I can rename the symbols to whatever the group consensus is.  I guess the naming convention doesn't really matter to me in comparison to the error handling.

Looking at the section of code you cited, the lines that follow it are significant, too…

```js
      const body = {
        enrollment: {
          user_id: `sis_login_id:${user.loginId}`,
          type: user.type,
```

We're asking Canvas to make an enrollment based on the ID and type of a user.  We give Canvas info about a user (from a CSV file of users) and it makes an enrollment.

I mention this to respond to your point and to show what my thinking was when I named the symbols.  If I'm using inconsistent or confusing naming, I'm willing to change it.

Although not entirely applicable, our conversation reminds me of this quote:

> "Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be." ― Mark Twain

😉

ssciolla: Yeah, again, I'm leaving this up to your discretion. Variable/method naming is important, in my opinion, but less important than error handling (as you say) since the names are OK at the moment.

All that said, if you think about it, the `type` value you provide has nothing inherently to do with the `user` entity; it's describing the relation between the `user` and the `course`/`section`, i.e. the `enrollment`. The `type` value is not scoped to all of Canvas, it's scoped to a particular context, changing from section to section. So you're really providing more than a `user` to the API.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
```diff
@@ -0,0 +1,99 @@
...
+          // 'sis_login_id:' prefix per...
+          // https://canvas.instructure.com/doc/api/file.object_ids.html
+          user_id: `sis_login_id:${user.loginId}`,
```
chrisrrowland, line 59: Is this where we are doing something assuming the client sent uniqnames?  It's now unclear to me what the expectation is in that regard.  There is talk of the user including non-umich emails in the csv and *somewhere* there is expected to be an automatic translation to append +gmail.com.

ssciolla: I think the automatic translation was `@` to `+`, maybe that's what you meant. Looking at the Slack discussion, I think maybe there was confusion on this point. I think Canvas just expects the `@` sign to be a `+`, for whatever reason. You're right in that we can't know what's after the `@`.

chrisrrowland: so the user could put in a user id of 'rowc' (@umich.edu) or 'rowc@gmail.com' and the client should send 'rowc' and 'rowc+gmail.com' respectively, and either way the above code is correct?

ssciolla: I think so. A note (which you may know), I don't think we would ever want to send `@umich.edu`. I think either the loginId would be a simple string OR a non-`umich` email address. I think what's important is that it matches what is in Canvas, and UM users wouldn't have the email ending.

lsloan: I'm using the `sis_login_id` prefix here, as directed by @pushyamig.  It enables references to the user by their Canvas "login ID".  I don't know about all users, but in my case, my login ID is my uniqname, "lsloan", and nothing else, not my email address.  Likewise, the test users I've worked with have login IDs "canvasa", "canvasb", etc., like uniqnames, not email addresses.  When I do the test of 400 users, I use a file of STATS 250 student uniqnames, not email addresses.

I saw the conversation about email addresses ("xxx@umich.edu") and some other kind of ID ("xxx+umich.edu"), but that didn't seem to apply to this situation, so I don't know the details about those.

This code doesn't do any validation of the names used because there had been discussion to let Canvas return errors about missing/improper/inaccessible users and this API would return those to the client.  That also solves the validation problem with our organization using names of 3–8 lower-case alphabetic characters while other organizations may use different criteria.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: Yes, I can imagine it that way, too.  It's an enrollment type, because each valid type includes that word in it, like `StudentEnrollment`, etc.

So, I'll keep an open mind about this.  After the rate limit problem is under control, I promise to revisit this.  Maybe by then, I'll look at this with fresher eyes.

---

Mel: What do you plan to do today?
Carl: First, I'm going to buy some new glasses.
Mel: What will you do after that?
Carl: Then, I'll see.

lsloan:
I've hidden the comments about returning mixed success/error results.  I've opened issue #178 to handle that in the future.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: I'm working on implementing the generic.  I think we won't need an argument like `inputCount: number`.  I'll change the logic to check for 0 errors rather than a full list of successes.  I think that's equivalent for our purposes.

lsloan: @ssciolla & @chrisrrowland: I'm generalizing `makeResponse` in `api.utils` for use by class `EnrollSectionUsersApiHandler`.

Should I also update the section class, `CreateSectionApiHandler`, to use it, too?  Or should that be a different issue?

I'm leaning towards opening a new issue for it because:

* It doesn't fit logically in this PR for bulk enrollment.
* If it were in this PR, it would mean that the bulk enrollment **_AND_** section creation API entries should be tested before this PR could be merged.  That seems like a lot of testing.

chrisrrowland: Seems like it makes sense as a separate issue to me.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
```diff
@@ -49,14 +49,18 @@ export class EnrollSectionUsersApiHandler {
   }
 
   async enrollUser (user: SectionUserDto): Promise<CanvasEnrollment | APIErrorData> {
+    const enrollLoginId = user.loginId
+      .replace(/@umich\.edu$/i, '')
```
chrisrrowland, line 25: Does there need to be something to handle @med.umich.edu (and if there are similar prefixes like this?)
I don't know the rules for the med accounts, just aware that they exist.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: So, I've thought about this and I can see your point of view, but I also think the way I've named it is valid, too.

So, I'm going to leave it as is.  If someone wants to change it in the future… 🤷‍♂️

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: I've added code to handle email addresses as SIS login ID values.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: I really don't know.

Maybe @pushyamig or @melindakraft know.

lsloan: I suppose in the way we handle `@umich.edu` addresses, by dropping the domain, we could do the same thing for `@med.umich.edu`.  Because I think med. campus people are usually identified by their uniqname.

lsloan: Admittedly, I had suggested along the way (I think in the description of #108) that the domain to remove, like `@umich.edu` should be configurable.  But then I ignored that when I decided to make it a regex that's anchored to the end of the string and case is ignored.  I thought that maybe we don't want to throw regexes into the app. config.  (Maybe I'm wrong.)

Alternatively, I could reimplement that part to not use regex.  That's probably a little "safer".  It may also make it easier to support a _list_ of domains to be stripped from the login ID. 🤷‍♂️

lsloan: @chrisrrowland, thanks for reviewing again and raising this excellent question.  I'm sorry I didn't notice your new review before I requested a new one from you!

I've been busy with crunch time support, so I didn't realize until now that you're probably back with IQ full time, right?

chrisrrowland: > I suppose in the way we handle `@umich.edu` addresses, by dropping the domain, we could do the same thing for `@med.umich.edu`. Because I think med. campus people are usually identified by their uniqname.

I was curious if the uniqname part was distinct across all *umich.edu email addresses.
If so it seems reasonable to strip anything that ends .umich.edu down to just the uniqname.
I would guess it works this way and nobody can come in and create rowc@med.umich.edu

If not, i suppose it would just be treated as a normal email address, b/c how else would you distinguish...

I'm actually still on loan until the end of the month.. not rid of me yet

ssciolla (commented):
Sharing some comments. Only significant concern would be the status code configuration for retries. Using the `EnrollmentType` enum in `CanvasEnrollment` would also be nice.

I'm moving on to testing, I may not finish today. Looks good though.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
ssciolla: I guess I'm okay either way here, but my inclination would be to just use it in both places now, as part of testing this new function is to make sure it will work in both places. I don't think there's a very big risk in doing so, nor do I think it would be that difficult to test.

ccm_web/server/src/api/api.service.ts
```diff
@@ -91,4 +89,10 @@ export class APIService {
     const createSectionsApiHandler = new CreateSectionApiHandler(requestor, sections, course)
     return await createSectionsApiHandler.createSections()
   }
+
+  async enrollSectionUsers (user: User, sectionId: number, sectionUsers: SectionUserDto[]): Promise<CanvasEnrollment[] | APIErrorData> {
+    const requestor = await this.canvasService.createRequestorForUser(user, '/api/v1/')
+    const enrollmentHandler = new EnrollSectionUsersApiHandler(requestor, sectionUsers, sectionId)
+    return await enrollmentHandler.enrollUsers()
```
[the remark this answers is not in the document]
chrisrrowland: Took me a bit to understand what @ssciolla was suggesting here, but I would agree that taking the users array as a parameter to `enrollUsers` makes sense to me design wise.  
I don't know that it will ever make a difference in how it's used, but it would allow the `EnrollSectionUsersApiHandler`  object to be reused, at least in theory, so it seems cleaner.

ccm_web/server/src/api/api.utils.ts
```diff
@@ -28,3 +37,31 @@ export function parseErrorBody (body: unknown): string {
   }
   return body.errors.map(e => e.message).join(' ')
 }
+
+export function makeResponse<T extends CanvasEntity> (multipleResults: Array<APIErrorData | T>): T[] | APIErrorData {
```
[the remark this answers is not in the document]
chrisrrowland: I don't actually have an opinion, part 1 just made me think of this
Alt Text

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -18,7 +18,37 @@ import baseLogger from '../logger'
...
+        logger.debug(`beforeRetry [${retryCount}]: error.code: "${error?.code}"`)
+      }
+    ],
+    beforeError: [
+      error => {
+        logger.debug('beforeError')
+        return error
+      }
+    ]
+  }
```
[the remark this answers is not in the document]
chrisrrowland: Posted a possible PR  that would eliminate some of the linter complaints.  I only tested a single call to 
​/api​/course​/{id}​/sections, though, so you  might want to test more thoroughly if you choose to use it.

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -18,7 +18,37 @@ import baseLogger from '../logger'
 const logger = baseLogger.child({ filePath: __filename })
 
 type SupportedAPIEndpoint = '/api/v1/' | '/api/graphql/'
-const requestorOptions: GotOptions = { retry: { limit: 2, methods: ['POST', 'GET', 'PUT', 'DELETE'] } }
+const requestorOptions: GotOptions = {
+  retry: {
+    limit: 2,
+    methods: ['POST', 'GET', 'PUT', 'DELETE'],
+    statusCodes: [403]
```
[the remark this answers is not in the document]
lsloan: Yes, you're right.  I had meant to go back and add in all the defaults, but I forgot.  I will try to get the list of defaults from Got and add `403` to it.  If I can't do that, then I'll hard-code their default list and include `403`.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -46,6 +46,16 @@ export interface CanvasCourseSection {
   total_students?: number
 }
 
+export interface CanvasEnrollment {
+  id: number
+  course_id: number
+  course_section_id: number
+  user_id: number
+  type: string
```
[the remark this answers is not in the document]
lsloan: Probably.  I forgot about that.

ccm_web/server/src/api/api.utils.ts
```diff
@@ -1,19 +1,28 @@
...
+  APIErrorData, APIErrorPayload, isAPIErrorData
+} from './api.interfaces'
+import { CanvasEntity, isCanvasErrorBody } from '../canvas/canvas.interfaces'
 
 import baseLogger from '../logger'
 
 const logger = baseLogger.child({ filePath: __filename })
 
+export enum HttpMethod {
+  Get = 'GET',
```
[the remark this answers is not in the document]
lsloan: I've seen some code does, some doesn't.  I used to lose sleep over whether I should or shouldn't.  I try not to any more.  If you prefer capitals, we can use that.

ccm_web/server/src/api/api.service.ts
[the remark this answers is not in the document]
lsloan: I think I can support this idea, too.  I was following the pattern used for the create section code, which passed the input to the handler's constructor.  It didn't make sense to me why it was done that way, but seeing as it passed through code review, I thought that was the preferred technique.

ccm_web/server/src/api/api.utils.ts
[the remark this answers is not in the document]
ssciolla: Not losing sleep seems wise. Let's disregard for now (at least until it name collision/confusion becomes an issue).

ssciolla:
@lsloan, I'm guessing you are in the midst of making a couple other changes? I started doing a little testing, but I can do a bit more once you are done. If you DM me with hints on how to test high-volume tasks (lots of enrollments), I can do some of that as well. I think we should try to merge this soon -- if possible.

lsloan:
> @lsloan, I'm guessing you are in the midst of making a couple other changes? I started doing a little testing, but I can do a bit more once you are done. If you DM me with hints on how to test high-volume tasks (lots of enrollments), I can do some of that as well. I think we should try to merge this soon -- if possible.

@ssciolla, indeed, I'm currently making changes.

Sample data is available in a subfolder of the "New Canvas Course Manager" project folder in Drive.  See (https://drive.google.com/drive/u/0/folders/14krS9rmQydq64pcAY634e_P_QBO-fpJX).  There are two JSON files there with 400 and 407 users in them, which were generated from the grade book CSV file.  The larger file is meant to test the 400-user limit of the API entry.

lsloan:
> Using the Got retry feature, this implementation is often (usually?), but not always, successful in enrolling a full set of 400 students. It's ignoring the Canvas API throttling and essentially using brute force. From team discussions, this is the technique that I was directed to pursue. It may be important to revisit this to use request throttling before merging this PR. I don't know, that's up to the reviewers and testers to determine. I've received conflicting directions on this matter.

The quoted text above is from the description of this PR.  That is what I referred to in the meeting this morning when I brought up the subject of possibly needing request throttling for this feature.

ccm_web/server/src/canvas/canvas.service.ts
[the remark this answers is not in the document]
lsloan: got@11.8.2 doesn't appear to have any `statusCodes` defined by default, which is probably why its retry feature didn't work as I expected until after I added `403`.  I'm hard-coding the list with the codes from got@12 and adding `403`.  In the future, we can replace the hard-coded list by replacing it with something like `defaultInternals.retry.statusCodes.concat(403)`.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: AFAIK, a user cannot have `name@med.umich.edu` without also having `name@umich.edu`. The `med` domain will not have anything that `umich.edu` will not.  We're also opening a can of worms here, I think, because we may have to start considering other domains like `engin.umich.edu` and others.

lsloan: I've not added code to handle the `med.umich.edu` addresses yet.  That may be an overoptimization.  We don't really know whether our users will need it.  (We don't even know whether they will need the `umich.edu` handling, either.)

I will open a new issue to consider handling a list of email domains to strip and to make it configurable.

lsloan:
@ssciolla and @chrisrrowland, I think this is finally ready for last review.  I've created some issues to cover some possible new features or changes.

If you're able to approve the PR, I can merge it.

ssciolla:
Looks okay, I will take another look in the morning.

ccm_web/server/src/api/api.enroll.section.users.handler.ts
lsloan: Opened #192.

ccm_web/server/src/api/api.service.ts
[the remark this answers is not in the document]
lsloan: This is a good idea, but wasn't essential for completing this issue.  I'll open a new issue to implement this in the future.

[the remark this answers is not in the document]
lsloan: Opened #193.

ccm_web/server/src/api/api.utils.ts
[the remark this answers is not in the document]
lsloan: Renamed to "Larry".

ccm_web/server/src/api/api.controller.ts
```diff
@@ -59,4 +62,13 @@ export class APIController {
     if (isAPIErrorData(result)) throw new HttpException(result, result.statusCode)
     return result
   }
+
```
ssciolla, line 65: There's a conflict in this file. I think we need to re-base and then add the `UseInterceptors(InvalidTokenInterceptor)`, introduced in my OAuth PR (#183), to this route.

ccm_web/server/src/canvas/canvas.service.ts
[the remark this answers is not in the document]
ssciolla: Thanks for figuring out this retry stuff and making sure it's working.

ccm_web/server/src/api/api.controller.ts
lsloan: OK.  I don't rebase often, so let me check that out.

lsloan: Conflict resolved; decorator added.

ccm_web/server/src/canvas/canvas.service.ts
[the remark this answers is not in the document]
ssciolla: Are you sure there are no status codes in 11.8.2? The docs from `npm` seem to suggest otherwise: https://www.npmjs.com/package/got/v/11.8.2#retry

Not sure if that means you can do your default thingy now, but I was confused by your comment.

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -14,11 +14,51 @@ import { User } from '../user/user.model'
...
+  'x-request-cost'?: string
+}
+
+const requestorOptions: GotOptions = {
+  retry: {
+    limit: 2,
+    methods: ['POST', 'GET', 'PUT', 'DELETE'],
+    // TODO: After got@12 upgrade, replace following list with something like…
+    // got.defaultInternals.retry.statusCodes.concat(403)
+    statusCodes: [403, 408, 413, 429, 500, 502, 503, 504, 521, 522, 524]
```
ssciolla, line 34: There is a `calculateDelay` option that allows for some pretty sophisticated logic on spacing out the retries. I did one test run with 400 enrollments with the options as is, and 13 of them failed. Then, I tried adding a very simplistic value for `calculateDelay` of `() => { return 1000 }` and all 400 of them went through. Not even sure a full second is necessary. Might be worth adding something like this while we discuss throttling, etc.

ssciolla (approved):
Thanks for your work on this, and the requested changes you've already made. I think we can merge this. I offered one possible solution to the "things sometimes fail" problem using `options.retry.calculateDelay`. Feel free to merge when you think it's ready.

ccm_web/server/src/canvas/canvas.service.ts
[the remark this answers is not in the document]
lsloan: I saw the same documentation here: https://github.com/sindresorhus/got/blob/v11.8.2/readme.md#retry

However, when I browsed through the code for 11.8.2 (in the GitHub repo web UI and inspecting the version that was installed on my machine), I couldn't find any code that was setting those defaults.

I concluded the documentation must be faulty.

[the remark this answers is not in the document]
ssciolla: @lsloan, what about here? https://github.com/sindresorhus/got/blob/v11.8.2/source/index.ts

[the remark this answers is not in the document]
lsloan: Wow!  OK.  I don't understand.  I was looking through the code for all mentions of `statusCodes` and some of the specific code numbers mentioned in the docs, like `429`, `524`, etc., but I didn't find it.  That's weird.

Now that I know it's there, I think I figured out why I didn't see it.  I think my IDE ignores JS files when there are TS files present, because it assumes the JS files are generated/compiled from TS files.

OK.  I'll add an issue to remove the hard-coded list.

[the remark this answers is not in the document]
lsloan: Opened #194.

PR #168, opened 2021-07-29
Add request logging approach using `morgan`, `winston` (#64)

### LINKED ISSUE #64: Add request logging with URL, method, and other details
Like with our other applications, we will want some kind of request logging to give us insight into how the application is being used. How we choose to implement this will depend in part on what API framework we choose (vanilla JS, Nest.JS or something else).

ssciolla:
This PR adds `morgan` (a popular request logging Express middleware) as global middleware, using the preset format `combined` and writing the log messages using the existing `winston` logger object. Please feel free to suggest an alternate format (see https://www.npmjs.com/package/morgan#predefined-formats). The PR aims to resolve #64.

Resource(s) referenced: 
- https://www.npmjs.com/package/morgan
- https://stackoverflow.com/a/56855684
- https://stackoverflow.com/questions/27906551/node-js-logging-use-morgan-and-winston


lsloan (commented):
I'm amazed at how simple is the code required to add morgan to the app.  I don't see how it actually works, though.  Does morgan intercept every log message sent through winston and apply its formatting?

What does the `logger.info(…)` part mean?  Is its formatting applied only to INFO level messages?  Or is morgan then logging every message with that log level?

ssciolla:
> I'm amazed at how simple is the code required to add morgan to the app.  I don't see how it actually works, though.  Does morgan intercept every log message sent through winston and apply its formatting?

This is Express middleware, and because it's being applied to the Nest app in `bootstrap`, it's global (it could also be applied in a module). Express middleware has access to the request and response objects, as well as next function to instruct the next middleware to proceed. Because it's globally applied, `morgan` logs details about the request/response cycle every time, on each route.

> What does the `logger.info(…)` part mean?  Is its formatting applied only to INFO level messages?  Or is morgan then logging every message with that log level?

The stream part allows us to write the `morgan`-generated message using the `winston` logger (rather than `console.log` say), so that we still a consistent format. I decided that all such request log messages should be reported at the INFO level, but this can be changed if necessary

I hope that helps explain. Please also look at the resources linked above too. I can also provide resources about Express middleware if you want them.

lsloan:
Thanks!  Yes, I understand that a bit better now.

lsloan (approved):
After discussing it with @ssciolla, I think this looks good.

:shipit:

ssciolla:
Thanks for the review, @lsloan!

PR #174, opened 2021-08-16
Get ready for OpenShift deployment (#42, #175)

### LINKED ISSUE #42: Investigate production deployment strategies, host/IP address handling
We are going to need to learn about how to deploy an Express/NodeJS application in an OpenShift environment. At some point after work on issue #23 is completed, we could attempt to set up a test instance, and this would be easier to test then. Production host name or IP address configuration might only need to be adjusted in OpenShift, but some code changes might be necessary.

Possibly relevant resources:
* https://www.digitalocean.com/community/tutorials/how-to-set-up-a-node-js-application-for-production-on-ubuntu-14-04
* https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs/deployment
* https://developers.redhat.com/blog/2019/08/30/easily-deploy-node-js-applications-to-red-hat-openshift-using-nodeshift/
* https://expressjs.com/en/advanced/best-practice-performance.html#use-a-reverse-proxy

ssciolla:
This PR prepares the project for OpenShift deployments by doing the following:
- Implementing the `pm2` process manager (in `cluster` mode) for production builds so we can utilize all cores in OpenShift pods, improving performance and resiliency in case of errors
- Creating an `init` script that starts and stops the NestJS application to ensure the configuration is valid and the `ltijs` database tables are initialized before the application is clustered (might be a better way to do this, still thinking about it)
- Working around an issue in OpenShift 3 for deployments by pulling from `node:14-slim` in `build` `Dockerfile` stage (instead of from the `base` stage)
- Working around Docker rate limits by creating an alternate `Dockerfile.openshift` (hopefully we can remove this when we migrate to OpenShift 4)
- Hardening configuration logic so that the `fallback` (read default) is only used if the variable is `undefined`; only non-`undefined` candidate `isNumber` values are cast using `Number()`; empty strings are considered invalid; and extra runtime checks are separated from type guards (#175).
- Enabled graceful shutdown in production by ensuring `SIGTERM` signals are passed from OpenShift to `start.sh` to `pm2-runtime`, and that `SIGINT` signals are passed from `pm2` to each child thread.

The PR aims to resolve #42 and #175.

### DIFF
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -3,7 +3,7 @@
 # - https://dev.to/chrsgrrtt/dockerising-a-next-js-project-1ck5
 
 # Base stage
-FROM node:14-slim AS ccm-base
+FROM node:14-slim AS base
 WORKDIR /base/
 COPY package*.json ./
 RUN npm install
@@ -13,17 +13,17 @@ ARG PORT
 EXPOSE ${PORT}
 
 # Build stage (build client and compile server to JS)
-FROM ccm-base AS ccm-build
+FROM node:14-slim AS build
 WORKDIR /build/
-COPY --from=ccm-base /base ./
+COPY --from=base /base ./
 RUN npm run build
 
 # Prod stage
-FROM node:14-slim AS ccm-prod
+FROM node:14-slim AS prod
 ENV NODE_ENV=production
 WORKDIR /app
 
-COPY --from=ccm-base /base/package.json \
+COPY --from=base /base/package.json \
     /base/package-lock.json \
     /base/start.sh /base/ecosystem.config.js \
     ./
@@ -31,7 +31,7 @@ RUN npm install --production
 
 RUN npm install pm2@5.1.0 -g
 
-COPY --from=ccm-build /build/dist/ ./
+COPY --from=build /build/dist/ ./
 
 ARG PORT
 EXPOSE ${PORT}
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -35,6 +35,8 @@ RUN npm install --production
 RUN npm install pm2@5.1.0 -g
 COPY --from=build /build/dist/ ./
 
+# Change HOME so that .pm2 files can be written to file system
+ENV HOME=/app/
 ARG PORT
 EXPOSE ${PORT}
 CMD ["sh", "-c", "./start.sh"]
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -41,6 +41,6 @@ RUN chmod -R g+rw /app/
 ENV HOME=/app/
 ARG PORT
 EXPOSE ${PORT}
-CMD ["sh", "-c", "./start.sh"]
+ENTRYPOINT ["./start.sh"]
 
 # Done!
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -23,8 +23,13 @@ FROM node:14-slim AS prod
 ENV NODE_ENV=production
 WORKDIR /app
 
-COPY --from=base /base/package.json /base/package-lock.json /base/start.sh ./
+COPY --from=base /base/package.json \
+    /base/package-lock.json \
+    /base/start.sh /base/ecosystem.config.js \
+    ./
 RUN npm install --production
+
+RUN npm install pm2@5.1.0 -g
 
 COPY --from=build /build/dist/ ./
 
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -34,11 +34,9 @@ COPY --from=base \
 RUN npm install --production
 RUN npm install pm2@5.1.0 -g
 COPY --from=build /build/dist/ ./
-# Let members of root group write to /app/
-# RUN chmod -R g+rw /app/
 
-# Change HOME so that .pm2 files are written in /app/
-ENV PM2_HOME=/tmp/
+# Set PM2_HOME so that .pm2 files are written in /tmp/
+ENV PM2_HOME=/tmp/.pm2
 ARG PORT
 EXPOSE ${PORT}
 ENTRYPOINT ["./start.sh"]
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -3,7 +3,7 @@
 # - https://dev.to/chrsgrrtt/dockerising-a-next-js-project-1ck5
 
 # Base stage
-FROM node:14-slim AS base
+FROM node:14-slim AS ccm-base
 WORKDIR /base/
 COPY package*.json ./
 RUN npm install
@@ -13,17 +13,17 @@ ARG PORT
 EXPOSE ${PORT}
 
 # Build stage (build client and compile server to JS)
-FROM base AS build
+FROM ccm-base AS ccm-build
 WORKDIR /build/
-COPY --from=base /base ./
+COPY --from=ccm-base /base ./
 RUN npm run build
 
 # Prod stage
-FROM node:14-slim AS prod
+FROM node:14-slim AS ccm-prod
 ENV NODE_ENV=production
 WORKDIR /app
 
-COPY --from=base /base/package.json \
+COPY --from=ccm-base /base/package.json \
     /base/package-lock.json \
     /base/start.sh /base/ecosystem.config.js \
     ./
@@ -31,7 +31,7 @@ RUN npm install --production
 
 RUN npm install pm2@5.1.0 -g
 
-COPY --from=build /build/dist/ ./
+COPY --from=ccm-build /build/dist/ ./
 
 ARG PORT
 EXPOSE ${PORT}
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -5,6 +5,7 @@
 # Base stage
 FROM node:14-slim AS base
 WORKDIR /base/
+
 COPY package*.json ./
 RUN npm install
 COPY . .
@@ -15,6 +16,7 @@ EXPOSE ${PORT}
 # Build stage (build client and compile server to JS)
 FROM node:14-slim AS build
 WORKDIR /build/
+
 COPY --from=base /base ./
 RUN npm run build
 
@@ -23,14 +25,14 @@ FROM node:14-slim AS prod
 ENV NODE_ENV=production
 WORKDIR /app
 
-COPY --from=base /base/package.json \
+COPY --from=base \
+    /base/package.json \
     /base/package-lock.json \
-    /base/start.sh /base/ecosystem.config.js \
+    /base/start.sh \
+    /base/ecosystem.config.js \
     ./
 RUN npm install --production
-
 RUN npm install pm2@5.1.0 -g
-
 COPY --from=build /build/dist/ ./
 
 ARG PORT
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -35,10 +35,10 @@ RUN npm install --production
 RUN npm install pm2@5.1.0 -g
 COPY --from=build /build/dist/ ./
 # Let members of root group write to /app/
-RUN chmod -R g+rw /app/
+# RUN chmod -R g+rw /app/
 
 # Change HOME so that .pm2 files are written in /app/
-ENV HOME=/app/
+# ENV HOME=/app/
 ARG PORT
 EXPOSE ${PORT}
 ENTRYPOINT ["./start.sh"]
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -34,8 +34,10 @@ COPY --from=base \
 RUN npm install --production
 RUN npm install pm2@5.1.0 -g
 COPY --from=build /build/dist/ ./
+# Let members of root group write to /app/
+RUN chmod -R g+rw /app/
 
-# Change HOME so that .pm2 files can be written to file system
+# Change HOME so that .pm2 files are written in /app/
 ENV HOME=/app/
 ARG PORT
 EXPOSE ${PORT}
--- a/ccm_web/Dockerfile
+++ b/ccm_web/Dockerfile
@@ -38,7 +38,7 @@ COPY --from=build /build/dist/ ./
 # RUN chmod -R g+rw /app/
 
 # Change HOME so that .pm2 files are written in /app/
-# ENV HOME=/app/
+ENV PM2_HOME=/tmp/
 ARG PORT
 EXPOSE ${PORT}
 ENTRYPOINT ["./start.sh"]
--- a/ccm_web/Dockerfile.openshift
+++ b/ccm_web/Dockerfile.openshift
@@ -29,6 +29,6 @@ RUN chmod -R g+rw /app/
 ENV HOME=/app/
 ARG PORT
 EXPOSE ${PORT}
-CMD ["sh", "-c", "./start.sh"]
+ENTRYPOINT ["./start.sh"]
 
 # Done!
--- a/ccm_web/Dockerfile.openshift
+++ b/ccm_web/Dockerfile.openshift
@@ -22,8 +22,10 @@ COPY --from=build \
 RUN npm install --production
 RUN npm install pm2@5.1.0 -g
 COPY --from=build /build/dist/ ./
+# Let members of root group write to /app/
+RUN chmod -R g+rw /app/
 
-# Change HOME so that .pm2 files can be written to file system
+# Change HOME so that .pm2 files are written in /app/
 ENV HOME=/app/
 ARG PORT
 EXPOSE ${PORT}
--- a/ccm_web/Dockerfile.openshift
+++ b/ccm_web/Dockerfile.openshift
@@ -22,11 +22,9 @@ COPY --from=build \
 RUN npm install --production
 RUN npm install pm2@5.1.0 -g
 COPY --from=build /build/dist/ ./
-# Let members of root group write to /app/
-# RUN chmod -R g+rw /app/
 
-# Change HOME so that .pm2 files are written in /app/
-ENV PM2_HOME=/tmp/
+# Set PM2_HOME so that .pm2 files are written in /tmp/
+ENV PM2_HOME=/tmp/.pm2
 ARG PORT
 EXPOSE ${PORT}
 ENTRYPOINT ["./start.sh"]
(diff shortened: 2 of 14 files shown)


ccm_web/server/src/main.ts
```diff
@@ -83,9 +85,11 @@ async function bootstrap (): Promise<void> {
     SwaggerModule.setup('swagger', app, document)
   }
 
+  if (!isDev) app.enableShutdownHooks(['SIGINT'])
```
ssciolla, line 88: This and some some of the Docker artifact and `.sh` script changes allow us to shutdown gracefully in production with `pm1`, which sends `SIGINT` signals to the child processes int he cluster.

We could also set up the application to shutdown gracefully in development, but I think that would require changing how we use `npm` scripts to start things, which started to feel like too much, or out of scope. It's more important for production, but if it's desirable, I can open a separate issue.

ccm_web/server/src/init.ts
```diff
@@ -0,0 +1,22 @@
...
+async function bootstrap (): Promise<void> {
+  const app = await NestFactory.createApplicationContext(AppModule, {
+    logger: ['error', 'warn']
+  })
+  await app.close()
+}
+
+bootstrap()
+  .then(() => logger.info('ltijs has been initialized and configuration is valid.'))
+  .catch((e) => logger.error('An error occurred while initializing: ', e))
```
ssciolla, line 22: I keep going back and forth about whether this makes sense. Maybe there's an alternative I'm not thinking of. I'm trying to avoid the case of either 1) `ltijs` migrations haven't been set up before clustering, and the setup process errors out in the threads that lag behind the first and 2) the configuration is invalid, and all the threads emit repetitive errors. Maybe neither of those scenarios is that big of a deal (the first only happens with an empty database, the second when the configuration is wrong, and it should error out, just not for each thread ideally). Still it seems somewhat logical to try to start one application successfully before starting several. I didn't see any `pm2` options like this, though you could probably do it with a few `pm2` commands (e.g. `start`, `scale`).

Even assuming this is a good idea, I wasn't exactly sure how to implement it.  I chose `createApplicationContext` -- see NestJS [standalone applications](https://docs.nestjs.com/standalone-applications) -- because it allowed for a more bare-bones start and the network listeners didn't seem necessary for these two checks. However, maybe it's better to use all the same middleware configurations that are in the `main.ts` `bootstrap` function. Alternatively, I could be even more minimal and selective, and call the `validateConfig` function directly from `config.ts` and only `createApplicationContext` with `LTIModule` (using a similar approach, though I'd need to play around with how it could be configured separately, but pretty sure it's possible).

Sorry, maybe this is confusing or too much detail, but I thought I'd lay it all out in case someone was interested and had an opinion. What I currently have "works", but may not be optimal for one or more reasons (listed above or not).

ssciolla:
I'm still looking at this and testing it (and may change a couple files), but I want to open it up for review so we can start discussing some of the choices and configuration.

ccm_web/ecosystem.config.js
```diff
@@ -0,0 +1,14 @@
+// https://pm2.keymetrics.io/docs/usage/docker-pm2-nodejs/
+
+module.exports = {
+  apps: {
+    name: 'CCM - NestJS app',
+    script: 'server/src/main.js',
+    exec_mode: 'cluster',
+    instances: 'max',
+    autorestart: true,
+    max_memory_restart: '1G',
```
ssciolla, line 10: 8GBs~ per pod? Maybe bump this to 1.8?

ccm_web/Dockerfile.openshift
```diff
@@ -0,0 +1,32 @@
...
+
+COPY --from=build \
+    /build/package.json \
+    /build/package-lock.json \
+    /build/start.sh \
+    /build/ecosystem.config.js \
+    ./
+RUN npm install --production
+RUN npm install pm2@5.1.0 -g
+COPY --from=build /build/dist/ ./
```
ssciolla, line 24: Reference `--from=0` instead of `--from=build`?

ccm_web/ecosystem.config.js
jonespm: Looks like this is "maximum memory amount before PM2 restarts your app" so in here to prevent a process that takes up too much memory. I wonder if this option should just be able to be dynamically set, like in the environment? 

I'd think you can pass some of these args at runtime in the start script to pm2-runtime. I'd probably set this as a default like we're doing with the MyLA start script but make it configurable. 

It would probably be a calculation of like `(90% * [Max Memory])/[# of threads]`.

ccm_web/Dockerfile.openshift
jonespm: Didn't know [it numbered the FROMs.](https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages)

I'd agree with Chris that if it's working don't change anything with this now. This isn't something going away in future 3.x releases.

ccm_web/ecosystem.config.js
ssciolla: @jonespm, are you thinking this would just be a simple environment variable, like `pm2-runtine ecosystem.config.js --max_memory_restart=$PM2_MEM_LIMIT`, or would we want to do something more fancy to calculate it on the fly? It looks like you can get limits from OpenShift using a pod template and environment variables, but I don't exactly know how we would do that with a DeploymentConfig so it happens with each new pod (https://docs.openshift.com/container-platform/4.1/nodes/containers/nodes-containers-downward-api.html); even then, I don't know if I can be a 100 percent certain how many processes `pm2` will spawn.

We could also do this in the `ecosystem.config.js` file, something like `max_memory_restart: process.env.PM2_MEM_LIMIT || '1G'`. Does command line arg or this seem better to you?

jonespm: Nope, I think we'd just pass it in and have a default like

```
if [ -z "${PM2_MEM_LIMIT}" ]; then
    PM2_MEM_LIMIT=1
fi
```

I think either way is fine, parameter or js. Didn't realize this was a javascript file and not a plain configuration file. That looks good to me.

jonespm: I guess you don't even have to check and set the default in the config if you're setting a default in the javascript, so your suggestion alone seems fine too.

jonespm: There are ways we could calculate it from the start script too like.

```
MEM_TOTAL=$(grep MemTotal /proc/meminfo | awk '{print $2}');NUM_THREADS=$(grep -c processor /proc/cpuinfo)

MEM_PER_PROCESSOR=$(expr ${MEM_TOTAL} / ${NUM_THREADS})
```

This would be in KB, or just divide by the 1000. Not sure if this is *really* necessary though.

ssciolla: That's cool, though it seems a tad fragile with the string parsing. I also tried it in the terminal for CCM Dev, and I don't understand the result; it looks like it's 8 GB per thread, which is not what I expected.

ccm_web/server/src/main.ts
```diff
@@ -83,9 +85,11 @@ async function bootstrap (): Promise<void> {
...
+
   await app.listen(
     serverConfig.port,
-    () => logger.info(`Server started on localhost and port ${serverConfig.port}`)
+    () => logger.info(`Server started on 0.0.0.0 and port ${serverConfig.port}`)
```
pushyamig, line 92: I looked at the Openshift setup log and this statement is repeated 4 time. Why is this behavior?

pushyamig:
When I run this command locally `docker-compose -f docker-compose-prod.yml up`. I don't see the application is starting up. any reason why this is broken?

pushyamig:
The `.env` file properties is showcased as secrets in Openshift. In my mind I would make then as config-map since that is easy to edit rather secrets where you have to use commandline. We can deal this later if we need to but wanted to understand why set leaned toward `secrets` vs `config-map`?

But I do see that you have used configMap for changing the log level. any reason for this setup?

ccm_web/server/src/main.ts
ssciolla: Well, `pm2` starts up four different server instances, so these log statements will show up 4 times. I could add a process ID to the log messages using `winston` I think, though I wasn't sure if that was needed.

ssciolla:
> When I run this command locally `docker-compose -f docker-compose-prod.yml up`. I don't see the application is starting up. any reason why this is broken?

What error are you seeing? I am having no problems with this. Feel free to DM me.

ccm_web/server/src/main.ts
pushyamig: does pm2 starts 4 instance by default? since I did not see a configuration around it?

ssciolla:
> The `.env` file properties is showcased as secrets in Openshift. In my mind I would make then as config-map since that is easy to edit rather secrets where you have to use commandline. We can deal this later if we need to but wanted to understand why set leaned toward `secrets` vs `config-map`?
> 
> But I do see that you have used configMap for changing the log level. any reason for this setup?

There's debate on this question. Mark advocated for secrets, as they are 'supposed' to be used to store sensitive data (which our environment variables contain) and may in the future offer better security (unclear whether they actually do now) or features (like secret key rotation). Matt seems to disagree. I actually don't care that much, although it would be nice to know when folks are changing these values. If you want me to change it to a ConfigMap, I can.

The ConfigMap that was already there was created by Matt to demonstrate something to me, I think.

ccm_web/server/src/main.ts
ssciolla: See `ecosystem.config.js`. We are using `cluster` mode and a `max` setting for instances. 4 seems to be what OpenShift pods want to go to so far.

pushyamig: ok

jonespm:
The way we have it configured, a Secret is *really* just a harder to edit Base64 encoded Config Map. The recommendation is to store sensitive information in secrets but from what I can tell it's only real advantage at present is being able to store binary data.

There's [future ideas](https://blog.aquasec.com/managing-kubernetes-secrets) such as third party secret management, auditing and access control that can be setup around secrets. But I don't think we're taking advantage of most/any of that yet. I did switch some things that were stored as Secrets to Config Maps because it was easier to edit. Would like to switch them all to something else in the future, like AWS Secrets Manager or Hashicorp Vault/CyberArk mentioned in that article.

pushyamig:
> Mark advocated for secrets, as they are 'supposed' to be used to store sensitive data (which our environment variables contain) and may in the future offer better security (unclear whether they actually do now) or features (like secret key rotation). @jonespm seems to disagree. 

OK, I vote for ConfigMap since for our applications in the past we were using Secrets and that time Jeff suggested this new ConfigMap option and all our application adopted for this approach except for old CCM. Since you are leading the effort and @jonespm  leads the ops stuff at TL. I will leave the decision to both of you

ccm_web/server/src/init.ts
pushyamig: @ssciolla Thanks for additional details. I really don't think I have a opinion. Lets keep this way and take it as we go. Thanks for thinking through on this.

ssciolla:
> OK, I vote for ConfigMap since for our applications in the past we were using Secrets and that time Jeff suggested this new ConfigMap option and all our application adopted for this approach except for old CCM. Since you are leading the effort and @jonespm leads the ops stuff at TL. I will leave the decision to both of you

I will change it to a ConfigMap.

pushyamig:
Nice Work @ssciolla . I reviewed and ran locally and the setup seems to be working. More than Me @jonespm and @lsloan were in the discussions with Ops as well. I will lean on making the approval decision at least from them. This looks good and can be merged.

ccm_web/ecosystem.config.js
ssciolla: Okay, I did the minimum here for now: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/174/commits/b12c0194a5fa7fc069234aacd33198ed5250cf3f

I think if I understood why we're getting that result, I'd consider using your approach. Something to think about for the future.

jonespm: Yeah, hmm, /proc/meminfo says it has 32GB, rather than 8.

MemTotal:       32393208 kB

It looks like for [Openshift it should look at ](https://stackoverflow.com/a/63047839/3708872)
cat /sys/fs/cgroup/memory/memory.limit_in_bytes

This is even more direct, and the number of threads [is similarly direct.
](https://stackoverflow.com/a/54112066/3708872)

```
MEM_TOTAL=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)
NUM_THREADS=$(nproc) 
MEM_PER_PROCESSOR=$(expr ${MEM_TOTAL} / ${NUM_THREADS})
```

ssciolla: I probably spent too much time today looking into this. I think something like this would work in OpenShift right now:

```
TOTAL_MEM=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)
NUM_PROC=$(nproc)
PM2_MEM_LIMIT_CALC=$(($TOTAL_MEM * 9 / 10 / $NUM_PROC))
```

but the paths seem to be set up differently in a Docker container. The `cat` line returns a ridiculously long number (`9223372036854771712`). I guess I'm also concerned that this path might change. Looks like there is a `cgget` utility library that might help out. However, I'm a little concerned that things might change in OpenShift, and I wonder if simply making it directly configurable would save us time and heartache over a long period.

ccm_web/server/src/config.ts
```diff
@@ -43,12 +43,21 @@ export interface Config {
 
 const logger = baseLogger.child({ filePath: __filename })
 
-const isString = (v: unknown): v is string => typeof v === 'string'
+// Tests if an unknown value is a non-empty string
+const isString = (v: unknown): v is string => typeof v === 'string' && v.length > 0
+
+// Tests if an unknown value is a number and not NaN
 const isNumber = (v: unknown): v is number => typeof v === 'number' && !isNaN(v)
```
ssciolla, line 50: I'm taking another look at these type guards; upon reflection, I'm not sure I should be adding more test conditions to them -- though it doesn't throw any errors. It might be clearer and more flexibile if the tests were separate. the `!isNan(v)` and `v.length > 0` tests can't be inferred from subsequent typing, I don't think. I'd like to do some refactoring. I will create a separate issue and link it to this PR.

ssciolla: Okay, see #175 and https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/174/commits/b989b87ad10402883b945ad25c843e21fa92353b

ssciolla:
> More than Me @jonespm and @lsloan were in the discussions with Ops. I will lean on making the approval decision at least from one of them. This looks good and can be merged.

Okay, I'm not planning on touching this anymore until and if I get more feedback. Sorry for the late-coming commit on the configuration. I just ended up thinking (almost compulsively) about it late last night and decided I wanted to change it.

@jonespm and @lsloan, @pushyamig seems to be waiting for an approval from one of you (though Lance wasn't at the operations meeting). If you have the time, I would greatly appreciate it. Harsh criticism and/or ideas on different paths are welcome, and I am happy to spend time explaining any pieces if that would be helpful. I just want to keep it moving forward.

ccm_web/ecosystem.config.js
jonespm: Yup, fine with just leaving it directly configurable. Humans can figure this out and the 1GB default seems like it should always be fine for us.

ccm_web/Dockerfile
```diff
@@ -23,13 +25,20 @@ FROM node:14-slim AS prod
...
-COPY --from=base /base/package.json /base/package-lock.json /base/start.sh ./
+COPY --from=base \
+    /base/package.json \
+    /base/package-lock.json \
+    /base/start.sh \
+    /base/ecosystem.config.js \
+    ./
 RUN npm install --production
-
+RUN npm install pm2@5.1.0 -g
```
jonespm, line 35: Is there a reason this is a separate step with a hardcoded version here in each of the Dockerfile's rather than just part of the package.json?

jonespm (approved):
I know we've talked about this and looks good to me.

ccm_web/Dockerfile
ssciolla: Good question, not 100 percent certain I made the right choice. We don't use `pm2` in development. I guess if you do it this way, `package.json` doesn't prescribe a certain way to deploy your application, so you could write different `Dockerfile` and `start.sh` files to use an alternative if you wanted, adding some flexbility. ROHQ follows a similar approach actually, I'm realizing: https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/b076017a63f52e44dd2b5140d36ef1c2eeca8c9f/src/Dockerfile#L27-L29 Honestly, though I think the major reason I did it this way was `pm2` docs said to: https://pm2.keymetrics.io/docs/usage/docker-pm2-nodejs/ They're saying to install it globally in the container, so I guess I wasn't sure if I could do it in `package.json`. I could try if you think that would be better.

ssciolla: I could also allow for patch upgrades here, I think. Pinning to some version range made sense to me though.

lsloan:
> @jonespm and @lsloan, @pushyamig seems to be waiting for an approval from one of you (though Lance wasn't at the operations meeting).

My apologies for not attending that meeting last week.  I was an optional attendee.  Although I wanted to attend, at the time I was busy working on how to handle the "rate limit exceeded" error in the enrollment API entry.

ccm_web/server/src/config.ts
lsloan: I like that you ended up separating the string type checking and length checking.

ccm_web/Dockerfile
ssciolla: @jonespm, any additional feedback here? Thinking about merging.

ccm_web/server/src/config.ts
```diff
@@ -44,60 +44,83 @@ export interface Config {
...
+
+// Tests if an unknown value is one of four allowed string log levels
 const isLogLevel = (v: unknown): v is LogLevel => {
   return isString(v) && ['debug', 'info', 'warn', 'error'].includes(v)
 }
 
+// Handles some edge cases and casts all other values using Number
+const prepNumber = (value: string | undefined): string | number | undefined => {
+  return (value === undefined) ? undefined : value.trim() === '' ? value : Number(value)
+}
```
ssciolla, line 60: I did some last minute testing of these config changes, and I found an edge case I wasn't properly handling. There's an oddity with `Number()` that empty or all-whitespace strings are converted to `0`, which didn't seem like desirable behavior. Now I'm just returning those values so that they don't pass the type check. This number handling is a pain 😢

ssciolla:
Okay, I've reviewed my own work again, and after one more last-minute change, I think this will work for now. Since @pushyamig reviewed and @jonespm approved, I will merge. We will likely revisit some aspects of this later. There shouldn't be any conflicts between this and the other pending PRs.

PR #176, opened 2021-08-24
#74 adding the missing course setting link

### LINKED ISSUE #74: Canvas Gradebook link to Canvas Settings
When the user uploads a csv that is missing letter grades, the error includes a link to go to the canvas settings for the course to enable the feature to include them in the csv.

Linking to this requires some back end support that was not yet ready when working on the initial feature implementation.

There is currently a non functional place holder link ([see in PR](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/65/files#diff-b377a52c0bd43ebc14a000a38d97d299f4d36f721063eef339762b6a7e28a46aR206)) that should be replaced with a functional one.

pushyamig:
Fixes #74 

Adding the link to the view when you error is caught when upload the no grading schema applied [sample csv](https://drive.google.com/drive/folders/11VARkGgQUpxrDZ3yIiY6zeivJlPrQG-j)

Screen Shot 2021-08-24 at 12 49 43 PM

### DIFF
--- a/ccm_web/client/src/pages/GradebookCanvas.tsx
+++ b/ccm_web/client/src/pages/GradebookCanvas.tsx
@@ -183,8 +183,8 @@ function ConvertCanvasGradebook (props: CCMComponentProps): JSX.Element {
 
   const handleNoLetterGradesError = (): void => {
     const { canvasURL, course } = props.globals
-    const settingsURL = `${canvasURL}/courses/${course.id}/settings`
-    setPageState({ state: GradebookCanvasPageState.InvalidUpload, errorMessage: [<Typography key='0'><Link href={settingsURL} target='_parent' rel='noopener'>Grading Scheme in settings</Link> needs to be check marked for letter grade to appear in the CSV file.</Typography>] })
+    const settingsURL = `${canvasURL}/courses/${course.id}/settings#course_grading_standard_enabled`
+    setPageState({ state: GradebookCanvasPageState.InvalidUpload, errorMessage: [<Typography key='0'><Link href={settingsURL} target='_parent'>Grading Scheme in settings</Link> needs to be check marked for letter grade to appear in the CSV file.</Typography>] })
   }
 
   const handleRowLevelInvalidationError = (errorMessage: JSX.Element[], invalidations: GradebookRowInvalidation[]): void => {
--- a/ccm_web/client/src/pages/GradebookCanvas.tsx
+++ b/ccm_web/client/src/pages/GradebookCanvas.tsx
@@ -184,7 +184,7 @@ function ConvertCanvasGradebook (props: CCMComponentProps): JSX.Element {
   const handleNoLetterGradesError = (): void => {
     const { canvasURL, course } = props.globals
     const settingsURL = `${canvasURL}/courses/${course.id}/settings`
-    setPageState({ state: GradebookCanvasPageState.InvalidUpload, errorMessage: [<Typography key='0'><Link href={settingsURL} target='_new' rel='noopener'>Grading Scheme in settings</Link> needs to be check marked for letter grade to appear in the CSV file.</Typography>] })
+    setPageState({ state: GradebookCanvasPageState.InvalidUpload, errorMessage: [<Typography key='0'><Link href={settingsURL} target='_parent' rel='noopener'>Grading Scheme in settings</Link> needs to be check marked for letter grade to appear in the CSV file.</Typography>] })
   }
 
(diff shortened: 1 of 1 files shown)


ccm_web/client/src/pages/GradebookCanvas.tsx
```diff
@@ -181,7 +182,9 @@ function ConvertCanvasGradebook (): JSX.Element {
   }
 
   const handleNoLetterGradesError = (): void => {
-    setPageState({ state: GradebookCanvasPageState.InvalidUpload, errorMessage: [<Typography key='0'><Link href='https://community.canvaslms.com/t5/Instructor-Guide/How-do-I-enable-a-grading-scheme-for-a-course/ta-p/1042' target='_new' rel='noopener'>Grading Scheme in settings</Link> needs to be check marked for letter grade to appear in the CSV file.</Typography>] })
+    const { canvasURL, course } = props.globals
+    const settingsURL = `${canvasURL}/courses/${course.id}/settings`
```
chrisrrowland, line 186: Do you think we should link to the page in general, or more specifically

const settingsURL = `${canvasURL}/courses/${course.id}/settings#course_grading_standard_enabled`

pushyamig: I thought about this, what if canvas changes the name at some point. I really don't know how often that happen so I left it generic. what you think? I can do either way but don't know what is best approach here

ccm_web/client/src/pages/GradebookCanvas.tsx
```diff
@@ -181,7 +182,9 @@ function ConvertCanvasGradebook (): JSX.Element {
...
+    setPageState({ state: GradebookCanvasPageState.InvalidUpload, errorMessage: [<Typography key='0'><Link href={settingsURL} target='_new' rel='noopener'>Grading Scheme in settings</Link> needs to be check marked for letter grade to appear in the CSV file.</Typography>] })
```
chrisrrowland, line 187: In BulkSectionCreate I opened with target='_parent'.  This seems like a similar scenario.  Do you think one makes more sense than the other?

pushyamig: Done

ccm_web/client/src/pages/GradebookCanvas.tsx
pushyamig: @jonespm what is your opinion on giving a direct link https://umich.beta.instructure.com/courses/403334/settings#course_grading_standard_enabled to a form option on canvas? 
I feel that these things could be easily renamed then these links will break and we need to have a hot fix for it.

chrisrrowland: I wondered that too.. it seems that when you have a non-existent anchor it just goes to the main page as if you didn't append one.  That seems fine, though it could break and nobody would notice.

It seems safe to add it on there since it can't really break, but it does introduce a dependency outside of our control that is not easy to keep track of.

I don't know if it's particularly important one way or the other..

pushyamig: @chrisrrowland I changed it as per your recommendation but it is not going to that exact location as we wanted. Infant it is default to setting APP tab.

chrisrrowland: Hmm.. so for me its resolving to https://canvas-test.it.umich.edu/courses/483458/settings#course_grading_standard_enabled and seems to work as expected

pushyamig: It's weird that it is not happening in Beta env like you described https://umich.beta.instructure.com/courses/403334/settings#course_grading_standard_enabled

pushyamig: Done

ccm_web/client/src/pages/GradebookCanvas.tsx
chrisrrowland: Thank you!

ccm_web/client/src/pages/GradebookCanvas.tsx
chrisrrowland: For posterity, the behavior here seems browser dependent and mildly inconsistent/infuriating.

Sometimes it works, sometimes it doesn't, but it seems like the best case is a slightly more user friendly experience, and the worst case is that it doesn't do anything.

chrisrrowland (approved):
Looking good.

PR #179, opened 2021-09-01
Bulk Section Create empty line validator

### LINKED ISSUE #160: Need a validator to prevent empty section names
Assuming we don't want to allow a blank section name (I assume that would an error on the backend)
We should implement an additional validator in BulkSectionCreateValidators to ensure the section name isn't blank (or an empty string)

chrisrrowland:
Fixes #160

Add a new empty section validator for bulk section creation
Also fixes an issue where blank lines were being ignored.. functionally ok but messed with line numbers in error messages.  This was present in bulk section create an
[IncludesABlank.csv](https://github.com/tl-its-umich-edu/canvas-course-manager-next/files/7093243/IncludesABlank.csv)
d bulk add UM users.

### DIFF
--- a/ccm_web/client/src/pages/BulkSectionCreate.tsx
+++ b/ccm_web/client/src/pages/BulkSectionCreate.tsx
@@ -333,7 +333,6 @@ function BulkSectionCreate (props: CCMComponentProps): JSX.Element {
       }
       handleParseSuccess(lines)
     }).catch(e => {
-      console.log(e)
       // TODO Not sure how to produce this error in real life
       handleSchemaError([{ error: 'Error processing file', type: InvalidationType.Error }])
     })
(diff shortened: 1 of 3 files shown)


ccm_web/client/src/pages/BulkSectionCreate.tsx
```diff
@@ -312,7 +312,7 @@ function BulkSectionCreate (props: CCMComponentProps): JSX.Element {
 
   const parseFile = (file: File): void => {
     file.text().then(t => {
-      let lines = t.split(/[\r\n]+/).map(line => { return line.trim() })
```
chrisrrowland, line 315: This was eating up consecutive newlines so any empty lines were getting ignored, which was fine if there were no errors, but if there was an error the line numbers might be incorrect.

ccm_web/client/src/pages/BulkSectionCreate.tsx
```diff
@@ -333,6 +333,7 @@ function BulkSectionCreate (props: CCMComponentProps): JSX.Element {
       }
       handleParseSuccess(lines)
     }).catch(e => {
+      console.log(e)
```
pushyamig, line 336: Do you want to leave this here? if yes, can we make it console.error(e). I think in past that's what we agreed.

pushyamig:
This Look Good. Thanks for thinking about this case and having a fix.

I tested with various scenarios. I couldn't break it.

pushyamig (approved):

PR #180, opened 2021-09-02
Validator(s) for section name too long

### LINKED ISSUE #163: Need a validator to prevent excessively long section names
Section names cannot be > 255 characters

chrisrrowland:
Fixes #163

### DIFF
--- a/ccm_web/client/src/utils/canvasSectionNameValidator.ts
+++ b/ccm_web/client/src/utils/canvasSectionNameValidator.ts
@@ -19,10 +19,19 @@ class DuplicateSectionNameValidator implements IValidator {
   }
 }
 
+class SectionNameTooLongValidator implements IValidator {
+  validate = (sections: CanvasCourseSection[], sectionName: string): ICanvasSectionNameInvalidError | undefined => {
+    if (sectionName.length > 255) {
+      return { reason: 'Section name too long.' }
+    }
+    return undefined
+  }
+}
+
 export class CanvasCoursesSectionNameValidator {
   course: Course
   _this = this
-  validators: IValidator[] = [new DuplicateSectionNameValidator()]
+  validators: IValidator[] = [new DuplicateSectionNameValidator(), new SectionNameTooLongValidator()]
 
   constructor (course: Course) {
     this.course = course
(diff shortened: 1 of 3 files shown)


ccm_web/client/src/utils/canvasSectionNameValidator.ts
```diff
@@ -19,10 +19,19 @@ class DuplicateSectionNameValidator implements IValidator {
   }
 }
```
pushyamig, line 21: Probably I did not notice this in with Bulk enroll user UI PR. Why is this file in `utils` folder? Since these are validation around create section widget. If you feel my suggestion is reasonable can you move this to components?

ccm_web/client/src/utils/canvasSectionNameValidator.ts
```diff
@@ -19,10 +19,19 @@ class DuplicateSectionNameValidator implements IValidator {
...
 
+class SectionNameTooLongValidator implements IValidator {
+  validate = (sections: CanvasCourseSection[], sectionName: string): ICanvasSectionNameInvalidError | undefined => {
+    if (sectionName.length > 255) {
+      return { reason: 'Section name too long.' }
```
pushyamig, line 25: I feel somebody putting a section name greater that 255 will never happen!  My simple suggestion here would be can we word this as  `Section name should be 255 characters or less`. Same text change goes above

pushyamig (approved):
The Code is working as expected. I felt the code could be generalized with section name length validation with Bulk create and section creation case since I see code duplication around it. But I feel that will be quite a bit of refactoring not needed now. 
I have couple of suggestions but I will leave up-to you if you want to fix them or not. Over all this can be merged as is

ccm_web/client/src/utils/canvasSectionNameValidator.ts
chrisrrowland: I think it makes more sense in utils.
It's utility/common functionality that could be used other places.  It doesn't really have any component-ness to it  IMO.

pushyamig: ok

ccm_web/client/src/utils/canvasSectionNameValidator.ts
chrisrrowland: Good call.

chrisrrowland:
> The Code is working as expected. I felt the code could be generalized with section name length validation with Bulk create and section creation case since I see code duplication around it. But I feel that will be quite a bit of refactoring not needed now.
> I have couple of suggestions but I will leave up-to you if you want to fix them or not. Over all this can be merged as is

Had you seen #181 ?

pushyamig:
> The Code is working as expected. I felt the code could be generalized with section name length validation with Bulk create and section creation case since I see code duplication around it. But I feel that will be quite a bit of refactoring not needed now.
I have couple of suggestions but I will leave up-to you if you want to fix them or not. Over all this can be merged as is
Had you seen #181 ?

I saw you opened an issue but never dig through more for details. Thanks for creating this issue.  I share same sentiments as you wrote in issue with code clean up

PR #183, opened 2021-09-08
Handle manually removed Canvas integration case (#101, #184)

### LINKED ISSUE #101: Dealing with when user invalidates the oAuth grant in canvas, outside of CCM
Test requests with using invalid Canvas token are properly redirected to OAuth flow
1. Launch application as normal.
2. Go to profile/settings in Canvas and delete all "Approved Integrations" entries for your API Developer Key.
3. Perform an action in the UI that involves a request to Canvas, checking that you are redirected to Canvas OAuth page.
4. Repeat steps 2 and 3 until all actions are tested.

Test requests involving a token refresh of an invalid Canvas token are properly redirect to OAuth flow

1. Launch application as normal.
2. Go to profile/settings in Canvas and delete all "Approved Integrations" entries for your API Developer Key.
3. Modify the expires_at value in the canvas_token record for your user in the database so it is in the past.
4. Perform an action in the UI that involves a request to Canvas, checking that you are redirected to Canvas OAuth page.
5. Repeat steps ii, iii, and iv until all actions are tested.

ssciolla:
The PR aims to resolve #101 and #184.

To Do
- [x] Implement an interceptor for catching errors related to an invalid token, deleting the token (if necessary), and sending custom response error data (including `redirect: true`).
- [x] Redirect in client to root using `location.href` when `401` is encountered and `redirect: true` is present in the error body (see `handleErrors`).
- [x] Modify globals to send a boolean indicating whether the user has a canvas token
- [x] Redirect in client using `location.href` to `canvas/redirectOAuth` if the user doesn't have a Canvas token
- [x] Always send the user to `/` first, instead of redirecting directly to `canvas/redirectOAuth` from `LTIService`
- [x] Enforce 1:1 relationship between `user` and `canvas_token` at database level by using unique constraint (#184)
 
Recommended testing procedures
- Ensure brand new user is successfully routed through OAuth and can subsequently make requests.
- Test requests with using invalid Canvas token are properly redirected to OAuth flow
  1. Launch application as normal.
  2. Go to `profile/settings` in Canvas and delete all "Approved Integrations" entries for your API Developer Key.
  3. Perform an action in the UI that involves a request to Canvas, checking that you are redirected to Canvas OAuth page.
  4. Repeat steps ii and iii until all actions are tested.
- Test requests involving a token refresh of an invalid Canvas token are properly redirect to OAuth flow
  1. Launch application as normal.
  2. Go to `profile/settings` in Canvas and delete all "Approved Integrations" entries for your API Developer Key.
  3. Modify the `expires_at` value in the `canvas_token` record for your user in the database so it is in the past.
  4. Perform an action in the UI that involves a request to Canvas, checking that you are redirected to Canvas OAuth page.
  5. Repeat steps ii, iii, and iv until all actions are tested.
- Test migration with `up` and `down` and `up` again (see `README.md`).

Resources
- https://docs.nestjs.com/interceptors
- https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html

### DIFF
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -45,12 +45,14 @@ function App (): JSX.Element {
   )
 
   useEffect(() => {
-    if (globals !== undefined) {
+    if (globals?.user.hasCanvasToken === true) {
       void doLoadCourse(globals.course.id)
     }
   }, [globals])
 
-  if (isAuthenticated === undefined || isLoading || isCourseLoading) return <div className='App'><p>Loading...</p></div>
+  const loading = <div className='App'><p>Loading...</p></div>
+
+  if (isAuthenticated === undefined || isLoading || isCourseLoading) return loading
 
   if (globalsError !== undefined) console.error(globalsError)
   if (csrfTokenCookieError !== undefined) console.error(csrfTokenCookieError)
@@ -60,6 +62,12 @@ function App (): JSX.Element {
         <p>You were not properly authenticated to the application.</p>
       </div>
     )
+  }
+
+  if (!globals.user.hasCanvasToken) {
+    // Initiate OAuth flow
+    location.href = '/canvas/redirectOAuth'
+    return loading
   }
 
   interface BreadcrumbProps {
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -10,6 +10,7 @@ import usePromise from './hooks/usePromise'
 import { CanvasCourseBase } from './models/canvas'
 import allFeatures from './models/FeatureUIData'
 import Home from './pages/Home'
+import redirect from './utils/redirect'
 import './App.css'
 
 const useStyles = makeStyles((theme) => ({
@@ -66,7 +67,7 @@ function App (): JSX.Element {
 
   if (!globals.user.hasCanvasToken) {
     // Initiate OAuth flow
-    location.href = '/canvas/redirectOAuth'
+    redirect('/canvas/redirectOAuth')
     return loading
   }
 
--- a/ccm_web/client/src/components/CreateSectionWidget.tsx
+++ b/ccm_web/client/src/components/CreateSectionWidget.tsx
@@ -1,11 +1,13 @@
-import { Button, Grid, makeStyles, TextField } from '@material-ui/core'
-import React, { ChangeEvent, useState } from 'react'
+import { CODE_NUMPAD_ENTER, CODE_RETURN } from 'keycode-js'
 import { useSnackbar } from 'notistack'
+import React, { ChangeEvent, useState } from 'react'
+import { Button, Grid, makeStyles, TextField } from '@material-ui/core'
+
 import { addCourseSections } from '../api'
 import { CanvasCourseSection } from '../models/canvas'
 import { CCMComponentProps } from '../models/FeatureUIData'
 import { CanvasCoursesSectionNameValidator, ICanvasSectionNameInvalidError } from '../utils/canvasSectionNameValidator'
-import { CODE_NUMPAD_ENTER, CODE_RETURN } from 'keycode-js'
+import { hasUnauthorized } from '../utils/handleErrors'
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -53,18 +55,22 @@ function CreateSectionWidget (props: CreateSectionWidgetProps): JSX.Element {
           .then(newSections => {
             props.onSectionCreated(newSections[0])
             setNewSectionName('')
-          }).catch(() => {
-            enqueueSnackbar('Error adding section', {
-              variant: 'error'
-            })
+          }).catch((e: Error) => {
+            if (hasUnauthorized([e])) {
+              location.href = '/'
+            } else {
+              enqueueSnackbar('Error adding section', { variant: 'error' })
+            }
           })
       } else {
         errorAlert(errors)
       }
-    }).catch(() => {
-      enqueueSnackbar('Error validating section name', {
-        variant: 'error'
-      })
+    }).catch((e: Error) => {
+      if (hasUnauthorized([e])) {
+        location.href = '/'
+      } else {
+        enqueueSnackbar('Error validating section name', { variant: 'error' })
+      }
     }).finally(() => {
       setIsCreating(false)
     })
--- a/ccm_web/client/src/components/CreateSectionWidget.tsx
+++ b/ccm_web/client/src/components/CreateSectionWidget.tsx
@@ -1,12 +1,11 @@
-import { CODE_NUMPAD_ENTER, CODE_RETURN } from 'keycode-js'
-import { useSnackbar } from 'notistack'
-import React, { ChangeEvent, useState } from 'react'
 import { Button, Grid, makeStyles, TextField } from '@material-ui/core'
-
+import React, { ChangeEvent, useState } from 'react'
+import { useSnackbar } from 'notistack'
 import { addCourseSections } from '../api'
 import { CanvasCourseSection } from '../models/canvas'
 import { CCMComponentProps } from '../models/FeatureUIData'
 import { CanvasCoursesSectionNameValidator, ICanvasSectionNameInvalidError } from '../utils/canvasSectionNameValidator'
+import { CODE_NUMPAD_ENTER, CODE_RETURN } from 'keycode-js'
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -54,14 +53,18 @@ function CreateSectionWidget (props: CreateSectionWidgetProps): JSX.Element {
           .then(newSections => {
             props.onSectionCreated(newSections[0])
             setNewSectionName('')
-          }).catch((e) => {
-            enqueueSnackbar('Error adding section', { variant: 'error' })
+          }).catch(() => {
+            enqueueSnackbar('Error adding section', {
+              variant: 'error'
+            })
           })
       } else {
         errorAlert(errors)
       }
-    }).catch((e: Error) => {
-      enqueueSnackbar('Error validating section name', { variant: 'error' })
+    }).catch(() => {
+      enqueueSnackbar('Error validating section name', {
+        variant: 'error'
+      })
     }).finally(() => {
       setIsCreating(false)
     })
--- a/ccm_web/client/src/components/CreateSectionWidget.tsx
+++ b/ccm_web/client/src/components/CreateSectionWidget.tsx
@@ -7,7 +7,6 @@ import { addCourseSections } from '../api'
 import { CanvasCourseSection } from '../models/canvas'
 import { CCMComponentProps } from '../models/FeatureUIData'
 import { CanvasCoursesSectionNameValidator, ICanvasSectionNameInvalidError } from '../utils/canvasSectionNameValidator'
-import { hasUnauthorized } from '../utils/handleErrors'
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -55,22 +54,14 @@ function CreateSectionWidget (props: CreateSectionWidgetProps): JSX.Element {
           .then(newSections => {
             props.onSectionCreated(newSections[0])
             setNewSectionName('')
-          }).catch((e: Error) => {
-            if (hasUnauthorized([e])) {
-              location.href = '/'
-            } else {
-              enqueueSnackbar('Error adding section', { variant: 'error' })
-            }
+          }).catch((e) => {
+            enqueueSnackbar('Error adding section', { variant: 'error' })
           })
       } else {
         errorAlert(errors)
       }
     }).catch((e: Error) => {
-      if (hasUnauthorized([e])) {
-        location.href = '/'
-      } else {
-        enqueueSnackbar('Error validating section name', { variant: 'error' })
-      }
+      enqueueSnackbar('Error validating section name', { variant: 'error' })
     }).finally(() => {
       setIsCreating(false)
     })
--- a/ccm_web/client/src/pages/AddUMUsers.tsx
+++ b/ccm_web/client/src/pages/AddUMUsers.tsx
@@ -14,6 +14,7 @@ import { CanvasCourseSection, canvasRoles } from '../models/canvas'
 import { addUMUsersProps } from '../models/feature'
 import { CCMComponentProps } from '../models/FeatureUIData'
 import ValidationErrorTable, { ValidationError } from '../components/ValidationErrorTable'
+import { hasUnauthorized } from '../utils/handleErrors'
 
 const USER_ROLE_TEXT = 'Role'
 const USER_ID_TEXT = 'Login ID'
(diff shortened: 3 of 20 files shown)


ccm_web/server/src/canvas/invalid.token.interceptor.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (user === undefined) throw new RequestWithoutUserError()
+
+    return next.handle().pipe(
+      catchError(async (err) => {
+        // Other unauthorized cases related to JWT or sessions will be caught by guards.
+        if (
+          err instanceof InvalidTokenRefreshError ||
+          (err instanceof HttpException && err.getStatus() === HttpStatus.UNAUTHORIZED)
+        ) {
+          await this.canvasService.deleteTokenForUser(user)
```
ssciolla, line 26: There seems to be some portion of programmers who espouse that exceptions shouldn't be used for controlling the flow of execution. I guess this could be considered an instance of that, and we could move the use of `canvasService.deleteTokenForUser` to `canvasService.refreshToken` and a handful of places in `APIService`. I was trying to keep the code DRY, but I can see the argument that the interceptor usage makes the code more difficult to read.

ccm_web/client/src/App.tsx
```diff
@@ -62,6 +64,12 @@ function App (): JSX.Element {
     )
   }
 
+  if (!globals.user.hasCanvasToken) {
+    // Initiate OAuth flow
+    location.href = '/canvas/redirectOAuth'
+    return loading
+  }
```
ssciolla, line 71: I think it might be reasonable from a UX perspective to add a banner or alert here instead that links to `/canvas/redirect`, telling the user that the Canvas API integration is missing and that they need to go through the OAuth flow. With the changes I've made, this would also be show to the user the first time they access CCM. What do people think?

ccm_web/server/src/canvas/canvas.controller.ts
```diff
@@ -32,7 +32,7 @@ export class CanvasController {
     }
 
     if (user.canvasToken !== null) {
-      logger.debug(`User ${user.loginId} already has a CanvasToken record; redirecting to home page...`)
+      logger.warn(`User ${user.loginId} already has a CanvasToken record; redirecting to home page...`)
       return res.redirect('/')
     }
```
ssciolla, line 37: Deciding to leave this for now as it's possible the user could navigate directly to `canvas/redirectOAuth` and we'd want to bail if they have a token in that case.

chrisrrowland:
Some things I have tested and the results.  Probably all working as expected.

Manually bungled refresh_token in db: No immediate problem.
Manually bungled refresh_token in db + expired token: Reauthorization required.
Manually bungled access_token in db: Reauthorization required.

When I  manually bungle the access_token, a new approved integration is created in canvas and the old one remains.
Not sure if there is a way to have the old one(s) automatically cleaned up, though this is also pretty unimportant I imagine, since we're not really worried about this scenario.

chrisrrowland:
Got an unexpected error:

1. Delete approved integration
2. Manually hit an API endpoint.  For example I got frame source and then added a suffix to go to view-source:https://rowc-ccm.ngrok.io/api/course/483458/sections
3. Try to do something again in CCM.  I tried renaming the course again and got an error.

ccm_web     | [Nest] 101   - 09/13/2021, 2:12:26 PM   [ExceptionsHandler] The user rowc does not have a CanvasToken and should; the database may have been modified or there may be a problem with the application code. +65493ms

ssciolla:
> ccm_web | [Nest] 101 - 09/13/2021, 2:12:26 PM [ExceptionsHandler] The user rowc does not have a CanvasToken and should; the database may have been modified or there may be a problem with the application code. +65493ms

Hm, what do you think the best way to handle this would be? I'm guessing if you refresh, the application sends you through OAuth? This feels like an edge case of edge cases, but we could try to handle elegantly. Otherwise, I guess I'd suggest we implement a global error message on `500` errors suggesting a refresh.

ccm_web/client/src/App.tsx
ssciolla: It sounds like @janelilr and the UX team recommend some kind of explanatory interstitial page. I think this can be a separate issue for now; I have a prototype that I will share with the team that I can open a PR for after this is approved/merged.

ssciolla:
Thanks, @chrisrrowland, for the testing! I think this is ready for review. I'm open to significant changes if someone proposes strong alternatives.

ccm_web/server/src/canvas/invalid.token.interceptor.ts
```diff
@@ -0,0 +1,42 @@
...
+  async intercept (context: ExecutionContext, next: CallHandler): Promise<Observable<unknown>> {
+    const req = context.switchToHttp().getRequest<Request>()
+    const user = req.user
+    if (user === undefined) throw new RequestWithoutUserError()
+
+    return next.handle().pipe(
+      catchError(async (err) => {
+        if (err instanceof CanvasTokenNotFoundError) {
+          throw new UnauthorizedException('The tool does not have a Canvas token for you.')
+        }
```
ssciolla, line 26: @chrisrrowland, I added this to handle the case you brought up; let me know what you think when you can.

ccm_web/client/src/App.tsx
ssciolla: See #189.

pushyamig:
I will start the review now.

pushyamig:
When I was testing the code for simple user authorization use case when making this call `redirectOAuth` I saw failure Cancellation on the GET call the second call `redirectOAuth` went through with 304. Why 2 call for single purpose? I verified this behavior is not see on CCM dev instance 
Screen Shot 2021-09-15 at 10 29 12 AM

ssciolla:
> When I was testing the code for simple user authorization use case when making this call `redirectOAuth` I saw failure Cancellation on the GET call the second call `redirectOAuth` went through with 304. Why 2 call for single purpose? I verified this behavior is not see on CCM dev instance

Thanks, I'm not sure immediately what's happening here, I'll take a look.

pushyamig:
Screen Shot 2021-09-15 at 10 48 21 AM

When I was testing case
1. I logged into CCM
2.  Revoked the token in the canvas console 
3.  From CCM made bulk create section call/course rename.

Behavior: Saw that BE is catching 401 authorized and redirecting back to canvas authorization page. But the browser console logs seems to have some failures before a successful redirect made. Why so many failures before a success? I verified this behavior from another browser as well. Something to do with Ngrok?

pushyamig:
After I logged into CCM, revoked the token and went into Add UM user case, I don't see the Redirect to authorization page is happening. The Add UM user on load fetches the section for the course. The previous cases I test were POST and PUT. This seem to be a bug.

ssciolla:
> After I logged into CCM, revoked the token and went into Add UM user feature, I don't see the Redirect to authorization page is happening. The Add UM user on load fetches the section for the course. The previous cases I test were POST and PUT. This seem to be a bug.

I haven't implemented anything on that route yet, because there are no API calls on that view. So not a bug unless, you want me to do that now?

pushyamig:
>Oh wait, you're saying it already fetches sections? I can fix that.

yeah The API call exits now and used by create sections/Add UM user feature

ccm_web/client/src/App.tsx
```diff
@@ -62,6 +64,12 @@ function App (): JSX.Element {
     )
   }
 
+  if (!globals.user.hasCanvasToken) {
+    // Initiate OAuth flow
+    location.href = '/canvas/redirectOAuth'
```
pushyamig, line 69: I did some reading with `location.href.` There are quiet a few option to do a redirect like `assign`, `replace`. This [article](https://marikalam.medium.com/how-do-you-redirect-to-another-page-in-javascript-6c7524c1f88a) suggest `replace` is best as it doesn't keep the browser history.

like this `location.replace(/canvas/redirectOAuth)`

ssciolla: Hm, interesting. I did see this, but I wasn't sure whether we should remove the history. Maybe we don't want them to be able to go back in this instance. I'll take a look in a minute.

pushyamig:
The Code workflow with revoking the authorization is good. 
I think it will good to get answers around why duplicate call will fail and finally successful. If this is something ngrok is doing, then should be configurable.

I looked with ngrok configuration `--host-header` and trying different redirect client option like `location.assign` and `location.replace` nothing seem to solve the problem of failure/success certain url `/`, 'returnOAuth'

ccm_web/client/src/components/CreateSectionWidget.tsx
```diff
@@ -53,18 +55,22 @@ function CreateSectionWidget (props: CreateSectionWidgetProps): JSX.Element {
...
-    }).catch(() => {
-      enqueueSnackbar('Error validating section name', {
-        variant: 'error'
-      })
+    }).catch((e: Error) => {
+      if (hasUnauthorized([e])) {
+        location.href = '/'
+      } else {
+        enqueueSnackbar('Error validating section name', { variant: 'error' })
+      }
```
ssciolla, line 73: My approach using `hasUnauthorized` becomes a bit clunkier when we're not using `usePromise`, because the errors are not all available in the same scope. I guess this works for now though? I didn't want to mess with what @chrisrrowland had put together too much.

ccm_web/client/src/pages/AddUMUsers.tsx
```diff
@@ -143,6 +144,9 @@ function AddUMUsers (props: AddUMUsersProps): JSX.Element {
     }
   )
 
+  const apiErrors = [getCanvasSectionDataError]
+  if (hasUnauthorized(apiErrors)) location.href = '/'
```
pushyamig, line 148: I wish we have some kind of interceptor on 401 in react as well, we are repeating these 2 line of code everywhere.

ccm_web/client/src/utils/canvasSectionNameValidator.ts
```diff
@@ -37,22 +37,9 @@ export class CanvasCoursesSectionNameValidator {
...
-    return sections
-  }
-
   public async validateSectionName (newName: string): Promise<ICanvasSectionNameInvalidError[]> {
     const errors: ICanvasSectionNameInvalidError[] = []
-    let sections: CanvasCourseSection[] = []
-    await this.getSections().then(s => {
-      sections = s
-    })
+    const sections = await getCourseSections(this.course.id)
```
ssciolla, line 42: @chrisrrowland, maybe I missed something but `getSections` didn't really seem to be doing anything that `await doTask` wouldn't do automatically. Actually, because error is passed through the `Error()` constructor, we were loosing the subclass typing, including `UnauthorizedError`, which I need for `hasUnauthorized`. Let me know if you have any thoughts.

ccm_web/client/src/pages/AddUMUsers.tsx
ssciolla: Yeah, it's not ideal. I'd like to just use this once on each page, but not all API calls are occurring at the top level (as in `CreateSectionWidget`).

I guess one alternative would be to send some way for the frontend to know it needs to redirect (like an error property), and then maybe we could do it `handleErrors`. I could try that. I'm less confident about the frontend solution, although I'm not THAT confident about any of this.

ssciolla: You may lose some readability though with moving it to a utility function, it's kinda nice to see, okay, if I'm using this component and an unauthorized error occurs, I'm going to get sent back to root.

ssciolla:
> yeah The API call exists now and used by create sections/Add UM user feature

The existing API calls in that view should be handled now.

ccm_web/client/src/pages/AddUMUsers.tsx
ssciolla: @pushyamig, does this seem better? https://github.com/ssciolla/canvas-course-manager-next/commit/452c3526787100ad2611304392965f5a722a90f4

pushyamig: Yup. Looks good! Thx for having this fix at the top level. This way we don't need to worry about  adding a redirect for every API call in future. Since these things can easy to miss.

ccm_web/client/src/utils/canvasSectionNameValidator.ts
pushyamig: Yeah.. I agree we should fix this so that all API call redirect call handled at `handleError()` block

ssciolla: I think this change isn't that relevant any more, because the redirect would be handled by the first `getCourseSections`. Still, I think I'm going to keep this change, as it'll make non-`UnauthorizedError` subclasses available to the error alerts. It's also more concise 😀

ccm_web/client/src/utils/redirect.ts
```diff
@@ -0,0 +1,5 @@
+type RedirectRoute = '/' | '/canvas/redirectOAuth'
+
+export default function redirect (route: RedirectRoute): void {
+  location.href = route
+}
```
ssciolla, line 5: This feels a bit silly to me, but it does allow us to be make changes to the `location` usage or add logging if we want to, makes things a little more readable, and developers have to register new routes beforehand. I dunno, we can remove it if someone doesn't like it.

ccm_web/client/src/App.tsx
ssciolla: I did some testing; this doesn't seem to be a huge difference. I did notice that the if the user hits the back button with `location.replace`, eventually they will get a `404`. This doesn't happen with `location.href`, it just seems to keep sending them to the application root. I can't explain that, but I think the latter is preferred, so I'm inclined to keep it as is.

ccm_web/client/src/components/CreateSectionWidget.tsx
ssciolla: No longer relevant, since I changed the implementation.

ccm_web/client/src/pages/AddUMUsers.tsx
ssciolla: Okay, change is incorporated. Seems better to me overall, much easier to keep things in sync over time. It's a little strange to me to be redirecting from this error handling abstraction layer, but other benefits outweigh that strangeness.

ssciolla:
@pushyamig, re-requesting a review, but I still need to look into the network issue. I'll deploy to OpenShift this afternoon so we can determine whether it's related to `ngrok` or not.

pushyamig (approved):
1. Changes with redirect on every API call in FE to a more centralized location is a very Huge improvement to the code. As this will eliminate any misses of adding a redirect for Future API calls
2. Tested the code on CCM-dev instance and all the failures/extra redirect call happening locally are not seen in Dev instance. This seem to be ngrok setup problem.
3. With that said, I had a private conversation with Sam on this, He mentioned that with addition of the Pre-Auth page he did not see the extra calls/redirect error Locally. If this issues solve with another PR then we should not spend more time on this and we could live with some extra redirects temporarily. 
4. DB migration are applied properly as well

Nice Job! This is important piece of integration that nears finishing upon the Canvas Oauth integration and we learned through this a need for Pre-auth page as per UX standards

ssciolla:
Thanks for the review, @pushyamig! Merging now...

PR #188, opened 2021-09-14
Handle when user hits Cancel on OAuth prompt (#187)

### LINKED ISSUE #187: Handle cleanly when user hits "Cancel" on Canvas authorize prompt
As the OAuth flow currently exists, if the user hits "Cancel" when they reach Canvas, the application errors out. Because the returned URL query has only an `error` key, it's not considered a "bad" query, and we end up making an invalid call to request a Canvas token (resulting in a `500` error on the request to `canvas/returnFromOAuth`).

Work on issue #101 is somewhat adjacent to this. We either need to show an error page, or redirect back to the application which then restarts the Canvas OAuth flow.

ssciolla:
The PR aims to resolve #187.

### DIFF
--- a/ccm_web/server/src/canvas/canvas.controller.ts
+++ b/ccm_web/server/src/canvas/canvas.controller.ts
@@ -49,13 +49,13 @@ export class CanvasController {
   ): Promise<void> {
     logger.debug(`Query: ${JSON.stringify(query, null, 2)}`)
     if (isOAuthErrorResponseQuery(query)) {
-      logger.error(`Canvas OAuth failed due to ${query.error}: ${String(query.error_description)}`)
       if (query.error === 'access_denied') {
         logger.debug('User rejected Canvas OAuth; sending them back to application root...')
         res.redirect('/')
         return
       }
 
+      logger.error(`Canvas OAuth failed due to ${query.error}: ${String(query.error_description)}`)
       const message = query.error_description !== undefined
         ? query.error_description
         : 'Canvas OAuth error was sent without a description.'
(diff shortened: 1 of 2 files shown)


ccm_web/server/src/canvas/canvas.controller.ts
```diff
@@ -47,15 +47,24 @@ export class CanvasController {
...
+        logger.debug('User rejected Canvas OAuth; sending them back to application root...')
+        res.redirect('/')
+        return
+      }
+
+      logger.error(`Canvas OAuth failed due to ${query.error}: ${String(query.error_description)}`)
+      const message = query.error_description !== undefined
+        ? query.error_description
+        : 'Canvas OAuth error was sent without a description.'
+      throw new InternalServerErrorException(message)
```
pushyamig, line 62: Does this code ever will ever hit? what are other scenarios to apart from access_denied to reach L58

ssciolla: I think you added this before in PR #130 to handle a scope error. So yes? I think even still, it's good to have a general fallback for scenarios you haven't anticipated.

pushyamig: Yeah.. I remember that. Agreed on fallback. 

But when I coded that it was sending the error message like ` Canvas OAuth failed due to invalid_scope: A requested scope is invalid, unknown, malformed, or exceeds the scope granted by the resource owner. The following scopes were requested, but not granted: url:POST|/api/v1/accounts/:account_id/account_notifications`
Not anymore. The UI just be as below
Screen Shot 2021-09-24 at 1 55 03 PM

This is a separate issue.

pushyamig (approved):
Looks Good! Works as intended to.

ccm_web/server/src/canvas/canvas.controller.ts
ssciolla: I don't remember that, but I'd have to look back to see.

PR #195, opened 2021-09-20
Add `AuthorizePrompt` component, extract `Breadcrumbs` (#189)

### LINKED ISSUE #189: Add page/content prior to Canvas OAuth redirect
I think it might be reasonable from a UX perspective to add a banner or alert here instead that links to `/canvas/redirectOAuth`, telling the user that the Canvas API integration is missing and that they need to go through the OAuth flow. With the changes I've made, this would also be show to the user the first time they access CCM. What do people think?

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/183#discussion_r705500715_

This was discussed in meetings, and the UX team is recommending some kind of intermediary/interstitial page or content that tells the user why they are being redirected. My current thinking is this will be shown both when the user joins the first time, and if they are redirected because of a removed integration (see #183).

I have a prototype in a branch (https://github.com/ssciolla/canvas-course-manager-next/commit/31ad4318ab477af4fe9099776eb55d983edb0e6b), and Janel has provided some mockups with modified language.

Questions:
- Should the page include the home breadcrumb?
- What is the exact preferred language? What are the overarching design principles here? Simplicity? Clarity? Ease of use?
- What should the button `aria-label` text say?

ssciolla:
This PR changes the OAuth redirect logic so that it shows the user an intermediary prompt with information about the missing integration and a button that when clicked takes them to the Canvas authorize page. The changes include an extraction of the `Breadcrumbs` implementation so that it could be more easily reused (as in the `AuthorizePrompt` block), styling specific to that component would be isolated, and the `App.tsx` component would be streamlined. Language used has been requested and cleared by Janel, but may be adjusted at a later date. The PR aims to resolve #189.

### DIFF
--- a/ccm_web/client/src/components/AuthorizePrompt.tsx
+++ b/ccm_web/client/src/components/AuthorizePrompt.tsx
@@ -15,7 +15,9 @@ export default function AuthorizePrompt (): JSX.Element {
   const classes = useStyles()
   return (
     <div className={classes.root}>
-      <Typography variant='h5' component='h1' gutterBottom>Need to Authorize Canvas Integration</Typography>
+      <Typography variant='h5' component='h1' gutterBottom>
+        Authorize Course Manager to Access Your Canvas Account
+      </Typography>
       <Typography gutterBottom>
         This tool is not integrated with Canvas.
       </Typography>
--- a/ccm_web/client/src/components/AuthorizePrompt.tsx
+++ b/ccm_web/client/src/components/AuthorizePrompt.tsx
@@ -1,0 +1,37 @@
+import React from 'react'
+import { Button, makeStyles, Typography } from '@material-ui/core'
+
+const useStyles = makeStyles(() => ({
+  root: {
+    padding: 25,
+    textAlign: 'left'
+  },
+  button: {
+    marginTop: 15
+  }
+}))
+
+export default function AuthorizePrompt (): JSX.Element {
+  const classes = useStyles()
+  return (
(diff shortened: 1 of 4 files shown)


ccm_web/client/src/components/AuthorizePrompt.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <b>Note:</b> If you have done this before, it is likely that your integration was removed in Canvas.
+      </Typography>
+      <Button
+        className={classes.button}
+        color='primary'
+        variant='contained'
+        href='/canvas/redirectOAuth'
+        aria-label='Go to authorize page'
+      >
+        GO TO AUTHORIZE PAGE
```
pushyamig, line 35: I don't see the failure of `redirectOAuth` error anymore Locally with ngrok. Probably this is wrapped up in Button component and that is some Material-UI is handling it in better way. Nice Improvement.

pushyamig (approved):
This is functional and working the way it supposed to be. The code Refactor around Breadcrumbs I don't see any concerns. But I would like @chrisrrowland input on it if he has time otherwise this looks good.

chrisrrowland:
I don't see anything concerning with the breadcrumbs refactor.  I didn't test it but I imagine any breaking  issues would have been identified pretty easily.

ssciolla:
Thanks for the feedback! I'm going to merge this.

Issue #113, opened 2021-06-10
handle simultaneous use in multiple courses


lsloan:
## Problem

When CCM is launched, the ID of the course from which it was launched is stored in the session.  Subsequent launches from other courses within the same browser (e.g., the user launches from another tab or window) will cause the stored course ID to be overwritten with the ID of the last course to launch the app.  Using the tab/window with the earlier launched app may make it refer to the later launching course.

## Possible solutions

1. Documentation should warn users not to use CCM simultaneously from multiple tabs/windows within a browser.
2. Use some browser window-specific cookie/data to keep track of which course the app is being used with.  This may not be supported the same way across all browsers.
3. Perhaps the frontend can keep track of which course from which it was launched (e.g., in React globals) and always pass the course ID along with API requests.  API entries should require the course ID as part of each request (e.g., "`/api/hello/𝒏𝒏𝒏𝒏𝒏𝒏`", "`/api/hello/course/𝒏𝒏𝒏𝒏𝒏𝒏`", or in the request body).
4. When launch requests are received, if the user has a "recent" session open with a different course number, warn the user.  Either tell them that their session with the other course will be terminated or prompt them whether they want to continue their session with a new course before overwriting the course ID.

Solutions that try to support simultaneous use (items 2 and 3) will probably require restructuring of the data in the session.  That is, the session could contain a list of courses associated with it.  Each of the courses in the list would contain data about that course, like the user's role within the course (if they're not a Canvas admin), etc.

lsloan:
I didn't know whether "bug" was an appropriate label for this issue.  Feel free to change it.  Also feel free to change the problem description and possible solutions.

pushyamig:
Need testing to see if this is a real issue or not

pushyamig:
@melindakraft Can you test this use case to see if this is a real issue or not?

melindakraft:
In testing as KalturaK in multiple course tabs at the same time I did not encounter any problems. The user was able to perform the tasks appropriate to the role in the course and those tasks did not generate errors. I am going to assume this is not a real issue until we receive reports from our users.

Issue #16, opened 2021-02-23
Feature: Merging Section for instructors and Admins


pushyamig:
This Feature takes care of cross listing and de-cross listing of section in a term. This feature has slightly different variation for service center/Subaccount/Account admin (who are admin in canvas),  and instructors. This feature is also available to subaccount admins

```
Cross-listing = Merging section
De cross-listing = Un merging section 
```
**Instructors**: The wireframes are located [here](https://github.com/tl-its-umich-edu/UX-Work/blob/master/ccm/wireframes/merge_feature/merge_feature_instructor.pdf)
1. Show his the list of sections in the course and list of sections from other courses of role type of instructor same term. The sections belonging to the course( launched from ) shouldn't be choosable 
2. He can chose which sections from other courses to cross list, follow the steps as in the wireframes. 
3. Decrosslisting is only added very end in case if he made a wrong choice for merging

**Admin**: The wireframes are located [here](https://github.com/tl-its-umich-edu/UX-Work/blob/master/ccm/wireframes/merge_feature/merge_feature_sc.pdf)
**Role**: Account Admin, Support Consultant, Subaccount Admin
1. Show him the sections that belongs to the course (right side).
2. Upon search based on course name ( from same term) show the list of sections belonging to that course
3. Upon search based on uniqname of instructors ( from same term) show the list of sections belonging to him
3. He can chose which sections from other courses to cross list, follow the steps as in the wireframes. 
4. Decrosslisting is only added very end in case if he made a wrong choice for merging

**Recommended Api calls for this Features to use:**
1. Get Sections in a course by instructor Uniqname: Based on admin access to which subaccount/root the :Account_Id is substituted. https://umich.beta.instructure.com/api/v1/accounts/:Account_id/courses?by_teachers[]=sis_login_id:instx&enrollment_term_id=284&per_page=100. This should fetch all the courses the instructor with active, invited_or_pending, completed
2. Get courses for current user:  https://umich.beta.instructure.com/api/v1/courses?enrollment_type=teacher. This could be used for instructor feature
2. Get Section in a course: https://umich.test.instructure.com/api/v1/courses/423008?include[]=sections&include[]=total_students
3. Get Sections in a course by course name: Based on admin access to which subaccount/root the :Account_Id is substituted:  https://umich.test.instructure.com/api/v1/accounts/:Account_id/courses?search_term=PHARMSCI 101 001 WN 2021
[...]

PR #199, opened 2021-09-22
Add merge sections backend functionality, refactor handlers (#193, #198)

### LINKED ISSUE #193: refactor API `enrollUsers()` method to accept list of `sectionUsers`
Minor API design question: should `enrollUsers` accept the list of `sectionUsers`? Then it could be re-used more effectively. I guess at that point, it might be more of a `SectionHandler` object that could do other section-related operations. Maybe these are niceties that are not necessary at this point, but it could get us closer to a `cavnvasapi`-like interface.

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/165#r702143192_

---

It's a better design than the one used for the creation section support code, which was the pattern for the bulk enrollment support code.  It would allow for the handler's methods to be used again within the same invocation without calling the constructor again.

### LINKED ISSUE #198: Create backend API endpoints, Canvas request processes for Merge Sections feature
This issue represents the backend work required to make the Merge Sections feature work. We will need slightly different mechanisms for instructors and admins to get sections:

Getting sections
- for instructors, simply get the sections they teach in a given term (probably `/courses/` endpoint)
- for admins, get courses in accounts they administer in a given term, with optional filtering by instructor and course name
- Course relationship should somehow be traverse-able, maybe by nesting the sections with their course

Merge/un-merge sections
- This should be more straightforward; section IDs will be needed

This work will almost certainly involve new scopes and refactoring in order  re-use existing processes.

Parent issue is #16, and sibling issue is #182.

ssciolla:
This PR introduces the necessary server routes and Canvas API calls to meet the requirements of the Merge Sections feature, including section fetching for instructors and admins (including filters for the latter). It also contains a significant refactor of separate API handler classes, started by @pushyamig and @lsloan, to be organized around Canvas entities and Canvas URL stems. The PR aims to resolve #193 and #198.

To Do
- [x] Add course section getting routes for instructor and admins and associated Canvas calls
- [x] Refactor `CreateSectionsApiHandler` as `CourseApiHandler`
- [x] Refactor `EnrollSectionUsersApiHandler` as `SectionApiHandler`
- [x] Add merge and un-merge routes to controller
- [x] Add service methods for above, or create SectionHandler
- [x] Re-name `privilegeLevelOneScopes` to `canvasScopes` (seems like everyone can access all routes, they just may not get data back)

Future Work?
- [ ] Consider limiting access to `/admin/sections/` by role from LTI launch (probably a separate issue)


ccm_web/server/src/api/api.controller.ts
```diff
@@ -77,4 +103,26 @@ export class APIController {
...
+      @UserDec() user: User
+  ): Promise<CanvasCourseSectionBase> {
+    const result = await this.apiService.mergeSection(user, sectionId, courseId)
+    if (isAPIErrorData(result)) throw new HttpException(result, result.statusCode)
+    return result
+  }
+
+  @UseInterceptors(InvalidTokenInterceptor)
+  @ApiSecurity('CSRF-Token')
+  @Delete('sections/:id/unmerge')
```
chrisrrowland, line 122: Pretty sure there is no unmerge requirement for CCM, was this  just out of completeness?

ssciolla: It's listed in issue #16, and I see it in the admin version wireframes ("unmerge" button on success screen?).

ccm_web/server/src/api/api.controller.ts
```diff
@@ -77,4 +103,26 @@ export class APIController {
     if (isAPIErrorData(result)) throw new HttpException(result, result.statusCode)
     return result
   }
+
+  @UseInterceptors(InvalidTokenInterceptor)
+  @ApiSecurity('CSRF-Token')
+  @Post('sections/:id/merge/:course')
+  async mergeSection (
```
chrisrrowland, line 110: What do you think about 
/course/:id/mergesections
with a body of section ids?

That would allow multiple sections to be merged into the target course in a single call.

Its possible I still don't have my  head wrapped around what is  actually going on though.

ccm_web/server/src/api/api.controller.ts
chrisrrowland: You're right.. I was remembering talking about no umerging in a zoom call but now I can't remember the whole context.  Carry on :)

ccm_web/server/src/api/api.controller.ts
ssciolla: Yeah, I could do this, seems like it would simplify things for you. Does it make sense for unmerge to be bulk too?

ssciolla: I have to think more about how unmerge bulk would look (like the route and parameters, body, etc.). You don't need the `courseId` for this.

chrisrrowland: > Yeah, I could do this, seems like it would simplify things for you. Does it make sense for unmerge to be bulk too?

In the UI it looks like unmerge is intended to be section by section, but would it be just the same on the backend if it accepted an array?

ssciolla: Yeah, I guess there is no reason we can't do this in a bulk way, but in practice, do we really need that functionality yet?

ssciolla: @chrisrrowland, okay, are you happier with these changes? https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/199/commits/0c009a3859045bd02cafd1ef114fca6263df892b and https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/199/commits/fa17d17b2013d892f7a0f29c8299719b31cfa55e

ccm_web/server/src/api/api.controller.ts
```diff
@@ -68,6 +69,32 @@ export class APIController {
...
+  }
+
+  @UseInterceptors(InvalidTokenInterceptor)
+  @ApiQuery({ name: 'term_id', type: Number })
+  @ApiQuery({ name: 'instructor', required: false, type: String })
+  @ApiQuery({ name: 'course_name', required: false, type: String })
+  @Get('admin/sections')
+  async getCourseSectionsInTermAsAdmin (
+    @UserDec() user: User,
+      @Query('term_id', ParseIntPipe) termId: number,
```
pushyamig, line 89: I don't think the UI know which terms this course belongs to. Term info mostly come into play Merge section feature. Either you have to send that info to FE through globals or store it into session or both.

ssciolla: I'm sending it through the course endpoint (see changes to `/course/:id` endpoint and `CanvasCourseBase`, which the frontend gets on load and can pass to the feature. It seemed better to me to get it that way, but I could do through the launch/session.

ssciolla: I *think* doing it this way may circumvent some of the issues related to #113, because the React application should still have in local state which course was sent in `globals` after the launch, and then be able to pull the latest term info even if session data changed.

pushyamig: That would be yet another Prop along with global props. I see your point since the course API call is made anyways after the global call and the term data is present there. If you go through Globals I think it would be making API call to get the course term via course API endpoint and then repeat. 

What you did make sense!

ccm_web/server/src/api/api.utils.ts
```diff
@@ -38,7 +39,7 @@ export function parseErrorBody (body: unknown): string {
   return body.errors.map(e => e.message).join(' ')
 }
 
-export function makeResponse<T extends CanvasEntity> (multipleResults: Array<APIErrorData | T>): T[] | APIErrorData {
```
ssciolla, line 41: I think I suggested this initially, but I was seeing some TypeScript errors when I tried to use it. Because the `T` isn't actually used anywhere, I don't think it's essential that we use `extends` to limit what types are accepted here. Basically, it will be `T` if it's not `APIErrorData`. One cool thing is that by removing this, we can actually use this for calls that return multiple objects, since we can pass, for example, `CanvasCourse[]` as the type parameter, and it all still works!

ssciolla: Besides, I am now seeing an error where this function returns `successes` that looks like this:
```
Type 'CanvasEnrollment' is not assignable to type 'T'.
          'CanvasEnrollment' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'CanvasEntity'.ts
```
This sent me down a rabbit hole of SO posts and TypeScript docs, and I'm not entirely sure I can explain it. Perhaps TS can't ensure that the same member of the `CanvasEntity` union type is accepted and returned from the function? The error goes away when you remove the `CanvasEntity` constraint.

chrisrrowland:
I started trying to integrate this with my front end PR and noted a couple of items which have been addressed.

The change to CourseWithSections requires some moderate refactoring of front end code so I haven't been able to actually test yet, however eyeballing the code it looks ok to me and I'd perfer just to merge and figure out if it needs more changes later.

ccm_web/server/src/api/dtos/api.section.ids.dto.ts
```diff
@@ -0,0 +1,16 @@
...
+  @ArrayMaxSize(250)
+  @ArrayUnique()
+  @IsInt({ each: true })
+  @IsNotEmpty({ each: true })
+  sectionIds: number[]
+
+  constructor (sectionIds: number[]) {
+    this.sectionIds = sectionIds
+  }
+}
```
ssciolla, line 16: Not sure about the `250` limit here, but it seems responsible to put --some-- limit on this.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -34,15 +34,20 @@ export interface TokenRefreshResponseBody extends TokenBaseResponseBody {}
...
-export interface CanvasCourseSection {
+export interface CanvasCourseSectionBase {
   id: number
   name: string
+  course_id: number
+}
+
+export interface CanvasCourseSection extends CanvasCourseSectionBase {
   total_students: number
 }
```
ssciolla, line 52: I think we need a `Base` because operations like `merge` and `unmerge` don't have access to `total_students`. Will double check that.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -62,8 +67,20 @@ export interface CanvasEnrollment {
...
+  parent_account_id: number | null
+}
+
 export type CanvasEntity = CanvasEnrollment | CanvasCourseSection | CanvasCourse
 
+// Composites
+
+export interface CourseWithSections extends CanvasCourseBase {
+  sections: CanvasCourseSection[]
+}
```
ssciolla, line 82: It was a deliberate choice not to use the `Canvas` prefix here. I thought maybe composite, non-atomic entities might be easier to identify if we didn't use that. Open to changing it though.

ssciolla:
I think this is ready for review. I'm going to look at this Codacy alert again, and also do some more testing. @lsloan, I think @pushyamig and @chrisrrowland can probably handle this, since they've already started looking at it, but I've incorporated some of the refactoring stuff we discussed on PR #165 and there are changes I want you to be aware of. Of course, you can review if you would like to.

chrisrrowland (approved):
Approving so we can merge and I can more effectively review.

Been trying to get the FE and BE to align to actually test but it's been quite a bit of changes and I'd rather just figure out if something needs to change after merging.

ccm_web/server/src/api/api.service.ts
```diff
@@ -37,66 +41,92 @@ export class APIService {
...
+    const courseHandler = new CourseApiHandler(requestor, targetCourseId)
+    return await courseHandler.mergeSections(sectionIds)
+  }
+
+  async unmergeSections (user: User, sectionIds: number[]): Promise<CanvasCourseSectionBase[] | APIErrorData> {
+    const requestor = await this.canvasService.createRequestorForUser(user, '/api/v1/')
+    const NS_PER_SEC = BigInt(1e9)
+    const start = process.hrtime.bigint()
+    const apiPromises = sectionIds.map(async (si) => {
+      const sectionHandler = new SectionApiHandler(requestor, si)
```
pushyamig, line 123: What is the thinking behind putting the merge/unmerge sections in separate handlers course/section. But in the course handler you are calling the section handler for unmerging it. I felt for the API with course_id dependency seems to be in course handler and things with sections dependency seems to be section. 

I think for future API call, If you document the intent that it would be easier for follow along the similar pattern

ccm_web/server/src/api/api.service.ts
```diff
@@ -37,66 +41,92 @@ export class APIService {
...
+    const apiPromises = coursesInTerm.map(async c => {
+      const courseHandler = new CourseApiHandler(requestor, c.id)
+      return await courseHandler.getSectionsWithCourse(CourseApiHandler.slimCourse(c))
+    })
+    const coursesWithSectionsResult = await Promise.all(apiPromises)
+    return makeResponse<CourseWithSections>(coursesWithSectionsResult)
   }
 
-  async getCourseName (user: User, courseId: number): Promise<CanvasCourseBase | APIErrorData> {
+  async getCourseSectionsInTermAsAdmin (
```
pushyamig, line 73: @ssciolla I think the sections from the API `/api/v1/accounts/:account_id/courses` from the API will pull all state like `created, claimed, available, completed` . We should document this here just for  clarity in code or swagger API( leave it if adds more work) 

I feel in the testing Melinda can decide which list to populate. Courses in canvas are create 6 month before the term. Who know when an instructor will decide to merge section. I don't think any action is needed here now but sharing some insights

ccm_web/server/src/api/dtos/api.section.ids.dto.ts
pushyamig: The use case don't call for bulk create that many section at-least based on todays Canvas usage by Admin staff . 60 is the limit we set and the Backend handles it. But 250 limit seem ok and code can handle it with async calls

ccm_web/server/src/canvas/canvas.interfaces.ts
pushyamig: Yeah... it is not returned as part of merge/unmerge sections. 

I think the course/section object in the FE is following the same object pattern as Backend. May be that should reflect this pattern? https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/main/ccm_web/client/src/models/canvas.ts#L6

ccm_web/server/src/api/api.utils.ts
pushyamig: I appreciate the effort here. We can leave as is now.

ccm_web/server/src/api/api.course.handler.ts
```diff
@@ -0,0 +1,126 @@
...
+    const start = process.hrtime()
+    const apiPromises = sections.map(async (section) => await this.createSection(section))
+    const sectionsOrErrorDataObjs = await Promise.all(apiPromises)
+    // https://codezup.com/measure-execution-time-javascript-node-js/
+    const stop = process.hrtime(start)
+    logger.debug(`Time taken to create ${sections.length} sections: ${(stop[0] * 1e9 + stop[1]) / 1e9} seconds`)
+    return makeResponse<CanvasCourseSection>(sectionsOrErrorDataObjs)
+  }
+
+  async mergeSections (sectionIds: number[]): Promise<CanvasCourseSectionBase[] | APIErrorData > {
```
pushyamig, line 113: @ssciolla Technically you can cross-list across terms. Like Summer 2021 to Spring 2020 etc. 
I tested the use-case from the PR it allows to cross-list and decross list multiple terms from Swagger-UI

I know our workflow only allows to presents sections from same terms. Should be allow more checks if the both course and section belongs to same term?

But this shouldn't be show stopper to merge this. This just needs more thoughts in the BE and user can't so this from the UI

ccm_web/server/src/api/api.service.ts
ssciolla: This is a good question. Let me try to explain why I did it this way, and then maybe you can give some advice on how to document it for future use.

I put `mergeSection` and `unmergeSection` in the `SectionHandler` because the routes in Canvas are section-oriented, meaning they use `sections/x` as a base, like the other section calls (`sections/${this.sectionId}/crosslist/${targetCourseId}` and `sections/${this.sectionId}/crosslist`).

In the code you commented on, I chose to only use `SectionApiHandler`s for `unmergeSections` because there isn't a need to identify a course in that operations (see above route). For `mergeSections`, I could have done something similar and not used `CourseApiHandler` at all, but I thought it might be convenient to have a method on that class that would merge multiple sections to one course. I used `SectionApiHandler` in that method since the operation for merging was already defined.

Hopefully that helps explain my logic? If you'd prefer, I can move the `CourseApiHandler.mergeSections` code to `APIService.mergeSections`, I'm not super attached to one implementation over another.

pushyamig:
>seems like everyone can access all routes, they just may not get data back/Consider limiting access to /admin/sections/ by role from LTI launch (probably a separate issue)

You are bringing some interesting insights that we never focused before. Yeah right now FE is the only place we check for the role and prevent feature access. With scoped token they won't see any data that they don't access to it. 

We can have a discuss if this is important for first phase, since we are guarded from incorrect access of data, But it's more making BE check explicitly  again.

ccm_web/server/src/api/api.service.ts
ssciolla: Okay, would you like me to add a comment about the states returned by default? I can see maybe removing `completed` from that list, otherwise, I think the others are probably okay. Maybe we should ask Melinda?

ccm_web/server/src/api/api.course.handler.ts
ssciolla: We could limit that, although it would probably involve more API calls to check the course and section.

I guess my thinking at this point is that if the user has modify permissions for the course and section (which scoped permissions will enforce), there's not technically any harm to keeping this possible. They would be able to do this through the API if they minted their own token and hit the Canvas API directly. Besides, I think it would be non-trivial for a user to hit our API directly, they'd have to find the API routes and the CSRF token.

ccm_web/server/src/canvas/canvas.interfaces.ts
ssciolla: Yes, good point, I can change that. Hopefully it doesn't mess with @chrisrrowland too much.

ccm_web/server/src/api/api.service.ts
pushyamig: @melindakraft What are your thoughts on this?

ccm_web/server/src/api/dtos/api.section.ids.dto.ts
ssciolla: This would be for merging and un-merging sections. I think I will leave it as is for now, but we can modify it later as needed.

There may be some decorators I'm using here that we should add to the `CreationSectionsDto`, like the `ArrayUnique` and `ArrayMaxSize`. Separate issue though.

ccm_web/server/src/api/api.course.handler.ts
pushyamig: OK, We can leave it for now and document it

ccm_web/server/src/api/api.service.ts
ssciolla: Maybe `completed` isn't really a problem since we're limiting by term. While I don't think people would really want to cross-list for past terms (where you'd have `completed` courses), I could see someone doing that for testing. In general, you wouldn't see too many `completed` courses in the current or future terms, so it wouldn't be confusing overly confusing. Also, I could imagine an instructor "complete"ing a course accidentally (instead of keeping it un-published) and still want to crosslist sections from it.

ccm_web/server/src/api/api.service.ts
pushyamig: > I can move the CourseApiHandler.mergeSections code to APIService.mergeSections

Don't need to move anything. Document it saying that based on the canvas API routing respective API call are placed in respective handler, that's should be good enough for understanding.

ccm_web/server/src/api/dtos/api.section.ids.dto.ts
pushyamig: >This would be for merging and un-merging sections. I think I will leave it as is for now, but we can modify it later as needed.

Sorry I missed noticing that. In that case LIMIT is OK.

May be this dto can have a name like `api.merge.unmerge.section.id's.dto.ts`?

ccm_web/server/src/api/api.service.ts
pushyamig: Yes. I agree on the accidentally marking Conclude the course. I don't know conclude or complete same thing. But We can show the completed course. No harm. IMO

ccm_web/server/src/api/dtos/api.section.ids.dto.ts
ssciolla: I prefer to leave it as general purpose/general naming, since there's nothing about it that is limited to those use cases. If you want me to add a comment indicating where it is used, I can do that.

ccm_web/server/src/canvas/canvas.interfaces.ts
ssciolla: I added a bit more typing to the client `models/canvas.ts` (see https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/199/commits/a014d63d6431cd0621d27e802816c06fb7a44a3e) to reflect the server changes.

I'm going to leave the `api.ts` definitions to Chris, since he's already done some of that work and may have opinions on what parameters they accept, how they are named, but @chrisrrowland, please note the nuance that the merge and unmerge routes do not return `total_students`, so you may need to hold the original section objects in state and update the `course_id` (or do some other client magic).

ccm_web/server/src/api/api.course.handler.ts
pushyamig: I think we ever open the Swagger API to Prod environment for Functional team, then I think we need to think back on all these things.

pushyamig (approved):
This Looks Good. There some documentation requested, other that the PR flow for Merge section feature is good. I tested all the case around it.

The refactoring also looks ok, done to fit the new API calls for Merge section.  I did not tested the API endpoint with refactored code, but should work as these are just code structuring and moved the methods based on that.

The termId is sent to FE via course API and FE should wire that by passing to respective components @chrisrrowland

ccm_web/server/src/api/api.service.ts
ssciolla: Okay, @pushyamig, see comment for requested documentation: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/199/commits/774571d19053c4057a3bfb1ba9ede1e4c7bcc97f

ccm_web/server/src/api/api.course.handler.ts
ssciolla: I'm not sure we'll ever want to do that, but we can discuss.

pushyamig: I am not saying opening the swagger UI for this first phase. I don't want to give the access to even dev other than local development. 

But there was some discussion around this early on among dev making it available in Prod for future. So I have to mention this now from that perspective.

ccm_web/server/src/api/api.service.ts
pushyamig: Thanks for the comments

melindakraft: @ssciolla @pushyamig For what it is worth, when a course is concluded, all enrollments are archived, including the instructor's access (technically, they are still there but in an inactive state). According to Canvas, I think we can determine that completed = concluded: https://community.canvaslms.com/t5/Instructor-Guide/What-are-the-different-states-of-a-Canvas-course/ta-p/1184

I don't think we should show concluded/completed courses for the purpose of section merging. If a course is accidentally concluded, there will be no section or enrollments. it is easy for an admin in Canvas to unconclude a course, though that won't put the enrollments back. That step has to be done separately. I would not expect CCM to fill that gap since it is an edge case. 

Regarding crosslisting across terms, it comes up only occasionally and can be handled outside of CCM. It is an edge case, too.

ccm_web/server/src/api/api.course.handler.ts
ssciolla: Gotcha. Was there something else I needed to do here documentation wise?

pushyamig: I wanted to document, that BE allows for cross-listing through swagger UI. But the FE flow disallows for user from UI

ccm_web/server/src/api/api.service.ts
ssciolla: Cool! There is more we could do with the handler refactoring. We could probably add an intermediary handler class for accounts (`AccountApiHandler`) and a non-admin handler (instructor/student), but I didn't see a huge benefit to those (at least not yet). We could also add abstract base classes, do other OOPsy stuff, but we can explore all this later if we wanted to build a more robust API. Also, GraphQL could change this significantly, whenever that rolls around. I will open a separate issue to capture the ideas here, but it's definitely Icebox material.

ccm_web/server/src/api/api.course.handler.ts
ssciolla: K, not sure where this should be documented, in code or not. I'll leave it as for now.

ccm_web/server/src/api/api.service.ts
ssciolla: Okay, let me add the state parameter then for everything but `deleted` and `completed`, and test before merging. Thanks for the input!

ssciolla: Okay, see https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/199/commits/3c03dc6314c08fd7ef15d3b2bb7b6070c809a63f

ssciolla:
Okay, I'm merging now, thanks for the reviews! If you see any bugs, please point them out to me, or create an issue and assign me.

ccm_web/server/src/api/api.service.ts
pushyamig: The new code make sense, did not test though

PR #200, opened 2021-09-23
Merge sections

chrisrrowland:
Draft of Merge Sections UI #182 

I expect there to be some issues, this turned out more complicated than I thought.

Actual section lookup is mocked up as the backend was not available while I was working on this.

Instructor view  should initialize with a view of sections (all sections the  user has a teacher role)

If they are a sub account admin or an account admin, they will be able to search by uniqname and course name, and the list will not preload anything.  (Both of these are currently just searching against the section name itself for re: BE)

Searches are auto initiated after a .5 second delay.

Clearing the search reverts to the original state.

MergeSections.tsx has some temporary code to allow you to simulate different user roles.  I actually attempted to make a utility for this but it required readable properties in globals so I went for this hack instead.

```
// TODO for dev testing remove all this before merging
// Can set to a specific role for testing purposes
const OVERRIDE_ROLE: RoleEnum | undefined = undefined // RoleEnum.Teacher
```

Course name should show in the list along with section  name and student count, but it is currently not returned by the backend.

### DIFF
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -117,7 +117,7 @@ function App (): JSX.Element {
               <div className={classes.swaggerLink}>
                 <Link href={`/swagger?csrfToken=${String(getCSRFToken())}`}>Swagger UI</Link>
               </div>
-              <div style={{ position: 'fixed', right: '25px', top: '25px' }}>
+              <div style={{ position: 'fixed', right: '25px', top: '25px', zIndex: 999 }}>
                 <ResponsiveHelper/>
               </div>
             </div>
--- a/ccm_web/client/src/App.tsx
+++ b/ccm_web/client/src/App.tsx
@@ -11,6 +11,7 @@ import { CanvasCourseBase } from './models/canvas'
 import allFeatures from './models/FeatureUIData'
 import Home from './pages/Home'
 import './App.css'
+import ResponsiveHelper from './components/ResponsiveHelper'
 
 const useStyles = makeStyles((theme) => ({
   breadcrumbs: {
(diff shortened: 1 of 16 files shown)


pushyamig:
@chrisrrowland, We have the Backend End for merge sections still as a [draft](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/199), there are endpoints created for fetching data for all the roles and use-case, if you want to try it.

chrisrrowland:
> @chrisrrowland, We have the Backend End for merge sections still as a [draft](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/199), there are endpoints created for fetching data for all the roles and use-case, if you want to try it.

Yeah, I am aware.  I wanted to test it as part of my review of that PR, however it's a somewhat fundamental change in the data structure used, which makes it a decent amount of work.

pushyamig:
I will review this sometime in the morning

chrisrrowland:
Nobody's banging on my door, but I am supposed to go back to IQ so any time spent reviewing would be appreciated :)

chrisrrowland:
> Screen Shot 2021-10-11 at 10 39 27 AM
> 
> It seems to be 2 calls are happening on load of the merge sections **FOR INSTRUCTOR**. the `section` seems to be happening first which get the course section and `sections?term_id=284` next which get sections from current course and other instructor courses. Both seems to targeted to fetching sections
> 
> 1. Why 2 call? since `sections?term_id=284` has all section from `section` call. Looks redundant. since course id is sent as part of the API call  for `sections?term_id=284`, UI can easily differentiate between current course sections vs other course sections
> 2. The `sections?term_id=284` is take some time to load all the section, I feel having a spinner would benefit.

There are 2 calls because there are 2 lists.  The "prepared to merge" side is doing the sections call, and the "sections I teach" side is doing the sections with the term call. 

I must have broken the spinner on a refactor at some point, I will fix that.

pushyamig:
>I also noticed when you are searching by course name the spinner was not working. same problem
Are you sure? It works for me

OK, I am convinced that it work. During some weird case I was found that. I don't see it any more.

pushyamig:
@chrisrrowland when your are preparing for merge can we have a counter how many selected? Since I can't tell when the selected section has unmerge button and it is not greyed out. This might be the Canvas bug with cross  listing
Screen Shot 2021-10-11 at 11 53 14 AM
So the CHEM course has unmerge button due to the canvas bug?

pushyamig:
@chrisrrowland I feel these 2 things needs to be address as things look broken. Summing up here so that you don't get lost in  pool of comment 
1. https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/200#issuecomment-940227331
2. https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/200#issuecomment-940116899

PR #202, opened 2021-09-24
use got default retry statusCodes (iss. #194)

### LINKED ISSUE #194: remove hard-coded got retry `statusCodes`
got@11.8.2 doesn't appear to have any `statusCodes` defined by default, which is probably why its retry feature didn't work as I expected until after I added `403`.  I'm hard-coding the list with the codes from got@12 and adding `403`.  In the future, we can replace the hard-coded list by replacing it with something like `defaultInternals.retry.statusCodes.concat(403)`.

_Originally posted by @lsloan in https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/165#discussion_r710448639_

---

As @ssciolla pointed out, got@11.8.2 **_DOES_** have a default list of status codes for retry.  I'm not sure why I didn't see this when I searched its GH repo.  It turns out my IDE is probably "helpfully" ignoring JS files when there are TS files with similar names, so that's why I didn't see it when I used that to search the installed module.

Lesson learned:  grep never lies.

lsloan:
Resolves #194.

### DIFF
--- a/ccm_web/server/src/canvas/canvas.service.ts
+++ b/ccm_web/server/src/canvas/canvas.service.ts
@@ -3,7 +3,7 @@ import CanvasRequestor from '@kth/canvas-api'
 import { HttpService, HttpStatus, Injectable } from '@nestjs/common'
 import { ConfigService } from '@nestjs/config'
 import { InjectModel } from '@nestjs/sequelize'
-import { Options as GotOptions } from 'got'
+import got, { Options as GotOptions } from 'got'
 
 import { CanvasOAuthAPIError, CanvasTokenNotFoundError, InvalidTokenRefreshError } from './canvas.errors'
 import { TokenCodeResponseBody, TokenRefreshResponseBody } from './canvas.interfaces'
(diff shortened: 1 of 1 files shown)


ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -29,9 +29,7 @@ const requestorOptions: GotOptions = {
   retry: {
     limit: 2,
     methods: ['POST', 'GET', 'PUT', 'DELETE'],
-    // TODO: After got@12 upgrade, replace following list with something like…
-    // got.defaultInternals.retry.statusCodes.concat(403)
-    statusCodes: [403, 408, 413, 429, 500, 502, 503, 504, 521, 522, 524]
+    statusCodes: got.defaults.options.retry.statusCodes.concat([403])
```
pushyamig, line 32: This Looks Good. As the way it is implemented is it is concatenating to existing list.

pushyamig (approved):

ccm_web/server/src/canvas/canvas.service.ts
lsloan: Yes, that's a literal reading of the code.

PR #205, opened 2021-09-28
Add additional validation to `GET` `admin/sections` route

### LINKED ISSUE #204: Prevent enormous search queries through `admin/sections`
As discussed during the team meeting, the backend can sometimes hit the Canvas rate limit when searching for courses/sections in accounts when no or minimal search inputs are provided. The following should make this more difficult to do:

1) We should require either `courseName` or `instructor`, but still allow both.
2) We should constrain the length of these inputs to >= 3 and <= 256.

ssciolla:
The PR aims to resolve #204.

### DIFF
--- a/ccm_web/server/src/api/api.controller.ts
+++ b/ccm_web/server/src/api/api.controller.ts
@@ -93,17 +93,19 @@ export class APIController {
 
   @UseInterceptors(InvalidTokenInterceptor)
   @ApiQuery({ name: 'term_id', type: Number })
-  @ApiQuery({ name: 'instructor', required: false, type: String })
+  @ApiQuery({ name: 'instructor_name', required: false, type: String })
   @ApiQuery({ name: 'course_name', required: false, type: String })
   @Get('admin/sections')
   async getCourseSectionsInTermAsAdmin (
     @Query() query: GetSectionsAdminQueryDto,
       @UserDec() user: User
   ): Promise<CourseWithSections[]> {
-    if (query.instructor === undefined && query.course_name === undefined) {
+    if (query.instructor_name === undefined && query.course_name === undefined) {
       throw new BadRequestException('You must specify either instructor or course_name as a URL parameter.')
     }
-    const result = await this.apiService.getCourseSectionsInTermAsAdmin(user, query.term_id, query.instructor, query.course_name)
+    const result = await this.apiService.getCourseSectionsInTermAsAdmin(
+      user, query.term_id, query.instructor_name, query.course_name
+    )
     if (isAPIErrorData(result)) throw new HttpException(result, result.statusCode)
     return result
   }
--- a/ccm_web/server/src/api/api.controller.ts
+++ b/ccm_web/server/src/api/api.controller.ts
@@ -1,6 +1,7 @@
 import { SessionData } from 'express-session'
 import {
-  Body, Controller, Delete, Get, HttpException, Param, ParseIntPipe, Post, Put, Query, Session, UseGuards, UseInterceptors
+  BadRequestException, Body, Controller, Delete, Get, HttpException, Param, ParseIntPipe,
+  Post, Put, Query, Session, UseGuards, UseInterceptors
 } from '@nestjs/common'
 import { ApiQuery, ApiSecurity } from '@nestjs/swagger'
 
@@ -8,6 +9,7 @@ import { Globals, isAPIErrorData } from './api.interfaces'
 import { APIService } from './api.service'
 import { CourseNameDto } from './dtos/api.course.name.dto'
 import { CreateSectionsDto } from './dtos/api.create.sections.dto'
+import { GetSectionsAdminQueryDto } from './dtos/api.get.sections.admin.dto'
 import { SectionIdsDto } from './dtos/api.section.ids.dto'
 import { SectionUserDto, SectionUsersDto } from './dtos/api.section.users.dto'
 import { JwtAuthGuard } from '../auth/jwt-auth.guard'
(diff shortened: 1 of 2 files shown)


ccm_web/server/src/api/api.controller.ts
```diff
@@ -95,12 +97,13 @@ export class APIController {
...
-    @UserDec() user: User,
-      @Query('term_id', ParseIntPipe) termId: number,
-      @Query('instructor') instructor?: string,
-      @Query('course_name') courseName?: string
+    @Query() query: GetSectionsAdminQueryDto,
+      @UserDec() user: User
   ): Promise<CourseWithSections[]> {
-    const result = await this.apiService.getCourseSectionsInTermAsAdmin(user, termId, instructor, courseName)
+    if (query.instructor === undefined && query.course_name === undefined) {
+      throw new BadRequestException('You must specify either instructor or course_name as a URL parameter.')
```
pushyamig, line 104: The Response in the FE is this Good! I would word this as `instructor_name` instead of `instructor`. That would better for understanding

{
  "statusCode": 400,
  "message": "You must specify either instructor or course_name as a URL parameter.",
  "error": "Bad Request"
}

chrisrrowland: Are you saying he should rename the parameter, because the parameter name and the message are currently consistent.

ccm_web/server/src/api/dtos/api.get.sections.admin.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+  @IsInt()
+  @IsNotEmpty()
+  @Transform(({ value }) => Number.parseInt(value))
+  term_id: number
+
+  @ApiProperty({ type: String, required: false })
+  @IsOptional()
+  @MinLength(3)
+  @MaxLength(255)
+  instructor?: string
```
pushyamig, line 16: Anytime search input is < 3 characters in instructor or course name the response will look like below. Probably a better variable name is `instructor_name` message saying `instructor_name` must be longer than or equal to 3 characters
```
{
  "statusCode": 400,
  "message": [
    "instructor must be longer than or equal to 3 characters"
  ],
  "error": "Bad Request"
}
```

pushyamig: An Observation from above comment w.r.t error response object. 
I don't think FE is capturing this object since `APIErrorData` `message ` is string instead of array. I think if something needs to change, it is beyond the scope of the issues
```
export interface APIErrorData {
  statusCode: number
  message?: string
  redirect?: true
  error?: string
  errors: APIErrorPayload[]
}
```

ccm_web/server/src/api/api.controller.ts
pushyamig: Yes! It will be clear for user if error case hit. I have some similar comments below

ssciolla: I can change the parameter name.

As a side note, we will probably want to tell the user that `instructor_name` has to be an exact match, while `course_name` can be a partial match.

pushyamig: >  want to tell the user that instructor_name has to be an exact match, while course_name can be a partial match.

Good Idea

ccm_web/server/src/api/dtos/api.get.sections.admin.dto.ts
ssciolla: Yeah, the typing needs to be improved here for sure in a couple ways. Probably it should be something like:

```
export interface APIErrorData {
  statusCode: number
  error?: string
}

export interface ApplicationAPIErrorData extends APIErrorData {
  message?: string | string[]
  redirect?: string
}

export interface CanvasAPIErrorData extends APIErrorData {
  errors: APIErrorPayload[]
}
```

with some maybe some type guards. This is indeed a separate issue.

ccm_web/server/src/api/api.controller.ts
ssciolla: @pushyamig, here you go: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/205/commits/25b6910ca5f91b01712587920e63c85eff307d7c

@chrisrrowland, sorry about the breaking change, but it is minor! It is slightly more clear, since `instructor` could be interpreted as a number ID or a name string. I guess `instructor_login_id` or would be the most clear...

pushyamig (approved):
Looks Good. Thanks for Quick turn around on this keeping @chrisrrowland time in mind.

ccm_web/server/src/api/dtos/api.get.sections.admin.dto.ts
ssciolla: https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/206

ccm_web/server/src/api/api.controller.ts
chrisrrowland: Does it really have to  be an exact match or does it support wild carding?  If it does support wildcarding do we want to automatically pad the search text on the backednd like "rowc" => "%rowc%" or whatever the appropriate method would be?  I did try searching for "row%" and "row*" through the front end and it didn't work, but I would assume the canvas documentation would say what was supported if anything is..

ssciolla: The instructor value gets passed to a `by_teachers` parameter, not a search parameter, so I believe it has to be exact.
https://canvas.instructure.com/doc/api/accounts.html#method.accounts.courses_api

chrisrrowland: Ok, looks like you'd have to do multiple api calls to lookup matching users or something, if  that's even possible, to allow that.  Thanks.

PR #207, opened 2021-09-29
Add `nonxlist_course_id` to `CanvasCourseSectionBase`

ssciolla:
This PR follows up on the work of #199, adding the `nonxlist_course_id` field to `CanvasCourseSectionBase` so that the FE can determine whether a course can be unmerged (de-crosslisted) or not (if it's `null`, it's currently part of its original course and can't be unmerged).

### DIFF
--- a/ccm_web/server/src/api/api.course.handler.ts
+++ b/ccm_web/server/src/api/api.course.handler.ts
@@ -74,6 +74,7 @@ export class CourseApiHandler {
         id: s.id,
         name: s.name,
         course_id: s.course_id,
+        nonxlist_course_id: s.nonxlist_course_id,
         total_students: s.total_students
       }))
     } catch (error) {
--- a/ccm_web/server/src/api/api.section.handler.ts
+++ b/ccm_web/server/src/api/api.section.handler.ts
@@ -25,7 +25,8 @@ export class SectionApiHandler {
     return {
       id: section.id,
       name: section.name,
-      course_id: section.course_id
+      course_id: section.course_id,
+      nonxlist_course_id: section.nonxlist_course_id
     }
   }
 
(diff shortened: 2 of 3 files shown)


pushyamig:
This looks good. I think it's mostly @chrisrrowland  call how you want to see the data.

chrisrrowland:
I will take a look momentarily.  I think it's going to be fine.  Functionally I think all I need to know is if that is null or not, so it will be utilized like a boolean.. I think.

I'm not aware of anywhere in the UI that it should be explicitly saying what course it has been crosslisted too, and even if it needed to, the course would be implied by context.

Long story short I think it's probably good this way but I'll check in a second.

ssciolla:
> I will take a look momentarily. I think it's going to be fine. Functionally I think all I need to know is if that is null or not, so it will be utilized like a boolean.. I think.

@chrisrrowland, I could add a calculated field for `crosslisted: boolean` if you'd like. I agree there probably isn't a need to know the course it came from, since it's not required in the `unmerge` call. I was inclined to just send the Canvas data as received (with thinning/slimming), but I didn't put too much thought into it.

chrisrrowland:
> > I will take a look momentarily. I think it's going to be fine. Functionally I think all I need to know is if that is null or not, so it will be utilized like a boolean.. I think.
> 
> @chrisrrowland, I could add a calculated field for `crosslisted: boolean` if you'd like. I agree there probably isn't a need to know the course it came from, since it's not required in the `unmerge` call. I was inclined to just send the Canvas data as received (with thinning/slimming), but I didn't put too much thought into it.

Totally  not necessary, just commenting on how it would be used.

chrisrrowland (approved):
This looks good to me.  Thanks for the quick work.

Issue #208, opened 2021-09-29
Integrate client Add UM Users feature with the backend enrollment endpoint


ssciolla:
With #165 and #177 merged, we can now connect the client code with the `api/sections/:id/enroll` endpoint. This will involve changes to `api.ts` and the `AddUMUsers.tsx` page, among other files.

Issue #206, opened 2021-09-29
Improve `APIErrorData` typing


ssciolla:
Yeah, the typing needs to be improved here for sure in a couple ways. Probably it should be something like:

```
export interface APIErrorData {
  statusCode: number
  error?: string
}

export interface ApplicationAPIErrorData extends APIErrorData {
  message?: string | string[]
  redirect?: string
}

export interface CanvasAPIErrorData extends APIErrorData {
  errors: APIErrorPayload[]
}
```

with some maybe some type guards. This is indeed a separate issue.

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/205#discussion_r718497669_

In reference to the `APIErrorData` types in the frontend. Some modifcations to the backend may also make sense to increase consistency in data returned.

pushyamig:
Thanks @ssciolla for opening the issue

pushyamig:
Test Passes. I don't see any typing errors

Issue #197, opened 2021-09-22
User not informed of upload file size limits


pushyamig:
@pushyamig wrote…
We want only to handle 400 users enrollments to section. A simple text some where along with the sample download file would be useful. Also throw a validation error after the file upload if saw > 400 user 

---

@lsloan wrote…
Users need to be informed of the file size limits when uploading data for the following features:

* "Create Sections" — 60 sections — Also see #217
* "Add UM Users" — 400 users
* "Canvas Gradebook Formatter" — Does this have a limit?

Implementation notes:

* As mentioned in #217, checks of file sizes need to occur in the UI **_AND_** the API.
* Limits should be a configuration item in `.env` and available to both UI **_AND_** API code.

ssciolla:
We need to fix this in a few of the other features.

pushyamig:
Test passes.

PR #214, opened 2021-10-05
Connect Add UM Users to backend, extract some common components from pages (#206, #208)

ssciolla:
This PR integrates the backend section enrollment endpoint with the Add UM Users feature. It also improves on logic and typing for API errors (see `handleErrors.tsx` and `models.ts`) and extracts a number of stateless components for re-use (`SuccessCard`, `APIErrorAlert`, `RowLevelErrorsContent`, `CSVFileName`) across the pages and to facilitate future feature development.

The PR is still a draft, but initial feedback would be welcome.


ccm_web/client/src/components/CanvasAPIErrorsTable.tsx
```diff
@@ -0,0 +1,30 @@
...
+  // Thinking this will just confuse users, so may not include it.
+  // { id: 'canvasStatusCode', label: 'Canvas Status Code', minWidth: 100 },
+  { id: 'failedInput', label: 'Failed Input', minWidth: 125 },
+  { id: 'message', label: 'Error Message', minWidth: 125 }
+]
+
+function CanvasAPIErrorsTable (props: CanvasAPIErrorsTableProps): JSX.Element {
+  const [page, setPage] = useState<number>(0)
+
+  const tableRows = props.errors.map((e, i) => ({ ...e, rowNumber: i + 1 }))
```
ssciolla, line 25: May need to figure out how to map back these errors to their row numbers, though I'm not exactly sure how to do that effectively. If we knew each record was unique, that'd be one thing, but we'd still need to parse `failedInput`, which is a bit fragile and hard to be consistent with.

lsloan:
WRT the grammar addition, I agree that "to which to add users" is a better way to end that sentence.  I still maintain that "select _one available_ section" is awkward and should be "…_an available_…" or "…_an existing_…".

Let's see what @janelilr thinks, as you suggest.

Thanks for updating the sorting.  I probably would've removed the sorting and used with the order Canvas' API returns, because that seems to be already sorted.  I guess there's no guarantee their result will always be sorted, though.

ssciolla:
I'm not entirely confident this is bug-free, but I'm going to open it up for review.

One thing I'd especially like feedback on is the use of `APIErrorAlert` and `RowLevelErrorsContent` (in combination with `CanvasAPIErrorsTable`) to make API errors consistent and give the user more feedback. I've reviewed the uses of these and **I think** they are appropriate, but I can see us needing to do more edge case testing.

janelilr:
> Please use: "Or select an existing section to add users to"

Thanks @ssciolla and @lsloan for the text suggestion

lsloan:
FYI:  I clicked "approve" expecting it to approve only the changes I requested earlier (grammar and sorting).  I didn't expect it to give my approval to the PR overall.  So, take that however you'd like.

So far, I only tested the feature I was most familiar with, the "Add UM Users" feature.  I found several problems with the CSV parsing and issues will be created or updated to address them.

I'm not yet done testing other features or looking closely at the code.  I'm doing that now.  If others feel the PR should be merged before I'm done, I won't be disappointed.

pushyamig:
@ssciolla  I think we have established the fact that We can support 400 user, We should have this in `Requirements` section saying that `The Tool has a limits of at most 400 the user enrollments` . I think I created an issues for that https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/197. I am ok if you want to deal with as separate issu

Note: Please change the Text as needed

ssciolla:
@lsloan and @pushyamig, thank you for the reviews. I just added a new commit (https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/214/commits/1c1a856d44afa16e27968ffda3b5baafc50835b7) that probably merits some additional testing of `AddUMUsers.tsx`. We should now be handling the empty `loginId` and max record number cases highlighted by @pushyamig. Could one of you please review the changes and do a quick test run of `AddUMUsers.tsx`?

ssciolla:
Thanks all for the feedback, I'm merging now.

Issue #235, opened 2021-11-03
Update NodeJS, TypeScript, npm packages


ssciolla:
Before an initial release, we ought to do a dependency update so we are resolve any known vulnerabilities and set the project up for the future. Here are some things to note (more may be added):

- Node 16 is now the LTS: https://nodejs.org/en/blog/release/v16.13.0/
- TypeScript has had a few minor releases; we're now at [4.4](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html).
- `@kth/canvas-api` now natively supports TypeScript, so we can remove the custom types we're using now (yay!)
- It'll probably make sense to update the NestJS dependencies at the same time (see #173).
- It will probably not make sense to tackle the Material UI 5 upgrade at the same time, since it will probably be a bigger lift. We can upgrade to the latest 4 version though (see #220).

pushyamig:
All the version upgrades mentioned here are updated on ccm-test. 

Test Pass

Issue #220, opened 2021-10-12
Upgrade Material UI


chrisrrowland:
Version [5.0](https://mui.com/versions/) of react material was recently released.

ssciolla:
Material UI 5 includes quite a few changes, many centered around styling, but there is a migration guide and some automatic code modification provided it seems: https://mui.com/guides/migration-v4/

ssciolla:
There is this issue in `React.StrictMode` when using `Backdrop` with Material UI 4 during local development. Let's check to see if this is resolved when we update to version 5. https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/75#issuecomment-950954119

ktowneUM:
create sections was tested

Issue #75, opened 2021-05-05
Create global UI style/theme


chrisrrowland:
Not sure what the correct terminology but we should be using some sort of global theme/style throughout the application.

This seemed to me like something you might implement after having more than one feature so you have an idea of common concepts, though it couldn't hurt to have the framework in place earlier.

Currently each component is composing its own style.

pushyamig:
@ssciolla Can you provide some estimation here?

ssciolla:
Well, it depends on what this issue is. If it's implementing a global theme, I can estimate that, but fixing the problem you identified may involve updating to Material UI 5, which is a bigger deal, harder to estimate. These should be split into separate issues.

pushyamig:
I am talking about the error, possible fix which I understood as having a global theme, if that is not the case. I feel we need to evaluate if we can live with this error for sometime or upgrade to Material UI.

ssciolla:
Okay, I gave an estimate. I still think that needs to be a separate issue. It's related but not the same as what Chris originally opened this issue about.

chrisrrowland:
I believe I did some investigating, and that led to some article that led to me to experimenting w/ a theme to eliminate the warning message.
I don't remember why the theme solved it, but since we wanted to implement a global theme anyways it seemed like a reasonable solution.

PR #236, opened 2021-11-08
Update dependencies (#173, #235)

### LINKED ISSUE #173: Upgrade to Nest version 8
Looks like there are some nice improvements around `HttpModule` and configuration; we should update this before the initial release.

https://docs.nestjs.com/migration-guide

ssciolla:
This PR updates most project dependencies, with some changes being held back or deferred. See the tasks list below for more information. The PR aims to resolve #173 and #235.

Task list
- [x] Upgraded from `node` 14 (`npm` 6) to `node` 16 (`npm 8`). (This resulted in many changes to `package-lock.json`.) **Contributors should switch to using `node` 16 locally (i.e. install using `brew` or other approach and update your `PATH` environment variable).**
- [x] Upgraded TypeScript version from 4.2 to 4.4 (a couple of code changes were needed because of a change in how caught errors are typed in strict mode (`unknown`). See [here for more info](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#using-unknown-in-catch-variables).
- [x] Migrated from `@nestjs` version 7 to version 8. Changes include the new use of `@nestjs/axios` library (where `HttpModule` was moved), tweaks to how `rxjs` was used with `HttpModule` due to a deprecation of `.toPromise()`, and a new approach to type inference of config variables and declaring they are defined when returned from `ConfigService.get`.
- [x] Adopt a pre-release version of `@kth/canvas-api` that provides native TypeScript support (allowing us to remove our custom types, yay!) (This required some adjustments to handle breaking changes related to method names, see https://github.com/KTH/canvas-api/releases/tag/v4.0.0-1). 
- [x] Change some prop usage with Material UI table and `Grid` components in response to deprecations/interface changes.
- [x] Adjust to a typing change in `papaparse` (specifically `Papa.unparse`).
- [x] Fix `docker-compose-prod.yml` so `wait-port` is used by overriding `entrypoint`.
- [x] Prep `client/src/index.tsx` and `client/src/App.tsx` for later `mui` 5 and `react-router-dom` 6 changes (see https://reactrouter.com/docs/en/v6/upgrading/v5).
- [x] Remove the following packages from `package.json` because they are unnecessary or will be installed automatically by `npm` 8 as they are peer dependencies: `@types/helmet` (`helmet` has its own types), `reflect-metadata` (peer)
`@types/sequelize` (`sequelize` has its own types), `eslint-plugin-import` (peer), `eslint-plugin-node` (peer), and `eslint-plugin-promise` (peer).

Postponements
- Upgrading to Material UI 5 is going to be a bigger deal, so I figured we'd do that work separately (see issue #220).
- I elected to wait on ESLint 8, because `eslint-config-with-standard` doesn't seem to support it yet (see https://github.com/standard/eslint-config-standard-with-typescript/issues/728).
- `react-router-dom` 6 was just released a few days ago (as of 11/9), so I am waiting on this one. I may make the switch (which involves breaking changes) at the same time as work on #220.


ccm_web/package.json
```diff
@@ -19,98 +19,93 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@kth/canvas-api": "^3.0.3",
+    "@kth/canvas-api": "^3.3.0-1",
```
ssciolla, line 22: I guess this is still technically a pre-release. Should we wait to adopt this?

ccm_web/package.json
```diff
@@ -14,103 +14,98 @@
...
+    "js-cookie": "^3.0.1",
     "keycode-js": "^3.1.0",
-    "ltijs": "~5.7.3",
-    "ltijs-sequelize": "~2.4.3",
+    "ltijs": "~5.8.2",
+    "ltijs-sequelize": "~2.4.4",
     "morgan": "^1.10.0",
-    "notistack": "^1.0.5",
-    "papaparse": "^5.3.0",
+    "mysql2": "^2.3.2",
```
ssciolla, line 50: `mysql` was already available before because of `litjs`, but because we're using it without `ltijs`, it seems like we should specify it going forward.

ssciolla:
I'm opening this up for review. Please read the notes above, as they should explain most of the changes I made (or did not make).

As an aside, I found this tool helpful for checking for new versions: https://github.com/raineorshine/npm-check-updates

pushyamig:
I plan to review it sometime in the afternoon.

pushyamig:
The pagination seems to be broken once you upload the csv file. I checked the with the version on the Dev instance and that is working fine.

ccm_web/client/src/components/CustomTable.tsx
```diff
@@ -81,7 +81,7 @@ function CustomTable<T extends TableEntity> (props: TableProps<T>): JSX.Element
                 inputProps: { 'aria-label': 'rows per page' },
                 native: true
               }}
-              onChangePage={handleChangePage}
+              onPageChange={handleChangePage}
```
pushyamig, line 84: Even though you have change the function name to `onPageChange` when i am clicking the pages button it is saying `onChangePage` is not a function. So did you changed in all places? [onChangePage](https://stackoverflow.com/a/67077268/724238)

ssciolla: Good catch, I'll take a look.

ssciolla: Okay, this should have fixed it, @pushyamig: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/236/commits/12f2efaeaa3bff68618217b7f27a1bde7a4ab3de

pushyamig:
I still see this error nothing new we did not know about. We were hoping upgrade to the libraries might fix this. This issue is commented here #75. Should we create a new issue to handle it or we still believe that resolving #75 will fix this? 
```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node
`````

**Comment by ssciolla** [comment]
> I still see this error nothing new we did not know about. We were hoping upgrade to the libraries might fix this. This issue is commented here #75. Should we create a new issue to handle it or we still believe that resolving #75 will fix this?
> 
> ```
> Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node
> ```

I think the best way to fix this will be to upgrade to Material UI 5. I think that version will have better support for `React.StrictMode`.

**Comment by pushyamig** [comment]
> I think the best way to fix this will be to upgrade to Material UI 5. I think that version will have better support for React.StrictMode.

Upgrade to MUI-5 is not in Icebox, so should we let this Warning be logged in Console log for user? Your estimate of MUI-5 upgrade is non-trival. I think we need to set the expectation to all that this Warning will happen so ignore for now. WDYT?

**Comment by pushyamig** [comment]
Canvas Auth flow revoking the authorization in canvas is not working. I verified it is working in CCM-Dev

1. Go the Canvas -> Accounts -> Settings-> Delete the authorization 
2. Launch CCM and you will see 'CourseInfo not found'
Expected result it should redirect to Pre-Auth page in React.

If you want to get around this you need to delete the token from DB then the pre-Auth page comes in to generate the token.

**Comment by ssciolla** [comment]
> Canvas Auth flow revoking the authorization in canvas is not working. I verified it is working in CCM-Dev
> 
>     1. Go the Canvas -> Accounts -> Settings-> Delete the authorization
> 
>     2. Launch CCM and you will see 'CourseInfo not found'
>        Expected result it should redirect to Pre-Auth page in React.
> 
> 
> If you want to get around this you need to delete the token from DB then the pre-Auth page comes in to generate the token.

Darn, okay, I will look into it.

**Comment by pushyamig** [comment]
Another test case odd behavior around Canvas Auth testing

1. Go To Canvas - > Developer Key -> go to YOUR_API_KEY_Config -> Undo a needed API call by CCM
2. You will see an error message in UI "CourseInfo doesn't exists"
Error in Logs 
```
{"body":{"errors":[{"message":"Invalid access token."}]}
```` 

I Think it is erring out but somewhere in BE it is allowing the API call `/api/course/403334` after the error.
The error should be appropriately displayed saying invalid access token and not say course info doesn't exist.

I think this might be somewhat related to this but not exactly the same https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/104

**Comment by pushyamig** [comment]
Testing the app from both prod and local mode
1. The FE Features are all seems to be working fine. Tested with different user by masquerading. Token was generated and things were fine. 
2. The webpack is loading the new JS file when code changes.
3. Nodemon is restarting and VScode debugger is working as expected.
4. Swagger upgrade is good
5. DB integration is looks good with new DB and existing DB
6. Wait ports are working fine

Found some issue around Canvas Auth flow but if those issues are fixed then this PR is good to go. This is a Huge dependency update and it might not possible to check every single case. I looked at the major integration point. Pesky little bugs might come up and we should handle it as we go as more people test this.

**Comment by ssciolla** [comment]
@pushyamig, please try testing the OAuth workflow after these changes: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/236/commits/2211bbc22825829e81a5fa26f338ac65b0eba601

To elaborate on the issue, KTH introduced a new CanvasApiError object which is a thinned version of `got.HTTPError`. It's possible we could have stayed with the same version if we had added to the types, but KTH already has a TypeScript version in the works. I elected to update to `4.0.0-0` so as to stay ahead of the curve (I tried `3.3.0-1` but I had issues with the typing). When I have a moment, I will follow up with them about their non-pre-release plans. I will also update the notes above related to this.

File: ccm_web/package.json (hunk 1 above)
Comment (ssciolla) [reply ccm_web/package.json]: Okay, because of the OAuth issue caused by an error type change, I had to change my approach here again. I am going with `4.0.0-0`, which required some method name changes as well.

**Review by pushyamig — APPROVED** [review]
Looks Good!

File: ccm_web/client/src/App.tsx [hunk 4]
```diff
@@ -1,7 +1,6 @@
 import React, { useEffect, useState } from 'react'
-import { BrowserRouter as Router, Route, Switch } from 'react-router-dom'
+import { Route, Switch, useLocation } from 'react-router-dom'
```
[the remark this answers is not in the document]
ssciolla: I alluded to this in the above note, this is part of changes that the`react-router-dom` is recommending, using their new hooks and `<Route>{children}</Route>` instead of `<Route render={() => {}} />`. See https://reactrouter.com/docs/en/v6/upgrading/v5 and https://reacttraining.com/blog/react-router-v5-1/.

ssciolla:
@pushyamig, thank you for the speedy and thorough review; you caught a couple very important issues and I would have been disappointed (in myself) if I committed them to `main`. I looked over everything again and did a quick round of testing. I agree that it's difficult to catch everything. Please let me know if you find dependency related issues going forward, I'm happy to work on resolving them. I'm merging now!

Issue #240, opened 2021-11-11
`InlineTextEdit` component is not accessible


ssciolla:
Issue was reported during the accessibility review on 11/2/21.

"
Issue
Edit title component is not accessible.

Impact
Screen reader and keyboard-only users will not be able to operate it.

Solution
Many possibilities. Possibly the least complicated fix is to make the pencil icon or the parent div (or whatever is responding to onclick) an actual button. Make sure to add an aria-label=""Edit title"" to the control and to pass the focus to the text input when the button is activated. If the element that is listening is the SVG, any of the labelling techniques will do:
https://dequeuniversity.com/rules/axe/4.0/svg-img-alt"
"

gsilver:
Fixed!

Issue #151, opened 2021-07-12
Modify Course name color codes 


pushyamig:
Comments from @janelilr 

![Screen Shot 2021-07-06 at 11 47 57 AM](https://user-images.githubusercontent.com/8579775/125303314-b5849b00-e2fa-11eb-900a-712d1098de82.png)
![Screen Shot 2021-07-06 at 11 47 43 AM](https://user-images.githubusercontent.com/8579775/125303335-b9b0b880-e2fa-11eb-8d75-77809f4caa6f.png)
 In regards to the above screenshot about the color difference in the course name title with and without edit roles, the color of text should be reversed. If a user is unable to edit, the text should be displayed in gray (indicates to users this cannot be changed). If a user can edit, the text would be black.

ssciolla:
@pushyamig and @janelilr, I feel maybe this was fixed a while ago and the issue was never resolved. The current version of the code seems to have a disabled style when in non-editing mode, and a regular style when editing.

That said, I find it strange that what is supposed to be the page's header (or at least that's purpose it seems to serve to me) has a disabled style. In PR #241, I am proposing an alternative solution, where we use regular text in both cases, but the text in the non-editing case is an `h1`. I feel the absence of the text edit field underline is key enough that the text can't be edited until you click the button. Let me know your thoughts. I will add screenshots below.
![Screen Shot 2021-11-12 at 1 30 23 PM](https://user-images.githubusercontent.com/35741256/141517141-26c505f5-9990-400c-99f6-248c18e1c61e.png)
![Screen Shot 2021-11-12 at 1 30 39 PM](https://user-images.githubusercontent.com/35741256/141517144-3127848a-6adf-4bd3-a7bd-e78a848ce21c.png)

pushyamig:
>I feel the absence of the text edit field underline is key enough that the text can't be edited until you click the button. Let me know your thoughts.

I will defer to @janelilr on this.

janelilr:
> @pushyamig and @janelilr, I feel maybe this was fixed a while ago and the issue was never resolved. The current version of the code seems to have a disabled style when in non-editing mode, and a regular style when editing.
> 
[...]

PR #241, opened 2021-11-12
Refactor `InlineTextEdit` and course name usage to improve usability/accessibility (#240)

ssciolla:
This PR refactors the `InlineTextEdit` component to provide a few improvements (see below), mainly the accessibility features suggested in issue #240. The PR aims to resolve issue #240 and maybe #151.

Improvements
- Pencil edit button is keyboard-accessible (changed underlying component to `Button`).
- Text field receives focus when the edit button is clicked.
- Text field has been made larger so edits are easier to see visually.
- The course name text is now an `h1` instead of a disabled text field when in not-editing mode. This also allows for display of long course names (which now wrap). It may also resolve other accessibility issues related to a missing `h1` on the home page.
- ARIA labels have been added to all involved buttons.
- When a request fails (because of a bad connection, or a Canvas validation error), the revised version is not retained in the component's state when editing begins again. This could be seen as a problem, but it seemed more logical to have the `h1` text and the temp text always match when a user begins editing. Let me know your thoughts.
- Text field remains in a disabled state while the change is being saved to Canvas.

### DIFF
--- a/ccm_web/client/src/components/InlineTextEdit.tsx
+++ b/ccm_web/client/src/components/InlineTextEdit.tsx
@@ -100,10 +100,18 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
                       <Button
                         className={classes.button}
                         disabled={props.text === tempTextValue || props.isSaving}
-                        onClick={save}>
+                        onClick={save}
+                        aria-label='Save course name'>
                           Save
                       </Button>
-                      <Button className={classes.button} disabled={props.isSaving} onClick={cancel}>Cancel</Button>
+                      <Button
+                        className={classes.button}
+                        disabled={props.isSaving}
+                        onClick={cancel}
+                        aria-label='Cancel editing course name'
+                      >
+                        Cancel
+                      </Button>
                     </Grid>
                   </Grid>
                 </form>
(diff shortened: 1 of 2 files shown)


pushyamig:
I plan to review this in the afternoon

pushyamig:
> When a request fails (because of a bad connection, or a Canvas validation error), the revised version is not retained in the component's state when editing begins again. This could be seen as a problem, but it seemed more logical to have the h1 text and the temp text always match when a user begins editing. Let me know your thoughts.

Can you explain what do you mean by that?

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -1,114 +1,134 @@
...
   const [isEditing, setIsEditing] = useState(false)
-  const [textValue, setTextValue] = useState(props.text)
   const [tempTextValue, setTempTextValue] = useState(props.text)
   const textInput = useRef(null)
 
-  const save = (): void => {
+  // Resets tempTextValue in case when save fails
+  useEffect(() => {
+    if (isEditing) setTempTextValue(props.text)
+  }, [isEditing])
```
pushyamig, line 47: Nice

pushyamig (approved):
Nice Work! Accessibility testing sounded right from Mac VoicerOver and Keyboard to me. 
We might need to create a new issue for validation around empty course name and name >255 character which we did not realize earlier. Adding a Validation will prevent from Canvas API call since canvas don't allow these cases either.

Issue #217, opened 2021-10-08
"Create Sections" fails with large input


lsloan:
# Problem

The "Create Sections" feature failed when given a file of 99,999 sections to create.  Sample file: [sections2.csv](https://github.com/tl-its-umich-edu/canvas-course-manager-next/files/7313298/sections2.csv)

The UI was able to parse the large file and show the rows on the page.  After submitting, the UI said…

> The last action failed with the following message: Internal server error

The server console output had the error…

```
ccm_web     | [Nest] 102   - 10/08/2021, 6:01:50 PM   [ExceptionsHandler] request entity too large +481950ms
ccm_web     | PayloadTooLargeError: request entity too large
ccm_web     |     at readStream (/base/node_modules/raw-body/index.js:155:17)
ccm_web     |     at getRawBody (/base/node_modules/raw-body/index.js:108:12)
ccm_web     |     at read (/base/node_modules/body-parser/lib/read.js:77:3)
ccm_web     |     at jsonParser (/base/node_modules/body-parser/lib/types/json.js:135:5)
ccm_web     |     at Layer.handle [as handle_request] (/base/node_modules/express/lib/router/layer.js:95:5)
ccm_web     |     at trim_prefix (/base/node_modules/express/lib/router/index.js:317:13)
ccm_web     |     at /base/node_modules/express/lib/router/index.js:284:7
ccm_web     |     at Function.process_params (/base/node_modules/express/lib/router/index.js:335:12)
ccm_web     |     at next (/base/node_modules/express/lib/router/index.js:275:10)
ccm_web     |     at /base/node_modules/express-session/index.js:506:7
ccm_web     |     at /base/node_modules/connect-session-sequelize/lib/connect-session-sequelize.js:22:7
ccm_web     |     at runMicrotasks (<anonymous>)
ccm_web     |     at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

# To-Do

Check whether the "Add UM Users" feature is also affected in this way.

# Solution

The solution for this has several parts:

* The error above indicates the problem was in sending the large input to the API.  Therefore, because the UI was able to handle the large input, it should give an error that the input is too long before the user is given the option of submitting it.
* The `CreateSectionsDto` should have a limit to disallow any input it received that is too large.  See `SectionUsersDto` for an example.
* Similar to issue #197, the user needs to be informed that the maximum number of sections they can upload is 60.

pushyamig:
[...]

PR #248, opened 2021-11-18
handle large input for section creation (iss. #217)

zqian:
- increased the request payload limit to avoid PayloadTooLargeError
- also enforced the class-validator check so that only 60 sections can be uploaded and created at one time


ccm_web/server/src/main.ts
```diff
@@ -47,6 +48,11 @@ async function bootstrap (): Promise<void> {
   const sessionStore = new SequelizeStore({ db: sequelize, tableName: 'session' })
   sessionStore.sync({ logging: (sql) => logger.info(sql) })
 
+  // this is to increase the size limit for request payload, which defaults to 100kb
+  // then the array element limit will be enforced by the class-validator in CreateSectionsDto
+  app.use(json({ limit: '50mb' }));
+  app.use(urlencoded({ extended: true, limit: '50mb' }));
```
lsloan, line 54: Is `50mb` a good size for the amount of data to be uploaded?  It seems too large.  Is it based on the total size of 60 lines of data with each field filled to its maximum capacity?  In this case, 60 lines of section IDs, which is just a few bytes each.

It would be better to specify the max. number of lines in some variable, then calculate a reasonable file size for that here.

The max. file size (in kb or mb) and the max. number of sections should be shown in the UI as well, so if the file size error or the number of lines exceeded error comes up, the user will not be caught by surprise.

What kind of error is generated when a file too large or too many lines is uploaded?  Is the error caught and shown to the user in an easily understandable way?

lsloan (commented):
I have a few questions…

ccm_web/server/src/main.ts
zqian: > Is 50mb a good size for the amount of data to be uploaded? 
I am not sure what is the best number for the limit here. I was testing with the file [sections2.csv](https://github.com/tl-its-umich-edu/canvas-course-manager-next/files/7313298/sections2.csv) as mentioned in the #217 description, which has size of 1.7MB. 

I agree that ideally the upload file size should not be too big, and should be within the limit of kb, less than MB.

But I figure once we setup up the payload limit to a big enough number, the upload process won't error out at this stage, but rather pass through all data into the next check for section array size ( less than 60 is allowed), and then user will get a nicely formatted message in UI.

> This does seem to be the way to do this, although the Nest folks seem to recommend using body-parser..
I saw the body-parser suggestion and also the suggest as in this PR. 

> My question is if this is really needed if we never hit the default limit with the line limits in place...
I did that since that was the error originally reported in #217

ssciolla: Okay, you are right, something like this is needed, since the record numbers aren't considered until the body is parsed.

I think maybe `50mb` is excessive, no one is going to upload something that size, and if they did, I wouldn't want the application to hang while processing it. How about something like `5mb`? Remember only `.csv` files are accepted.

I will test this this afternoon.

zqian: It is 5mb now

ccm_web/server/src/main.ts
```diff
@@ -10,6 +10,7 @@ import { ConfigService } from '@nestjs/config'
 import { NestFactory } from '@nestjs/core'
 import { NestExpressApplication } from '@nestjs/platform-express'
 import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger'
+import { urlencoded, json } from 'express';
```
ssciolla, line 13: Codacy has flagged 4 issues that probably also show up in your IDE if you have linting set up. We're using `standard`, so "no unnecessary semicolons".

ccm_web/server/src/api/dtos/api.create.sections.dto.ts
```diff
@@ -1,13 +1,19 @@
...
+  ArrayMaxSize,
+  ArrayMinSize,
+  IsArray
+} from 'class-validator'
 import { ApiProperty } from '@nestjs/swagger'
 
 export class CreateSectionsDto {
   @ApiProperty({ type: [String] })
-  @IsNotEmpty({ each: true })
-  @MaxLength(250, { each: true })
```
ssciolla, line 7: Why did you remove these? I think we still want to enforce that each section name string is `> 0` and `< 256` characters, because otherwise users will hit up on Canvas validation errors.

ccm_web/server/src/api/dtos/api.create.sections.dto.ts
```diff
@@ -1,13 +1,19 @@
...
+  @IsArray()
+  @ArrayMinSize(1)
+  @ArrayMaxSize(60)
```
ssciolla, line 12: This is working!

ccm_web/server/src/main.ts
ssciolla: Maybe consider moving this up above `import session from 'express-session'`?

ccm_web/server/src/main.ts
```diff
@@ -47,6 +48,11 @@ async function bootstrap (): Promise<void> {
   const sessionStore = new SequelizeStore({ db: sequelize, tableName: 'session' })
   sessionStore.sync({ logging: (sql) => logger.info(sql) })
 
+  // this is to increase the size limit for request payload, which defaults to 100kb
+  // then the array element limit will be enforced by the class-validator in CreateSectionsDto
+  app.use(json({ limit: '5mb' }));
+  app.use(urlencoded({ extended: true, limit: '5mb' }));
```
ssciolla, line 54: So the benefit of increasing limit this is that the Express server can process the large payload, which it can then validate? I guess that's something. It'd be nice to know if hitting `100K` is actually that common with CSV files (in any feature). 400 enrollment records is around `7K`, so linearly, that's like 5200. If we think people will accidentally upload files with 10,000 records, then we are giving them a marginally better error message. It just feels like a small edge case. What happens now in the frontend (if you circumvent the CSV record limits in the `client`) is that user will see a "request entity too large" message, which is worse than "60 (or 400) is the limit". I guess that's a long-winded way of saying, I don't think it's essential we add this, but it also shouldn't hurt anything.

ssciolla: Comments feel a bit unnecessary. You could just say briefly `// Controls size limit of data in payload and URL`

ssciolla (commented):
This looks okay, I think we want to retain the string length validation that you removed though. Would be nice to have the minor Codacy/linting stuff fixed.

ccm_web/server/src/main.ts
zqian: Done: removed the semicolons and moved up the line.

ccm_web/server/src/api/dtos/api.create.sections.dto.ts
zqian: You are right. I added those checks back.

ccm_web/server/src/main.ts
zqian: updated the comment line

zqian:
@ssciolla @lsloan I think I have addressed all the review comments and Codacy issues. Please take another look.

pushyamig:
@ssciolla and @lsloan , This needs a re-review and approval.

ssciolla:
Looks good, I will do a little testing and then approve.

ccm_web/server/src/main.ts
```diff
@@ -47,6 +48,10 @@ async function bootstrap (): Promise<void> {
   const sessionStore = new SequelizeStore({ db: sequelize, tableName: 'session' })
   sessionStore.sync({ logging: (sql) => logger.info(sql) })
 
+  // Controls size limit of data in payload and URL
+  app.use(json({ limit: '5mb' }))
+  app.use(urlencoded({ extended: true, limit: '5mb' }))
```
ssciolla, line 53: Could extract `5mb` as a variable, not a huge deal though.

ssciolla (approved):
I think this works. Good to know about the limit capability, though I don't think it will basically ever come up, since the client will know the line number before data gets sent to the backend. Thanks for adding the missing validation to the endpoint!

PR #255, opened 2021-12-02
Issue-142 

zqian:
- [x] defined the allowed role list for using CCM tool as: ['TeacherEnrollment', 'TaEnrollment','ObserverEnrollment','DesignerEnrollment', 'Account Admin','Sub-Account Admin']
- [x] show error message to user when the user has at one role not in the above list

image


ssciolla:
Looks like this includes commits from your previous PR. Please remove those, and then I'll review.

Issue #253, opened 2021-11-30
Redirect user to "Access Denied" page once their JWT cookie or session expires


ssciolla:
Right now, if a 401 or 403 occurs during a backend call (unrelated to Canvas), the application gives generic error messages but doesn't send the user anywhere or give them valuable feedback.

I'm not sure of the best solution or implementation here, but one possible solution is to send the user to a generic "Access Denied" page if the client receives a `401` or `403` un-related to Canvas. This page could be the same or similar to what is described in issue #209, the text for which is currently in `App.tsx`.

There is some awkwardness with how the UI currently determines whether the user is authenticated (it makes a call to `/api/globals` to determine whether it has the right cookie, etc.). Some of this may need to be examined or reworked to resolve this issue.

pushyamig:
Node backend issue, hence closing.

PR #262, opened 2021-12-02
issue-142 prevent user with wrong roles to access CCM tool

zqian:
- [x] defined the allowed role list for using CCM tool as: ['TeacherEnrollment', 'TaEnrollment','ObserverEnrollment','DesignerEnrollment', 'Account Admin','Sub-Account Admin']
- [x] show error message to user when the user has at one role not in the above list

image

### DIFF
--- a/ccm_web/server/src/canvas/canvas.interfaces.ts
+++ b/ccm_web/server/src/canvas/canvas.interfaces.ts
@@ -66,7 +66,7 @@ export interface CanvasCourseSection extends CanvasCourseSectionBase {
   total_students: number
 }
 
-export enum CanvasRoles {
+export enum CanvasRole {
   AccountAdmin = 'Account Admin',
   SubAccountAdmin = 'Sub-Account Admin',
   SupportConsultant = 'Support Consultant',
@@ -79,23 +79,23 @@ export enum CanvasRoles {
 }
 
 export enum UserEnrollmentType {
-  DesignerEnrollment = CanvasRoles.DesignerEnrollment,
-  ObserverEnrollment = CanvasRoles.ObserverEnrollment,
-  StudentEnrollment = CanvasRoles.StudentEnrollment,
-  TaEnrollment = CanvasRoles.TaEnrollment,
-  TeacherEnrollment = CanvasRoles.TeacherEnrollment
+  DesignerEnrollment = CanvasRole.DesignerEnrollment,
+  ObserverEnrollment = CanvasRole.ObserverEnrollment,
+  StudentEnrollment = CanvasRole.StudentEnrollment,
+  TaEnrollment = CanvasRole.TaEnrollment,
+  TeacherEnrollment = CanvasRole.TeacherEnrollment
 }
 
 // valid role types for LTI launch
 // as defined in https://docs.google.com/spreadsheets/d/1pm5y9FX0zrDX7Zy3mOyDLxmA-iKoWfmlxvbtNkWg6Zw/edit#gid=1360549473
-export enum LTIEnrollmentTypes {
-  AccountAdmin = CanvasRoles.AccountAdmin,
-  SubAccountAdmin = CanvasRoles.SubAccountAdmin,
-  SupportConsultant = CanvasRoles.SupportConsultant,
-  TeacherEnrollment = CanvasRoles.TeacherEnrollment,
-  DesignerEnrollment = CanvasRoles.DesignerEnrollment,
-  TaEnrollment = CanvasRoles.TaEnrollment,
-  Assistant = CanvasRoles.Assistant
+export enum LTIEnrollmentType {
+  AccountAdmin = CanvasRole.AccountAdmin,
+  SubAccountAdmin = CanvasRole.SubAccountAdmin,
+  SupportConsultant = CanvasRole.SupportConsultant,
+  TeacherEnrollment = CanvasRole.TeacherEnrollment,
+  DesignerEnrollment = CanvasRole.DesignerEnrollment,
+  TaEnrollment = CanvasRole.TaEnrollment,
+  Assistant = CanvasRole.Assistant
 }
 
 export interface CanvasEnrollment {
--- a/ccm_web/server/src/canvas/canvas.interfaces.ts
+++ b/ccm_web/server/src/canvas/canvas.interfaces.ts
@@ -74,6 +74,18 @@ export enum UserEnrollmentType {
   TeacherEnrollment = 'TeacherEnrollment'
 }
 
+// valid role types for LTI launch
+// as defined in https://docs.google.com/spreadsheets/d/1pm5y9FX0zrDX7Zy3mOyDLxmA-iKoWfmlxvbtNkWg6Zw/edit#gid=1360549473
+export const LTIEnrollmentTypes = [
+  'Account Admin',
+  'Sub-Account Admin',
+  'Support Consultant',
+  'TeacherEnrollment',
+  'DesignerEnrollment',
+  'TaEnrollment',
+  'Assistant'
+]
+
 export interface CanvasEnrollment {
   id: number
   course_id: number
--- a/ccm_web/server/src/canvas/canvas.interfaces.ts
+++ b/ccm_web/server/src/canvas/canvas.interfaces.ts
@@ -128,7 +128,7 @@ interface CanvasError {
   message: string
 }
 
-function isCanvasError(value: unknown): value is CanvasError {
+function isCanvasError (value: unknown): value is CanvasError {
   return hasKeys(value, ['message'])
 }
 
@@ -136,7 +136,7 @@ export interface CanvasErrorBody {
   errors: CanvasError[]
 }
 
-export function isCanvasErrorBody(value: unknown): value is CanvasErrorBody {
+export function isCanvasErrorBody (value: unknown): value is CanvasErrorBody {
   if (!hasKeys(value, ['errors']) || !Array.isArray(value.errors)) {
     return false
   } else {
--- a/ccm_web/server/src/canvas/canvas.interfaces.ts
+++ b/ccm_web/server/src/canvas/canvas.interfaces.ts
@@ -28,7 +28,7 @@ export interface TokenCodeResponseBody extends TokenBaseResponseBody {
   refresh_token: string
 }
 
-export interface TokenRefreshResponseBody extends TokenBaseResponseBody {}
+export interface TokenRefreshResponseBody extends TokenBaseResponseBody { }
 
 // Entities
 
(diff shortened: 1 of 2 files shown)


ccm_web/server/src/lti/lti.service.ts
```diff
@@ -19,6 +19,9 @@ const createLaunchErrorResponse = (res: Response, action?: string): Response =>
   return res.status(HttpStatus.INTERNAL_SERVER_ERROR).json(message)
 }
 
+// the array of user roles that are allowed to use CCM tool
+const allowedRoles = ['TeacherEnrollment', 'TaEnrollment', 'ObserverEnrollment', 'DesignerEnrollment', 'Account Admin', 'Sub-Account Admin']
+
```
ssciolla, line 24: Are these roles defined anywhere else in the backend? I'm not exactly sure where the best place to put them is. This may be okay for now, but if we ever needed these roles in other parts of the application (like for role-based authorization), we'd need to re-think. Other places we might put them? `canvas.interfaces.ts`? I think these strings are Canvas specific, not part of the LTI spec. Let me know your thoughts. I'm not asking for any changes yet.

ssciolla (commented):
Looks okay! I will finish testing later, probably early this afternoon, and then give an approval if I don't see any issues.

ccm_web/server/src/lti/lti.service.ts
zqian: if the canvas.interfaces.ts is the central place for defining const and interfaces for server side code, then this definition could move there as another enum: AllowedUserEnrollmentType?

pushyamig: I see some of the these roles already mentioned in the `canvas.interface.ts`
```
export enum UserEnrollmentType {
  DesignerEnrollment = 'DesignerEnrollment',
  ObserverEnrollment = 'ObserverEnrollment',
  StudentEnrollment = 'StudentEnrollment',
  TaEnrollment = 'TaEnrollment',
  TeacherEnrollment = 'TeacherEnrollment'
}
```
I think we can create a new interface

```
export enum LTIRole extend UserEnrollmentType {
  AccountAdmin = 'Account Admin'
  SubaccountAdmin = 'Sub-Account Admin'
  Support Consultant = 'Support Consultant'
}
```

zqian: I have updated the lti role definitions into the canvas.interfaces.ts. Please review

pushyamig:
I will review this in the morning.

pushyamig:
This issue Works as expected giving access only roles as per CCM rules. 

But there were some discussion initially how the error be presented to the user from the BE (simple response message which is what this PR is doing) or via React UI using un authenticated route (`/`) once LTI launch is successfully. This is something @ssciolla suggested, so I will leave the approval upon him if he wants the design that way or leave it as is.

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -66,6 +67,14 @@ export class LTIService implements BeforeApplicationShutdown {
       const roles = customLTIVariables.roles as string
       const isRootAdmin = customLTIVariables.is_root_account_admin as boolean
 
+      // check whether the user roles are all included in the allowed LTI user roles
+      // otherwise show error message to the user
+      const rolesArray = roles.length > 0 ? roles.split(',') : []
+      const roleDiff = rolesArray.filter(x => !LTIEnrollmentTypes.includes(x))
```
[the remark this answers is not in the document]
zqian: The above lines are outdated. The variable now named `ltiAllowedRoles` in the new version

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -66,6 +67,14 @@ export class LTIService implements BeforeApplicationShutdown {
...
+      if (roleDiff.length > 0) {
+        return createLaunchErrorResponse(res, 'Your role in this course does not allow access to this tool. If you feel this is in error, please contact 4help@umich.edu.')
+      }
```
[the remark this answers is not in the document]
zqian: Right. The logic has changed accordingly, so that user will be granted access if he has at least one "allowed LTI" role

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -66,6 +67,14 @@ export class LTIService implements BeforeApplicationShutdown {
       const roles = customLTIVariables.roles as string
       const isRootAdmin = customLTIVariables.is_root_account_admin as boolean
 
+      // check whether the user roles are all included in the allowed LTI user roles
+      // otherwise show error message to the user
+      const rolesArray = roles.length > 0 ? roles.split(',') : []
```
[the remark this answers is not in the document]
zqian: renamed

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -66,14 +66,38 @@ export interface CanvasCourseSection extends CanvasCourseSectionBase {
...
+// as defined in https://docs.google.com/spreadsheets/d/1pm5y9FX0zrDX7Zy3mOyDLxmA-iKoWfmlxvbtNkWg6Zw/edit#gid=1360549473
+export enum LTIEnrollmentTypes {
+  AccountAdmin = CanvasRoles.AccountAdmin,
+  SubAccountAdmin = CanvasRoles.SubAccountAdmin,
+  SupportConsultant = CanvasRoles.SupportConsultant,
+  TeacherEnrollment = CanvasRoles.TeacherEnrollment,
+  DesignerEnrollment = CanvasRoles.DesignerEnrollment,
+  TaEnrollment = CanvasRoles.TaEnrollment,
+  Assistant = CanvasRoles.Assistant
+}
```
ssciolla, line 99: This seems like it would work but also feels kinda repetitive. Did you consider using arrays of enum member values instead? For example:
```
AllowedEnrollmentTypes = [CanvasRoles.DesignerEnrollment, CanvasRoles.ObserverEnrollment, CanvasRoles.StudentEnrollment, CanvasRoles.TaEnrollment, CanvasRoles.TeacherEnrollment]
```
You'd have to tweak the enrollment API endpoint validation, since I think it uses an enum right now, but that's not an insurmountable task. Just a thought.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -66,14 +66,38 @@ export interface CanvasCourseSection extends CanvasCourseSectionBase {
   total_students: number
 }
 
-export enum UserEnrollmentType {
+export enum CanvasRoles {
```
ssciolla, line 69: `CanvasRole`?  Minor thing, but I think it's typical to use the singular form for enums. For example, see `Direction` and `LogLevel` in the TypeScript documentation: https://www.typescriptlang.org/docs/handbook/enums.html

ccm_web/server/src/lti/lti.service.ts
ssciolla: Considering this one resolved since you moved the role definitions.

ccm_web/server/src/lti/lti.service.ts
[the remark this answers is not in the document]
ssciolla: So the current code doesn't have this change, right? I was a student but also an admin in my test site, and I was denied access.

ssciolla:
With respect to whether the user should see the plain text error, or be re-directed to the root, I guess what @zqian has now is fine. I don't think this will happen that often. We can reconsider the logic if/when we look at #253.

pushyamig:
@zqian and @ssciolla Can we have this merged soon? since the CCM QA testing will start from this week and it is good idea to get this merged.

ccm_web/server/src/lti/lti.service.ts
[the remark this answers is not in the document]
ssciolla: Hm, maybe I missed a commit or something, this seems to be resolved now.

ccm_web/server/src/canvas/canvas.interfaces.ts
ssciolla: I guess we can postpone this for now. I started looking at it, not impossible, but it may result in the typing on `CanvasEnrollment` becoming a bit relaxed (just a string, instead of an enum member). We can go with this for now and maybe refactor later when an opportunity arises.

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -66,6 +67,17 @@ export class LTIService implements BeforeApplicationShutdown {
       const roles = customLTIVariables.roles as string
       const isRootAdmin = customLTIVariables.is_root_account_admin as boolean
 
+      // check whether the user has at least one of the allowed LTI user roles
+      // otherwise block tool access and show error message
+      const rolesLTI = roles.length > 0 ? roles.split(',') : []
+      const ltiAllowedRoles = rolesLTI.filter(
+        x => Object.entries(LTIEnrollmentTypes).find(
```
ssciolla, line 74: Nit-pick: You could move `Object.entries(LTIEnrollmentTypes)` above  the `filter` and save the result in a `const`. I'm guessing the code would repeat the `entries` creation this way for each element in `rolesLTI`, which is probably a trivial calculation but not one you need to repeat.

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -66,6 +67,17 @@ export class LTIService implements BeforeApplicationShutdown {
...
+          ([key, value]) => value === x)
+      )
+      if (ltiAllowedRoles.length === 0) {
+        return createLaunchErrorResponse(res, 'Your role in this course does not allow access to this tool. If you feel this is in error, please contact 4help@umich.edu.')
```
ssciolla, line 78: Just suggesting you lowercase "your" here the text before ends with a semicolon.
```suggestion
        return createLaunchErrorResponse(res, 'your role in this course does not allow access to this tool. If you feel this is in error, please contact 4help@umich.edu.')
```

ssciolla (approved):
Okay, I think we can proceed with this. I agree with @pushyamig that it'd be nice to have it working for testers.

I did not test all possible role scenarios, but I did test that it works for admins and teachers. I also created a user with all the disallowed roles, and they were not granted access.

pushyamig:
@zqian Can we get this merged sooner?

zqian (commented):
I have made several changes based on the review comments

ccm_web/server/src/lti/lti.service.ts
zqian: text updated.

ccm_web/server/src/lti/lti.service.ts
zqian: updated the code block as suggested.

PR #267, opened 2021-12-08
Add `yup` for basic string validation, implement field validation in `InlineTextEdit` (#244, #276)

### LINKED ISSUE #244: Add client-side validation for course renaming
We might need to create a new issue for validation around empty course name and name >255 character which we did not realize earlier. Adding a Validation will prevent from Canvas API call since canvas don't allow these cases either.

_Originally posted by @pushyamig in https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/241#pullrequestreview-806518945_

ssciolla:
This PR tries to enhance and simplify validation in the front-end by leveraging the [`yup` library](https://github.com/jquense/yup#stringrequiredmessage-string--function-schema) to create schemas (encapsulating multiple tests and associated error messages) and use them to validate strings. The resulting `validation` module is used for field validation in `InlineTextEdit` and to replace identical basic string length validation in a couple other modules. The PR aims to resolve #244 and #276.

Note(s):
- This builds on -- but does not exactly reproduce work -- I did in ROHQ: https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/master/src/assets/src/validation.ts In ROHQ, I fired validation on each keystroke and reported remaining characters; in the interest of speed and simplicity, I did not do that here (could be done later).
- Includes a new package, so testers need to re-build their Docker images.

### DIFF
--- a/ccm_web/client/src/components/BulkSectionCreateValidators.ts
+++ b/ccm_web/client/src/components/BulkSectionCreateValidators.ts
@@ -43,7 +43,7 @@ class SectionNameLengthValidator implements SectionRowsValidator {
       const result = validateString(sectionName, sectionNameSchema)
       if (!result.isValid) {
         invalidations.push({
-          message: result.messages.length > 0 ? result.messages[0] : 'Section name is invalid.',
+          message: result.messages.length > 0 ? result.messages[0] : 'The section name is invalid.',
           rowNumber: row + 2,
           type: InvalidationType.Error
         })
--- a/ccm_web/client/src/components/BulkSectionCreateValidators.ts
+++ b/ccm_web/client/src/components/BulkSectionCreateValidators.ts
@@ -43,7 +43,7 @@ class SectionNameLengthValidator implements SectionRowsValidator {
       const result = validateString(sectionName, sectionNameSchema)
       if (!result.isValid) {
         invalidations.push({
-          message: result.messages.length > 0 ? result.messages[0] : 'The section name is invalid.',
+          message: result.messages.length > 0 ? result.messages[0] : 'Value for the section name is invalid.',
           rowNumber: row + 2,
           type: InvalidationType.Error
         })
--- a/ccm_web/client/src/components/BulkSectionCreateValidators.ts
+++ b/ccm_web/client/src/components/BulkSectionCreateValidators.ts
@@ -1,4 +1,5 @@
 import { InvalidationType } from '../models/models'
+import { sectionNameSchema, validateString } from '../utils/validation'
 
 // For validating row level issues
 interface SectionsRowInvalidation {
@@ -35,24 +36,17 @@ class DuplicateSectionInFileSectionRowsValidator implements SectionRowsValidator
   }
 }
 
-class EmptySectionNameValidator implements SectionRowsValidator {
+class SectionNameLengthValidator implements SectionRowsValidator {
   validate = (sectionNames: string[]): SectionsRowInvalidation[] => {
     const invalidations: SectionsRowInvalidation[] = []
     sectionNames.forEach((sectionName, row) => {
-      if (sectionName.trim().length === 0) {
-        invalidations.push({ message: 'Empty section name is not allowed', rowNumber: row + 2, type: InvalidationType.Error })
-      }
-    })
-    return invalidations
-  }
-}
-
-class SectionNameTooLongValidator implements SectionRowsValidator {
-  validate = (sectionNames: string[]): SectionsRowInvalidation[] => {
-    const invalidations: SectionsRowInvalidation[] = []
-    sectionNames.forEach((sectionName, row) => {
-      if (sectionName.trim().length > 255) {
-        invalidations.push({ message: 'Section name must be 255 characters or less', rowNumber: row + 2, type: InvalidationType.Error })
+      const result = validateString(sectionName, sectionNameSchema)
+      if (!result.isValid) {
+        invalidations.push({
+          message: result.messages.length > 0 ? result.messages[0] : 'Section name is invalid.',
+          rowNumber: row + 2,
+          type: InvalidationType.Error
+        })
       }
     })
     return invalidations
@@ -60,4 +54,4 @@ class SectionNameTooLongValidator implements SectionRowsValidator {
 }
 
 export type { SectionsRowInvalidation, SectionRowsValidator }
-export { InvalidationType, DuplicateSectionInFileSectionRowsValidator, EmptySectionNameValidator, SectionNameTooLongValidator }
+export { InvalidationType, DuplicateSectionInFileSectionRowsValidator, SectionNameLengthValidator }
--- a/ccm_web/client/src/components/InlineTextEdit.tsx
+++ b/ccm_web/client/src/components/InlineTextEdit.tsx
@@ -80,9 +80,11 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
   }
 
   const error = validationResult !== undefined && !validationResult.isValid
-  const helpMessage = validationResult !== undefined && validationResult.messages.length > 0
-    ? validationResult.messages[0]
-    : undefined
+  const helpMessage = validationResult === undefined || validationResult.isValid
+    ? undefined
+    : validationResult.messages.length > 0
+      ? validationResult.messages[0]
+      : 'Value for the field is invalid.'
 
   return (
     <div className={classes.root}>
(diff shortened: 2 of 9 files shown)


ccm_web/client/src/components/BulkSectionCreateValidators.ts
```diff
@@ -35,29 +36,22 @@ class DuplicateSectionInFileSectionRowsValidator implements SectionRowsValidator
...
-class SectionNameTooLongValidator implements SectionRowsValidator {
-  validate = (sectionNames: string[]): SectionsRowInvalidation[] => {
-    const invalidations: SectionsRowInvalidation[] = []
-    sectionNames.forEach((sectionName, row) => {
-      if (sectionName.trim().length > 255) {
-        invalidations.push({ message: 'Section name must be 255 characters or less', rowNumber: row + 2, type: InvalidationType.Error })
+      const result = validateString(sectionName, sectionNameSchema)
+      if (!result.isValid) {
+        invalidations.push({
+          message: result.messages.length > 0 ? result.messages[0] : 'The section name is invalid.',
```
ssciolla, line 46: You'll see a similar pattern to this in a few places. I don't think it's possible that `messages` will be of length 0, but I added this just for thoroughness.

ccm_web/client/src/utils/validation.ts
```diff
@@ -0,0 +1,46 @@
...
+  let transformedValue: string | undefined
+  let isValid = true
+  let messages: string[] = []
+  try {
+    transformedValue = schema.validateSync(value)
+  } catch (error) {
+    if (!(error instanceof ValidationError)) {
+      throw new Error(String(error))
+    } else {
+      transformedValue = error.value as string | undefined
```
ssciolla, line 40: I'd rather not use `as` here to coerce the type, but it's basically just as good as what `yup` is provided, which is `any` or `unknown`. I am very confident that the `transformedValue` (which we're not using yet, but could use for trimming white space or changing case) will be of type `string | undefined`, since that is what the wrapper function accepts and passes to `validateSync`.

pushyamig:
I plan to review this today.

pushyamig:
@ssciolla I feel the code is refactoring/changing the how you validate not just Course rename, but also other features like create sections, 3rd-party gradebook. I feel the Issues should be updated based on what is being changed. This way when we are doing testing then they exactly know what they are testing for

ssciolla:
> @ssciolla I feel the code is refactoring/changing how you validate not just Course rename, but also other features like create sections, 3rd-party gradebook. I feel the Issues should be updated based on what is being changed. This way when we are doing testing then they exactly know what they are testing for

Okay, @pushyamig, how about this? If you think this change (consolidating string validation) is justified and a good change, I will add a new issue about it and attach it to this PR. I'd like your feedback first before I do that though.

ccm_web/client/src/utils/validation.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const createExceededMessage = (fieldName: string, max: number): string => {
+  return `Value for the ${fieldName} must be ${max} characters in length or less.`
+}
+
+const createBlankMessage = (fieldName: string): string => `Value for the ${fieldName} may not be blank.`
+
+const createCanvasEntityNameSchema = (fieldName: string): StringSchema => {
+  return string().required(createBlankMessage(fieldName))
+    .max(255, ({ max }) => createExceededMessage(fieldName, max))
```
pushyamig, line 15: Optional: `maxNameLength = 255` should be made as constant than getting embedded inside a function call

pushyamig:
> @ssciolla I feel the code is refactoring/changing how you validate not just Course rename, but also other features like create sections, 3rd-party gradebook. I feel the Issues should be updated based on what is being changed. This way when we are doing testing then they exactly know what they are testing for
Okay, @pushyamig, how about this? If you think this change (consolidating string validation) is justified and a good change, I will add a new issue about it and attach it to this PR. I'd like your feedback first before I do that though.

Ok.

ccm_web/client/src/utils/validation.ts
```diff
@@ -0,0 +1,46 @@
...
+export const assignmentHeaderSchema = createCanvasEntityNameSchema('assignment header')
+export const sectionNameSchema = createCanvasEntityNameSchema('section name')
+
+// Type validator(s)
+
+export interface ValidationResult {
+  transformedValue: string | undefined
+  isValid: boolean
+  messages: readonly string[]
+}
```
pushyamig, line 28: Observation: I felt you were organizing interfaces top of the file and all the function/methods calls below.

ccm_web/client/src/utils/validation.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function validateString (value: string | undefined, schema: StringSchema): ValidationResult {
+  let transformedValue: string | undefined
+  let isValid = true
+  let messages: string[] = []
+  try {
+    transformedValue = schema.validateSync(value)
+  } catch (error) {
+    if (!(error instanceof ValidationError)) {
```
pushyamig, line 37: Can you tell me a case which trigger this?

ssciolla: Actually, no, this shouldn't happen, but because of TypeScript, I do have to handle it. If it does happen, we'll see an error in the logs.

ccm_web/client/src/components/InlineTextEdit.tsx
```diff
@@ -69,6 +79,11 @@ function InlineTextEdit (props: InlineTextEditProps): JSX.Element {
     }
   }
 
+  const error = validationResult !== undefined && !validationResult.isValid
+  const helpMessage = validationResult !== undefined && validationResult.messages.length > 0
+    ? validationResult.messages[0]
+    : undefined
```
pushyamig, line 85: So in a situation when the `helpMessage` `IF` statement is false, `undefined` is picked up. Why undefined? and why can you have some custom message as you did [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/267/files#diff-eb168987f5c7eaf6d186216d62d3555de5cd7706fd48e8cc06e8d01c8df2d3caR64)

ccm_web/client/src/utils/validation.ts
pushyamig: Yeah... I have done that before. In order to make typescript happy i need to add checks.

ccm_web/client/src/components/InlineTextEdit.tsx
ssciolla: We could add a custom message there. I guess I figured that case is not likely to ever happen, and if it did, the field would still have the red border styling indicating it's an error. I don't mind adding a fallback though if you want, something like "The value for course name is invalid."

pushyamig:
The `yup` library seems to be take out effort of custom code when validating empty value, date or not etc. I feel this util file might grow and it is a good start.

I feel all the schemas could come from different file in case more validation are added. But for now this is OK.

```
export const courseNameSchema = createCanvasEntityNameSchema('course name')
export const assignmentHeaderSchema = createCanvasEntityNameSchema('assignment header')
export const sectionNameSchema = createCanvasEntityNameSchema('section name')

```

I have tested the cases for course rename and create section case. I felt what is done here is logical and improvement. 

I have few comments left but nothing major for getting merged. 

Create issues based on the changes you made. May be add test plan to help QA analyst for testing. 

Over all Good improvement.

pushyamig (approved):

ssciolla:
@pushyamig, see #276 for additional issue documentation.

ccm_web/client/src/components/InlineTextEdit.tsx
pushyamig: Yeah.. lets have some default message as a fallback

ssciolla: Okay, I tweaked this a bit here: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/267/commits/66aa65cb757663131dd04b7fec9335de0d04a4d5 I realized that I wanted to keep the message general since theoretically we may want to use `InlineTextEdit`. Also, since the fallback is an invalid message, I added a check to make sure the result is invalid, and just not show anything if it is. As of now, this is a non-issue, since I only store the validation result if it is invalid. I think what I added should be okay for now, this should be all edge edge case handling.

ssciolla:
@pushyamig, I ignored a couple Codacy issues because I wasn't seeing them locally with the linter, it's probably related to the other weirdness around libraries Chris has seen before, I'm not sure.

ssciolla:
@pushyamig, thanks for the review, I will merge now so I can start using this elsewhere.

PR #269, opened 2021-12-09
Escaping any sort .env files from commiting

pushyamig:
Git ignoring any sort of .env files from repo

Screen Shot 2021-12-09 at 11 56 34 AM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ yarn-error.log*
 .env.development.local
 .env.test.local
 .env.production.local
+*.env
 
 # vscode ignores
 .history


.gitignore
```diff
@@ -32,6 +32,7 @@ yarn-error.log*
 .env.development.local
 .env.test.local
 .env.production.local
+*.env
```
ssciolla, line 35: This is fine. Looks like you could also just modify line 30 to be `*.env`, it'll also ignore `.env`.

ssciolla (approved):
LGTM

PR #275, opened 2021-12-13
Use `encodeURIComponent` with CSV download data for better character handling (#271)

### LINKED ISSUE #271: Strange behavior when generating CSV file with a "#" character
I noticed this bug when doing some testing with "Format Third-Party Gradebook." When an assignment header includes the "#" character, content after the "#" seems to be lost in generated files. I think we may need to add a call to `encodeURIComponent` somewhere in the file output process, so the data can be properly downloaded via a URL; the "#" is probably being interpreted differently. 

https://github.com/mholt/PapaParse/issues/640
----------------------------------------------------
Test plan:
1. for the CSV file try placing # in the assignment name header
2. do the usual workflow with this feature
3. It should trim the file based on the section enrollment and Dowloaded. The assignment name with # should be in tact

ssciolla:
This PR aims to resolve #271.

Resource(s):
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

### DIFF
--- a/ccm_web/client/src/pages/FormatThirdPartyGradebook.tsx
+++ b/ccm_web/client/src/pages/FormatThirdPartyGradebook.tsx
@@ -364,10 +364,12 @@ export default function FormatThirdPartyGradebook (props: FormatThirdPartyGradeb
     processedRecords: GradebookUploadRecord[], assignmentHeader: string, file: File
   ): JSX.Element => {
     const recordsToReview = processedRecords.map((r, i) => ({ rowNumber: i + 2, ...r }))
-    const dataToDownload = 'data:text/csv;charset=utf-8,' + csvParser.createCSV<GradebookUploadRecord>({
-      fields: [...REQUIRED_ORDERED_HEADERS, assignmentHeader],
-      data: processedRecords
-    })
+    const dataToDownload = 'data:text/csv;charset=utf-8,' + encodeURIComponent(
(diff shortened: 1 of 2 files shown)


ccm_web/client/src/pages/FormatThirdPartyGradebook.tsx
```diff
@@ -364,10 +364,12 @@ export default function FormatThirdPartyGradebook (props: FormatThirdPartyGradeb
     processedRecords: GradebookUploadRecord[], assignmentHeader: string, file: File
   ): JSX.Element => {
     const recordsToReview = processedRecords.map((r, i) => ({ rowNumber: i + 2, ...r }))
-    const dataToDownload = 'data:text/csv;charset=utf-8,' + csvParser.createCSV<GradebookUploadRecord>({
-      fields: [...REQUIRED_ORDERED_HEADERS, assignmentHeader],
-      data: processedRecords
-    })
+    const dataToDownload = 'data:text/csv;charset=utf-8,' + encodeURIComponent(
```
pushyamig, line 367: Optional: Besides the change, I feel the text with charset we can take it as constant. So that we just pull it.

pushyamig (approved):

pushyamig:
@ssciolla Why isn't this merged? as this got approval.

ssciolla:
Sure, we can merge it.

PR #282, opened 2021-12-15
#265 #75 creating theme and fixing the error icon color

### LINKED ISSUE #265: Incase of error view use Red exclamation
Blue exclamation point icon used instead of the red exclamation point icon when alerting user. Use Red exclamation instead.

pushyamig:
Fixes #265, #75

1. This issue took a first pass at creating a theme and  getting all the generalized styling at one place. We could add more stuff to theme but this is a bare bone implementations
2. So far error, warn, success and info to users are coming from theme

This issue touches all of the features. Based on the Feedback i can add more here or later.

### DIFF
--- a/ccm_web/client/src/index.tsx
+++ b/ccm_web/client/src/index.tsx
@@ -1,18 +1,22 @@
 import { SnackbarProvider } from 'notistack'
 import React from 'react'
 import ReactDOM from 'react-dom'
-import { BrowserRouter } from 'react-router-dom'
+import { BrowserRouter } from 'react-router-dom' 
+import { ThemeProvider } from '@material-ui/core'
 
+import ccmTheme from './theme'
 import App from './App'
 import './index.css'
 
 ReactDOM.render(
   <React.StrictMode>
+   <ThemeProvider theme={ccmTheme}>
     <SnackbarProvider maxSnack={3}>
       <BrowserRouter>
         <App />
       </BrowserRouter>
     </SnackbarProvider>
+    </ThemeProvider>
   </React.StrictMode>,
   document.getElementById('root')
 )
--- a/ccm_web/client/src/index.tsx
+++ b/ccm_web/client/src/index.tsx
@@ -16,7 +16,7 @@ ReactDOM.render(
         <App />
       </BrowserRouter>
     </SnackbarProvider>
-    </ThemeProvider>
+   </ThemeProvider>
   </React.StrictMode>,
   document.getElementById('root')
 )
--- a/ccm_web/client/src/index.tsx
+++ b/ccm_web/client/src/index.tsx
@@ -1,7 +1,7 @@
 import { SnackbarProvider } from 'notistack'
 import React from 'react'
 import ReactDOM from 'react-dom'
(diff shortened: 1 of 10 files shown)


ccm_web/client/src/index.tsx
```diff
@@ -2,17 +2,21 @@ import { SnackbarProvider } from 'notistack'
...
 
 ReactDOM.render(
   <React.StrictMode>
+   <ThemeProvider theme={ccmTheme}>
     <SnackbarProvider maxSnack={3}>
       <BrowserRouter>
         <App />
       </BrowserRouter>
     </SnackbarProvider>
+    </ThemeProvider>
```
ssciolla, line 19: Looks like the `</ThemeProvider>` closing tag is indented once too far. Unfortunately this doesn't show up on linting, not sure why yet, some JSX rules I think aren't working.

ccm_web/client/src/theme.tsx
```diff
@@ -0,0 +1,20 @@
+import { createTheme } from '@material-ui/core/styles'
+
+const ccmTheme = createTheme({
```
ssciolla, line 3: I think this file doesn't need to be a `.tsx`. There's no JSX here, `.ts` should be fine.

ssciolla (approved):
This looks good! Couple minor comments for you to consider.

I will note that I did a search for "color: " in `client` and there were a number of other hex values, mostly grayscale stuff I think. It would be nice to tidy this up at some point, but it doesn't have to be as part of this issue.

ssciolla:
Also, this doesn't introduce a `primary` color, which we may want to do before we release.

pushyamig:
> Also, this doesn't introduce a primary color, which we may want to do before we release.

I think is fairly simple to do, But what should be that color? Can we do what Myla is doing `#00274C`

ssciolla:
> > Also, this doesn't introduce a primary color, which we may want to do before we release.
> 
> I think is fairly simple to do, But what should be that color? Can we do what Myla is doing `#00274C`

Mmm, maybe check with Janel, but that's probably reasonable. It doesn't need to be done in this PR.

ccm_web/client/src/theme.tsx
pushyamig: Renamed to `.ts`

ccm_web/client/src/index.tsx
pushyamig: I couldn't catch that. Hopefully I fixed it now :)

ccm_web/client/src/theme.ts
```diff
@@ -0,0 +1,23 @@
+import { createTheme } from '@material-ui/core/styles'
+
+const ccmTheme = createTheme({
+  palette: {
+    primary: {
+      main: '#00274C'
```
pushyamig, line 6: I have added Primary color based on Janel suggestion.

ssciolla (approved):
LGTM!

zqian:
Test successful: 

After I changed the palette colors in theme.ts file (primary, error, warning, success, info), I can verify consistent color changes in all CCM subtask screens.

PR #286, opened 2022-01-04
Rewire `ConfirmDialog` button to prevent double-focusing (#256)

### LINKED ISSUE #256: Gradebook formatter: Submit button focuses twice.
Category: Semantics and Structure, impact:  Minor

Issue
Submit button focuses twice.

Impact
Minor cognitive load to screen reader and keyboard only users.

Solution
It does so because it is a button inside a link. Other similar buttons (like in Add UM users) elsewhere do not follow this pattern.

ssciolla:
The PR aims to resolve #256.


pushyamig:
Ok, I don't see the double focus anymore. Fixed!

pushyamig (approved):

PR #287, opened 2022-01-05
#283 Upgrading to Umzug v3

### LINKED ISSUE #283: Upgrade the Umzug to 3.0.0
We are using Umzug library for doing the DB migrations. CCM has been using the beta version of the v3.0.0. Since it had the migration cli support which was very convenient which was not part of v2.0. 

Upgrading to this library will be next step to continues use of the CLI and other great stuff this library bring in.  

https://github.com/sequelize/umzug/releases/tag/v3.0.0

pushyamig:
Fixes #283 

1. Tested both the Prod and local docker compose
2. Migration ran fine, with the existing and new DB setup
3. Ran the migration up/down command seem to work with the existing setup.

### DIFF
--- a/ccm_web/package.json
+++ b/ccm_web/package.json
@@ -59,7 +59,7 @@
     "sequelize": "^6.9.0",
     "sequelize-typescript": "^2.1.1",
     "swagger-ui-express": "^4.1.6",
-    "umzug": "^3.0.0-beta.16",
+    "umzug": "~3.0.0",
     "wait-port": "^0.2.9",
     "winston": "^3.3.3",
     "yup": "^0.32.11"
--- a/ccm_web/package.json
+++ b/ccm_web/package.json
@@ -59,7 +59,7 @@
     "sequelize": "^6.9.0",
     "sequelize-typescript": "^2.1.1",
     "swagger-ui-express": "^4.1.6",
-    "umzug": "~3.0.0",
+    "umzug": "^3.0.0",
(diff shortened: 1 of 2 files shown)


ccm_web/package.json
```diff
@@ -59,7 +59,7 @@
     "sequelize": "^6.9.0",
     "sequelize-typescript": "^2.1.1",
     "swagger-ui-express": "^4.1.6",
-    "umzug": "^3.0.0-beta.16",
+    "umzug": "~3.0.0",
```
ssciolla, line 62: I see you are allowing only patch version (`~`) and not minor version  (`^`) updates (we're doing the latter in most other cases, I think cause that is the default). https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

Do you think we should change all these ranges to `~` before we release?

ssciolla:
I'll run this a bit later today, but looks fine.

ccm_web/package.json
pushyamig: It is a debatable question which way we want to go. I will change it to `^`  and decide on changing it to `~` later

pushyamig: I changed it and ran `npm install` so that `package-lock.json` get updated as well

ssciolla (approved):
Looks good to me! I tested locally -- with new database and with old database, all three commands. I also checked a local `npm install` didn't modify `package-lock.json`.

Issue #285, opened 2021-12-20
Refactor loading and `Backdrop` usages


ssciolla:
We have a functional way of setting up loading display in the UI right now, but it requires repetitive styling and Material UI component use to make it work. Some of this could be extracted into a wrapper component. This may also be a good opportunity though to make sure this is the preferred and accessible way to present a loading state.

pushyamig:
@ssciolla is this going in this release? if yes, can you give me a estimate on this issue

ssciolla:
Kinda depends on whether the loading pattern is deemed to be accessible.

PR #288, opened 2022-01-05
Add "Back" and "Start Again" buttons where relevant (#238)

### LINKED ISSUE #238: Back/Start Again navigation in features?
I kinda snuck in a "Back" button and "Start Again" button to the `FormatThirdPartyGradebook.tsx` feature because I found them useful. How do we feel generally about adding these buttons to all the features (where applicable)? Currently, if the user wants to do something again, they have to go back to the Home page via `Breadcrumbs`, and then click the feature again. Adding these buttons should be relatively straightforward/fast, since we have "Cancel" buttons which do something like a "Back" already in many instances.

Screen Shot 2021-11-09 at 6 44 23 PM

ssciolla:
This PR adds any missing "Back" and "Start Again" buttons to features. In addition, 1) it makes some changes to the Add Non-UM Users feature so as to establish a common leaf component interface for `UserEnrollmentForm` and `MultipleUserEnrollmentWorkflow`, and 2) it ensures section data is reloaded whenever the user navigates to a section selection list, so accurate section and student number data are presented. The PR aims to resolve #238.

### DIFF
--- a/ccm_web/client/src/pages/AddNonUMUsers.tsx
+++ b/ccm_web/client/src/pages/AddNonUMUsers.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useState } from 'react'
 import { Backdrop, CircularProgress, Grid, makeStyles, Typography } from '@material-ui/core'
 
 import * as api from '../api'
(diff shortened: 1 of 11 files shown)


pushyamig:
This PR has conflicts with Main branch.

ccm_web/client/src/pages/MergeSections.tsx
```diff
@@ -209,21 +212,44 @@ function MergeSections (props: CCMComponentProps): JSX.Element {
...
+      <CourseSectionList canUnmerge={isSubAccountAdminOrAccountAdmin()} {...props} />
+      <Grid container className={classes.buttonGroup} justifyContent='flex-start'>
+        <Button
+          variant='outlined'
+          aria-label={`Start ${props.title} again`}
+          onClick={() => setPageState(PageState.SelectSections)}
+        >
+          Start Again
+        </Button>
+      </Grid>
```
pushyamig, line 250: I see that `Start Again` component is used across features. Can we generalize component as `StartAgain` component take props appropriately?

ssciolla: We could certainly, but this was what I was trying to say in the meeting: I'm not sure if it confers enough benefits to do so. For example, compare the current usage of out-of-the-box `Button` above to this:

```
<StartAgainButton
  title={props.title} // Used by aria-label
  onClick={() => setPageState(PageState.SelectSections}
/>
```

So the main benefits are we don't have to repeat `variant='outlined'` or `Start Again`, or use a closing tag, and some of the `aria-label` prep is taken care of. That's something. If you want me to do this, I will, I just think it's only providing incremental utility.

We could also include the `Grid` wrapper, but I use this for multiple different kinds of buttons, so I'd be included to have that be it's own `ButtonGroup` component that accepts a flex setting and children.

ssciolla:
> This PR has conflicts with Main branch.

@pushyamig, conflicts are resolved.

ccm_web/client/src/pages/MergeSections.tsx
pushyamig: I see your point on this. Let me think about it a bit. I will get back to you tomorrow.

pushyamig:
Sorry, I haven't gotten to review this. But I will try for today, if not On Monday.

pushyamig:
I will review it now

ccm_web/client/src/pages/AddUMUsers.tsx
```diff
@@ -92,7 +92,7 @@ function AddUMUsers (props: AddUMUsersProps): JSX.Element {
   const classes = useStyles()
```
pushyamig, line 92: 1. Pick a section( say `section 1`)
2. Upload a CSV file with one valid uniqname (say `studenta`) and others with invalid uniqname (say `testa, testb`) roles as students. 
2. After the submit, it will error saying `[testa, testb]` not valid users and `studenta` is success at this point
3. When you hit `Back` button, then section calls should be fetch again. Since it won't show the newly added student ( with count indicator) nor if a new section is manually created then it won't show it in the list.

ccm_web/client/src/pages/AddNonUMUsers.tsx
```diff
@@ -117,38 +117,35 @@ export default function AddNonUMUsers (props: AddNonUMUsersProps): JSX.Element {
     )
   }
 
-  const resetFeature = (): void => setActivePageState(PageState.SelectInputMethod)
+  const resetFeature = (): void => {
```
pushyamig, line 120: As mentioned [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/288/files#r786158224) Probably this is true here as well since the BE is not integrated, I cannot say it.

ccm_web/client/src/pages/GradebookCanvas.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react'
-import { Box, Grid, Link, makeStyles, Typography } from '@material-ui/core'
```
pushyamig, line 2: For this tool the start again button is functional

ccm_web/client/src/pages/MergeSections.tsx
pushyamig: Ok, Keep it as you designed now. No need to generalize it

ccm_web/client/src/pages/AddUMUsers.tsx
pushyamig: Probably for the features where ever we allow users to pick a section, then using the Back button should fetch latest section.

ccm_web/client/src/pages/MergeSections.tsx
```diff
@@ -37,6 +37,9 @@ const useStyles = makeStyles((theme) => ({
   },
   submitButton: {
```
pushyamig, line 38: Working as expected for MergeSection with StartOver button

pushyamig:
Overall startover button seems be working as expected, but with back button which is mostly used with pick a section features like ( Add users, 3rd-party Gradebook) there is bug that it won't show all the latest sections when back is hit. Workaround is launch of the tool.

ccm_web/client/src/pages/AddUMUsers.tsx
ssciolla: Thanks for flagging this! I have some ideas on how to handle it, I'll work on that today.

ssciolla:
Okay, @pushyamig, I don't know that this is the best way to do things, but it seems to do what you asked me to do. I elected to stay with the current practice of having the loading pane only cover the part of the UI where data will be loaded into or collected from, but I think we may want to revisit that as part of #285.

ccm_web/client/src/pages/AddUMUsers.tsx
ssciolla: As discussed, I have rewired `AddUMUsers.tsx`, `AddNonUMUsers.tsx` (and leaf components), and `FormatThirdPartyGradebook.tsx` so that they fetch the latest section data whenever the stage that uses that data appears, including on "Back" button clicks.

pushyamig:
I will review this today

pushyamig:
I saw this issue #302 appearing, but later i figured you did not merge  latest from main to this branch.

pushyamig (approved):

ssciolla:
Thanks, @pushyamig!

Issue #258, opened 2021-12-02
Footer n-nn of n will not be announced when it changes.


pushyamig:
Location: confirmation screen of Gradebook formatter, Add UM users, Create sections.
Impact: Critical. Type: Semantics and Structure
Issue
Footer n-nn of n will not be announced when it changes.

Impact
Screen reader users will be unaware anything has happened after a paging action.

Solution
Add aria-live="polite" to the container so that changes are announced.

pushyamig:
@ssciolla check with @gsliver on this.

gsilver:
aria-live="polite" for the win

PR #294, opened 2022-01-07
Modify `CustomTable` so table semantics are used by screenreaders (#259)

### LINKED ISSUE #259: Instrumentation of confirmation screen table is problematic
Location: Gradebook formatter, Add UM users, Create sections confirmation screen.
Impact: High/Blocker 
Type: Semantics and Structure

Issue
Instrumentation of confirmation screen table is problematic

Impact
Screen reader users will not get good information of what items have been checked since the value (checked/not checked) does not change on activation.

Keyboard-only users will not be able to make a selection because the items are not focusable.

Since each data row has a role="checkbox" - it is overriding the native semantics of a table row. Impact here is uncertain, but not complying to the DOCTYPE is potentially problematic: user agents (and assistive technology) are not required to figure out what the element "really" is.

Solution
This feels like a framework component, so we are not sure what to recommend, what is feasable. At the very least, ensure that each item is focusable (tabindex="0" rather than tabindex="-1") and manage the state of each row via aria-checked="true/false". Also ensure that there are listeners for both onclick and onkeypress. See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

If this is not a readymade, would seriously consider either rebuilding it so that the semantics of the table are not muddled. Either give the role="checkbox" to an inline element containing the row title or rebuild as a list with inner grids for each item, and give the role="checkbox" to an inline element containing the list item title.

ssciolla:
The PR aims to resolve #259.

### DIFF
--- a/ccm_web/client/src/components/CustomTable.tsx
+++ b/ccm_web/client/src/components/CustomTable.tsx
@@ -51,7 +51,7 @@ function CustomTable<T extends TableEntity> (props: TableProps<T>): JSX.Element 
         <TableBody>
           {tableRows.slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage).map((row) => {
             return (
-              <TableRow hover role="checkbox" tabIndex={-1} key={row.rowNumber}>
+              <TableRow hover key={row.rowNumber}>
                 {columns.map((column) => {
                   const value = row[column.id]
                   return (


ccm_web/client/src/components/CustomTable.tsx
```diff
@@ -51,7 +51,7 @@ function CustomTable<T extends TableEntity> (props: TableProps<T>): JSX.Element
         <TableBody>
           {tableRows.slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage).map((row) => {
             return (
-              <TableRow hover role="checkbox" tabIndex={-1} key={row.rowNumber}>
+              <TableRow hover key={row.rowNumber}>
```
pushyamig, line 54: I have tested the confirmation screen problem that is described in the issue, seems to the fix is implemented as per the recommendation.
But one thing I am not clear is, the table content is not read by screen readers in case of success/failure. Is that a separate not relavent to this fix?

ssciolla: Yeah, I think I want to resolve that as part of #258. We need to be smarter with our use of `aria-live`.

pushyamig: OK, in that case, this issues is resolved to me

pushyamig (approved):

PR #298, opened 2022-01-11
Conditionally enable/disable section creation in `CreateSelectSectionWidget` (#284, #299)

### LINKED ISSUE #284: Create section Widget should not be shown for TA and check for Designer and Assistant
Current Non-UMICH user workflow allows create section, but TA are not allowed to create sections in Canvas so API response is 401. So we should not show Create section widget for roles that canvas don't allow

As per discussed in CCM meeting, limit the section widget access to Admins and Teachers.

### LINKED ISSUE #299: Support Consultant role is not able to unmerge sections nor can search by course name/Uniqname
This is broken at some point later when this #16 got merged

ssciolla:
This PR changes the `CreateSelectSectionWidget` component so that only admins (or equivalent) or teachers can see and use the section creation portion of the component.  The PR aims to resolve #284. @pushyamig also added changes to resolve #299.

### DIFF
--- a/ccm_web/client/src/components/CreateSelectSectionWidget.tsx
+++ b/ccm_web/client/src/components/CreateSelectSectionWidget.tsx
@@ -30,6 +30,8 @@ const useStyles = makeStyles((theme) => ({
 
 interface WidgetWithoutCreateProps {
   readonly canCreate: false
+  course?: never
+  onSectionCreated?: never
 }
 
 interface WidgetWithCreateProps {
--- a/ccm_web/client/src/components/CreateSelectSectionWidget.tsx
+++ b/ccm_web/client/src/components/CreateSelectSectionWidget.tsx
@@ -28,43 +28,59 @@ const useStyles = makeStyles((theme) => ({
   }
 }))
 
-interface CreateSelectSectionWidgetProps {
+interface WidgetWithoutCreateProps {
+  readonly canCreate: false
+}
+
+interface WidgetWithCreateProps {
+  readonly canCreate: true
   course: CanvasCourseBase
+  onSectionCreated: (section: CanvasCourseSection) => void
+}
+
+export type CreateSelectSectionWidgetCreateProps = WidgetWithoutCreateProps | WidgetWithCreateProps
+
+interface CreateSelectSectionWidgetBaseProps {
   sections: CanvasCourseSectionWithCourseName[]
   selectedSection?: CanvasCourseSectionWithCourseName
   setSelectedSection: (section: CanvasCourseSectionWithCourseName) => void
-  onSectionCreated: (section: CanvasCourseSection) => void
 }
+
+type CreateSelectSectionWidgetProps = CreateSelectSectionWidgetBaseProps & CreateSelectSectionWidgetCreateProps
 
 export default function CreateSelectSectionWidget (props: CreateSelectSectionWidgetProps): JSX.Element {
   const classes = useStyles()
   return (
     <>
-      <div className={classes.createSectionContainer}>
-        <div className={`${classes.newSectionHint} ${classes.spacing}`}>
-          <Typography>Create a new section to add users</Typography>
-          <Tooltip className={classes.tooltip} placement='top' title='Enter a distinct name for this section'>
-            <HelpOutline fontSize='small' />
-          </Tooltip>
+    {
+      props.canCreate && (
+        <div className={classes.createSectionContainer}>
+          <div className={`${classes.newSectionHint} ${classes.spacing}`}>
+            <Typography>Create a new section to add users to</Typography>
+            <Tooltip className={classes.tooltip} placement='top' title='Enter a distinct name for this section'>
+              <HelpOutline fontSize='small' />
+            </Tooltip>
+          </div>
+          <div className={classes.createSectionWidget}>
+            <CreateSectionWidget {...props} />
+          </div>
         </div>
-        <div className={classes.createSectionWidget}>
-          <CreateSectionWidget {...props} />
-        </div>
-      </div>
-      <Typography variant='subtitle1' className={classes.spacing}>
-        Or select an existing section to add users to
-      </Typography>
-      <div className={classes.sectionSelectionContainer}>
-        <SectionSelectorWidget
-          height={300}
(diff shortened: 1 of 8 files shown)


ccm_web/client/src/components/CreateSelectSectionWidget.tsx
```diff
@@ -28,43 +28,59 @@ const useStyles = makeStyles((theme) => ({
...
-interface CreateSelectSectionWidgetProps {
+interface WidgetWithoutCreateProps {
+  readonly canCreate: false
+}
+
+interface WidgetWithCreateProps {
+  readonly canCreate: true
   course: CanvasCourseBase
+  onSectionCreated: (section: CanvasCourseSection) => void
+}
```
pushyamig, line 39: Why do you need to 2 interfaces `canCreate: true/false `? Can this be something like below? I see that you are assigning `canCreate`  true or false [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/298/files#diff-c30ee72622169f2c71d21d46e688fde3c6758cd5964030732fdeda72bee0aa3fR130) when calling the Create section wizard.
 I sure there must be good reason why you are doing this. But I might not catching it. 
```
interface WidgetWithoutCreateProps {
  readonly canCreate: boolean
}
```

ccm_web/client/src/pages/MergeSections.tsx
```diff
@@ -127,7 +127,7 @@ function MergeSections (props: CCMComponentProps): JSX.Element {
   }
 
   const isSubAccountAdminOrAccountAdmin = (): boolean => {
-    return isAuthorizedForRoles([RoleEnum['Subaccount admin'], RoleEnum['Account Admin']], props.globals.course.roles, 'MergeSections')
+    return isAuthorizedForRoles([RoleEnum['Subaccount admin'], RoleEnum['Account Admin']], props.globals.course.roles)
```
pushyamig, line 130: Why was these removed? that seems to be right thing to do. But what did we miss?

ccm_web/client/src/models/feature.ts
```diff
@@ -9,12 +9,14 @@ interface FeatureDataProps {
   helpURLEnding: string
 }
 
+const adminRoles = [RoleEnum['Subaccount admin'], RoleEnum['Account Admin'], RoleEnum['Support Consultant']]
+
```
pushyamig, line 13: add type `const adminRole: : RoleEnum[]`?

ccm_web/client/src/models/feature.ts
```diff
@@ -63,10 +65,11 @@ const addNonUMUsersProps: FeatureDataProps = {
   helpURLEnding: '/add-non-um-users.html'
 }
 
-const courseRenameRoles: RoleEnum[] = [RoleEnum['Account Admin'], RoleEnum['Subaccount admin'], RoleEnum['Support Consultant']]
+const courseRenameRoles: RoleEnum[] = adminRoles
+const createSectionRoles: RoleEnum[] = [RoleEnum.Teacher, ...adminRoles]
```
pushyamig, line 69: This can going very top next to `adminRoles` constant

ccm_web/client/src/pages/MergeSections.tsx
ssciolla: The string `'MergeSections'` is not used at all, nor were the other stings. From my perspective, it's not necessary, we could always just add a log message for a particular use case if we needed it while we're developing.

ccm_web/client/src/components/CreateSelectSectionWidget.tsx
ssciolla: I can play around this some more, but this is the main way I know how to have the extra props `course` and `onSectionCreated` not be necessary when `canCreate === false`. We could just pass the props and not use them, but I leaned towards the component only enforcing their presence if they're needed.

ccm_web/client/src/models/feature.ts
ssciolla: TypeScript already knows that is the variable's type, so I guess you are just saying to guard against future changes by adding the type?

ccm_web/client/src/models/feature.ts
```diff
@@ -63,10 +65,11 @@ const addNonUMUsersProps: FeatureDataProps = {
...
 
 export type { FeatureDataProps }
 export {
   mergeSectionProps, formatCanvasGradebookProps, formatThirdPartyGradebookProps,
-  createSectionsProps, addUMUsersProps, addNonUMUsersProps, courseRenameRoles
+  createSectionsProps, addUMUsersProps, addNonUMUsersProps, courseRenameRoles, createSectionRoles
```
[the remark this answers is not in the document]
ssciolla: If it's not being used anywhere yet, I would lean towards not exporting it yet.

ccm_web/client/src/models/feature.ts
pushyamig: Yes.

ccm_web/client/src/models/feature.ts
ssciolla: You took care of this, resolved.

ccm_web/client/src/models/feature.ts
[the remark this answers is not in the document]
ssciolla: You're using it now, so that's okay.

ccm_web/client/src/models/feature.ts
ssciolla: You added these.

ccm_web/client/src/components/CreateSelectSectionWidget.tsx
ssciolla: I think this is a standard pattern. See https://www.benmvp.com/blog/conditional-react-props-typescript/ and https://stackoverflow.com/questions/66917440/typescript-conditional-types-for-props.

pushyamig: Thanks for additional info.

pushyamig (approved):
The change is functional and the create section widget is only given access to Admin and teacher users.

PR #247, opened 2021-11-18
enroll non-UMich users (iss. #221, #169, #252, #291)

lsloan:
Resolves #221.

As well as…

Resolves #169.
Resolves #252.
Resolves #291.

## PR review etiquette requests

To my reviewer friends, I humbly request…

* When making comments about the PR, especially ones requesting changes, please use the "code review" form of comments, not "single comment".  The review style ones allow for direct replies which make threaded, easier to follow conversations.  The single comments do not.
* If you make a review comment requesting that I change something, when I've completed the change (or made a comment reply), I will reply with "🏁" or "✅" when I think the conversation has been resolved.  As the reviewer, I will leave it up to you to decide whether the conversation needs to continue or it's time to click the "Resolve conversation" button.

## To Do

<details>
<summary>Older to do items (click to expand)</summary>

---
* [X] - Resolve build error.
  * Reported by @pushyamig in [comment](#issuecomment-1028991165).
  * Fixed on afternoon of Friday, 04 Feb.
* [x] - Verify user info as admin function appears in Swagger API UI
  * See [comment](#issuecomment-1031903734)
* [x] - user info as admin (`/api/admin/user/:loginId`) function may cause an error
  * @pushyamig reported trouble with this function.  `InvalidTokenInterceptor` may need to be removed from this route, and possibly others that use admin token.  The user’s token shouldn’t be deleted in this case.
  * See [conversation in Slack](https://umich-its-annarbor.slack.com/messages/G0130F688N6/p1644339699163449).
* [x] - Change response structure.
  * Rather than return all account creations together, all invitations together, and all enrollments together in keys specifically for those, organize it by user.  A key for each user ID requested, each of which contains keys for their account creation, invitation, and enrollment.
  * This is to remove the need to parse through the data in the UI.
  * @ssciolla made significant contributions to this to help with a TS problem and to make further changes to the response structure.  See f350a78.
* [x] - Replace invitation service Axios calls with NestJS HttpModule calls
  *  See @ssciolla's [comment](#discussion_r797867138)
* [x]  - Role type checking.  I.e., `@IsIn` decorator for enrollment type in DTO
  * See @ssciolla's [comment](#discussion_r797854903)
* [x]  - Move checks for allowed roles from the service level to controller level
  * See @ssciolla's [comment](#discussion_r797848225)
* [x] - Package lockfile changes
  * See @ssciolla's [comment](#discussion_r797880390) and my earlier [acknowledgement](#issuecomment-1027309637).
[...]

PR #301, opened 2022-01-14
Enroll users to multiple sections with one CSV (#278)

### LINKED ISSUE #278: Add-on to Add-UM user: supports enrolling multiple users to multiple section in one CSV file
This is a feature request when Admins can have one CSV file to add users to multiple sections in a course.

```
SECTION_ID, LOGIN_ID,  ROLE
1233, testa, student
1233, testb, student
1234, testc, ta
1234, testd, teacher
1235, teste, observer
1235, testf, designer
```

Role that user can add is same as how the Add-UM designed. No change there

FE Validation
1. Make sure that sections the is provided in CSV already exists( in other words don't allow user to add other course section that doesn't belong to the course) otherwise throw validation error.
2. All validation rules that applies for enrolling user with single section selection case

We are still going to impose 400 user limit per CSV file. since we will get into rate limiting with canvas, again nothing new.

UX design:
[add_um_user_homepage_multiple_sections.pdf](https://github.com/tl-its-umich-edu/canvas-course-manager-next/files/7824338/add_um_user_homepage_multiple_sections.pdf).

Note: We will follow the error validation  as per the existing CCM design and the UX mock up did not capture with verbal consent and will discuss as needed basis.

ssciolla:
The PR aims to resolve #278.

**To Do List**
- [x] Align `AddUMUsers.tsx` with composite/leaf patterns from `AddNonUMUsers.tsx`
- [x] Replace `StaticContentDownloadLink` with `prepDownloadDataString` + `Link`
- [x] Do some more basic testing
- [x] Create/tidy-up/share test files
- [x] Comment on refactoring opportunities


ssciolla:
@pushyamig, I believe this is functional, so if you'd like to try it and give me feedback, you are welcome to. Keep in mind I did this all rather quickly, so there may be bugs, and I still plan to consider some significant refactoring of `AddUMUsers.tsx`.

ccm_web/client/src/components/BulkEnrollUMUserToSectionsConfirmationTable.tsx
```diff
@@ -0,0 +1,30 @@
...
+const columns: TableHeaderColumnInfoShouldUseMatUIType[] = [
+  { id: 'rowNumber', label: 'Row Number', minWidth: 25 },
+  { id: 'loginId', label: 'Login ID', minWidth: 100 },
+  { id: 'role', label: 'Role', minWidth: 100 },
+  { id: 'sectionId', label: 'Section ID', minWidth: 100 }
+]
+
+function BulkEnrollUMUserSectionsConfirmationTable (props: BulkEnrollUMUserSectionsConfirmationTableProps): JSX.Element {
+  const [page, setPage] = useState<number>(0)
+  const tableRows = props.enrollments
```
ssciolla, line 26: I think some of these patterns could be further refactored and generalized. With a little tinkering, we could move the page state into `CustomTable` and then use that component directly (i.e. prepping `columns` and `tableRows` outside) rather than creating all these separate `{specificName}Table` components. Or, if we like the wrapper layer, we could add a generic to these wrappers and move the columns interface into `CustomTable`.

ccm_web/client/src/components/MultipleSectionEnrollmentWorkflow.tsx
```diff
@@ -0,0 +1,341 @@
...
+    const errors = invalidations.map(
+      (invalidation, i) => <Typography key={i}>{invalidation.message}</Typography>
+    )
+    return (
+      <>
+      {file !== undefined && <CSVFileName file={file} />}
+      <ErrorAlert messages={errors} tryAgain={resetUpload} />
+      </>
+    )
+  }
```
ssciolla, line 131: This is mostly just repeated between CSV features. It would be nice to extract these at some point.

ccm_web/client/src/components/MultipleUserEnrollmentWorkflow.tsx
```diff
@@ -132,8 +135,12 @@ export default function MultipleUserEnrollmentWorkflow (props: MultipleUserEnrol
     if (props.getSectionsError !== undefined) return getSectionsErrorAlert
 
     const canCreate = isAuthorizedForRoles(props.userCourseRoles, createSectionRoles)
+    const onSectionCreated = (section: CanvasCourseSection): void => {
+      setSelectedSection(injectCourseName([section], props.course.name)[0])
+      props.onSectionCreated(section)
+    }
```
ssciolla, line 141: I realized when working on `AddUMUsers.tsx` that I had forgotten to make sure the new section was selected here.

ccm_web/server/src/api/dtos/api.section.enrollment.dto.ts
```diff
@@ -0,0 +1,50 @@
...
+export class SectionEnrollmentDto {
+  @ApiProperty()
+  @IsNotEmpty()
+  @MaxLength(255)
+  loginId: string
+
+  @ApiProperty({ enum: UserEnrollmentType })
+  @IsNotEmpty()
+  @IsIn(Object.values(UserEnrollmentType))
+  type: UserEnrollmentType
```
ssciolla, line 24: Some of these properties are repeated from other DTOs. There appear to be ways to build variations on other DTOs using `mapped-types` (https://docs.nestjs.com/openapi/mapped-types#mapped-types). I think I may leave this the way it is for now, as I think we'll need to reexamine the set of DTOs once PR #247 is completed.

ssciolla:
Okay, I think this is ready for review. I will tidy up my testing files now and upload them to Google Drive. @lsloan, I think @pushyamig can take the lead here, but you may want to take a look, especially at the `server` changes.

pushyamig:
Screen Shot 2022-01-27 at 3 17 21 PM

For the multiple sections use-case should we have 2 downloads? will it be simple for user they download a single file with all the header that is needed for user and they can delete the section name column or while parsing the file the code won't look at the section_name column at all( i think this will work with the way it is coded)

@janelilr @ssciolla Thoughts?? 

Note: Not asking for re-work, just thinking it loud here

ssciolla:
> For the multiple sections use-case should we have 2 downloads? will it be simple for user they download a single file with all the header that is needed for user and they can delete the section name column or while parsing the file the code won't look at the section_name column at all( i think this will work with the way it is coded)
> 

I think it would be confusing to do it the way you are suggesting -- too many steps involved, between removing a column and then possibly rows if they don't want to use all sections. That's my view.

janelilr:
> For the multiple sections use-case should we have 2 downloads? will it be simple for user they download a single file with all the header that is needed for user and they can delete the section name column or while parsing the file the code won't look at the section_name column at all( i think this will work with the way it is coded)

I would suggest keeping the downloads separate. Since the "Download an example" relates to that particular feature, it is consistent with the other features that offer an example. The other download is specific to canvas section IDs and is clear based on the download label.

pushyamig:
> I would suggest keeping the downloads separate. Since the "Download an example" relates to that particular feature, it is consistent with the other features that offer an example. The other download is specific to canvas section IDs and is clear based on the download label.

>I think it would be confusing to do it the way you are suggesting -- too many steps involved, between removing a column and then possibly rows if they don't want to use all sections. That's my view.

OK

ccm_web/client/src/components/BulkEnrollUMUserToSectionsConfirmationTable.tsx
```diff
@@ -0,0 +1,30 @@
...
+  label: string
+  minWidth: number
+  align?: 'left' | 'right' | undefined
+}
+
+const columns: TableHeaderColumnInfoShouldUseMatUIType[] = [
+  { id: 'rowNumber', label: 'Row Number', minWidth: 25 },
+  { id: 'loginId', label: 'Login ID', minWidth: 100 },
+  { id: 'role', label: 'Role', minWidth: 100 },
+  { id: 'sectionId', label: 'Section ID', minWidth: 100 }
```
pushyamig, line 21: I like the flexibility the code is adding when uploading a csv file for single section, it is honoring multiple section csv file  as well. this is better if user does single and multiple often then he can deal with one file :)

ccm_web/server/src/api/dtos/api.section.enrollment.dto.ts
pushyamig: Yeah... the mapped-types seems to keep with reusable and organized but not for this phase of the project. you can create an issue for this and put in the icebox

ccm_web/server/src/api/api.controller.ts
```diff
@@ -90,6 +91,31 @@ export class APIController {
...
+    const invalidSectionIds: number[] = []
+    enrollments.forEach(e => {
+      if (!validSectionIds.includes(e.sectionId) && !invalidSectionIds.includes(e.sectionId)) {
+        invalidSectionIds.push(e.sectionId)
+      }
+    })
+    if (invalidSectionIds.length > 0) {
+      const message = `Course with ID ${courseId} does not have sections with the following ID(s): ${invalidSectionIds.join(', ')}`
+      throw new BadRequestException(message)
+    }
```
pushyamig, line 113: I was trying to see a scenario where the frontend validation missed identifying the course sections vs other course sections, if don't see that is possible. So to me this additional check if all these sections belong to course is redundant. The FE is loading latest section info appropriately.

ccm_web/client/src/components/Accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+  title: string
+  id: string
+  children: React.ReactNode
+}
+
+export default function Accordion (props: AccordionProps): JSX.Element {
+  const classes = useStyles()
+  return (
+    <div className={classes.container}>
+      <MUIAccordion defaultExpanded >
```
pushyamig, line 31: I was happy to see Material UI has the Accordion component we can leverage from

ccm_web/client/src/components/StaticContentDownloadLink.tsx
```diff
@@ -1,16 +0,0 @@
-import { Link } from '@material-ui/core'
-import React from 'react'
-
-interface StaticContentDownloadLinkProps {
-  data: string
-  fileName: string
-  linkText: string
-}
-
-function StaticContentDownloadLink (props: StaticContentDownloadLinkProps): JSX.Element {
```
pushyamig, line 10: Any reason for deleting this re-usable component?

ccm_web/client/src/utils/validation.ts
```diff
@@ -51,3 +62,21 @@ export function validateString (value: string | undefined, schema: StringSchema)
   }
   return { transformedValue, isValid, messages }
 }
+
+export function validateNumberString (value: string | undefined, schema: NumberSchema): ValidationResult<number> {
+  let isValid = true
+  let messages: string[] = []
+
+  const valueToTest = value === undefined ? value : value.trim() === '' ? NaN : Number(value)
```
pushyamig, line 70: in case of `value` undefined or empty the code allows to go to next step which is `schema.validateSync(valueToTest, { strict: true })` 
1. what will happen when undefined is passed to `validateSync`?
2.  can the return statement in this scenario be done earlier, without passing through try-catch block.

ccm_web/client/src/utils/fileUtils.ts
```diff
@@ -1,3 +1,9 @@
+const CSV_LINK_DOWNLOAD_PREFIX = 'data:text/csv;charset=utf-8,'
+
+const prepDownloadDataString = (data: string): string => {
+  return CSV_LINK_DOWNLOAD_PREFIX + encodeURIComponent(data)
```
pushyamig, line 4: Nice!

ccm_web/client/src/utils/enrollmentValidators.ts
```diff
@@ -66,3 +74,34 @@ export class LoginIDRowsValidator implements EnrollmentRowsValidator {
...
+  }
+
+  validate (sectionIds: string[]): EnrollmentInvalidation[] {
+    const invalidations: EnrollmentInvalidation[] = []
+    sectionIds.forEach((sectionId, i) => {
+      const result = validateNumberString(sectionId, sectionIdSchema)
+      if (result.transformedValue === undefined || !result.isValid) {
+        invalidations.push({
+          rowNumber: getRowNumber(i),
+          message: result.messages.length > 0 ? result.messages[0] : 'Value for the section ID was invalid.',
```
pushyamig, line 91: Can this be worded like `Value for the section ID is invalid.`?

ccm_web/server/src/api/dtos/api.section.enrollment.dto.ts
ssciolla: See https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/313

ccm_web/server/src/api/api.controller.ts
ssciolla: I guess I was just trying to be thorough, and protect the API endpoint from the same potential validation problem. I see your perspective though, the Canvas token will limit the user to editing their sections (across all their courses). I can remove it.

ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
```diff
@@ -0,0 +1,334 @@
...
+    )
+  }
+
+  const getUploadContent = (): JSX.Element => {
+    if (schemaInvalidations !== undefined) return renderSchemaInvalidations(schemaInvalidations)
+    if (rowInvalidations !== undefined) return renderRowValidationErrors(rowInvalidations)
+
+    const fileData =
+    `${REQUIRED_ENROLLMENT_HEADERS.join(',')}
+    student,studenta
```
pushyamig, line 216: `studenta` is real test user we should be giving that in the sample csv file. Change it like userf

ccm_web/client/src/utils/validation.ts
ssciolla: It could, but if you look at `createCanvasIdentifierSchema` factory function above I used to create `sectionIdSchema`, I'm setting `typeError` and `required`. These should be the messages when the value is `NaN` or `undefined`, respectively. I wanted `yup` to handle the creation of error messages, rather than pass more information into `validateNumberString`.

I know this looks somewhat convoluted. Basically my problem was I had to use `{ strict: true }`  because `yup` uses `parseFloat` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat) for value transformation. `parseFloat` has some weird behavior; basically it lops off non-numeric stuff at the end, where I wanted it to consider that an error. `Number` also has issues, since it treats an empty string as `0`. Hopefully this helps explain my decisions. I'm happy to consider alternatives if you have ideas.

pushyamig:
Over all this PR functionally looking good. I have few suggestions mostly removing the redundant section check in the BE if that issue is resolved then this PR is ready to merge.

ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
pushyamig: I mean to say we shouldn't be giving that user since the user may accidentally adding to course, causing some noise.

ssciolla: Sure, this has been in the feature since it was first started I think (before the expansion). How is this? https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/301/commits/ac2422b35af39bada44d0899ca7d43200974c1f2

ccm_web/client/src/utils/enrollmentValidators.ts
ssciolla: Sure, done. https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/301/commits/db94057bb6ddf0a8b638cb992275fde9d709d4f7 I also removed "the" since the other default messages don't have that.

ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
pushyamig: the change looks good. Change what you can here, not to worry of changing in other case, we can deal with that once we step into that code.

ccm_web/client/src/utils/validation.ts
pushyamig: Thanks for explanation. We can keep it as is, I don't see a huge problem since this is not making any API call things as such, so the amount of time it takes stepping into that code in case of `undefined` and `NaN` is very little not worth our time now. This would be a case of keeping the code simple and readable.

ccm_web/client/src/components/StaticContentDownloadLink.tsx
pushyamig: OK

ccm_web/server/src/api/api.controller.ts
pushyamig: Yes... Scoped token approach is turning out as good choice for most of the time

ssciolla: Okay, @pushyamig, see here: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/301/commits/5c714c6571ba715af719191b5d27581061debced

ccm_web/client/src/components/BulkEnrollUMUserToSectionsConfirmationTable.tsx
ssciolla: https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/314

ccm_web/client/src/components/MultipleSectionEnrollmentWorkflow.tsx
ssciolla: #315

pushyamig (approved):
Nice work! this feature/Add-on was added as you saw the disparity in the feature delivery from old vs new CCM when stakeholder flagged.  Proactively convinced the team for adding this and delivered in reasonable amount of time.

ssciolla:
Thanks, @pushyamig, I'm gonna merge this so we can get folks to start testing.

PR #303, opened 2022-01-14
Remove `colSpan` from `TablePagination` usages (#302)

### LINKED ISSUE #302: Table pagination not right-aligned when four columns wide
This is a cosmetic issue, but with our table implementation(s), when the data rows extend past three columns, the pagination buttons and label are no longer fully right-aligned.

Screen Shot 2022-01-14 at 2 40 24 PM

ssciolla:
The PR aims to resolve #302.

### DIFF
--- a/ccm_web/client/src/components/CustomTable.tsx
+++ b/ccm_web/client/src/components/CustomTable.tsx
@@ -73,7 +73,6 @@ function CustomTable<T extends TableEntity> (props: TableProps<T>): JSX.Element 
           <TableRow>
             <TablePagination
               rowsPerPageOptions={[rowsPerPage]}
-              colSpan={3}
               count={tableRows.length}
               rowsPerPage={rowsPerPage}
               page={page}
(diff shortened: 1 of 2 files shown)


ccm_web/client/src/components/CustomTable.tsx
```diff
@@ -73,7 +73,6 @@ function CustomTable<T extends TableEntity> (props: TableProps<T>): JSX.Element
           <TableRow>
             <TablePagination
               rowsPerPageOptions={[rowsPerPage]}
-              colSpan={3}
```
ssciolla, line 76: This was not an issue previously, because we didn't have any tables greater than 3 columns wide. `colSpan` does not seem to be present in the documentation for Material UI 4 or 5, but it still works? https://v4.mui.com/api/table-pagination/#tablepagination-api

Anyways, removing it seems to trigger a default that works in all uses cases.

pushyamig:
I did notice this while reviewing, but It slipped out of my minds when reviewing #247 . I did not test this, but logic make sense to me.

pushyamig (approved):

PR #306, opened 2022-01-20
Only intercept token-related `401` errors (#305)

### LINKED ISSUE #305: `InvalidTokenInterceptor` should not intercept all `401` responses from Canvas
This problem arose during discussions with the penetration testing team. Currently, the `InvalidTokenInterceptor` -- which was designed originally to delete invalid Canvas tokens (see #101) -- is catching all HTTP exceptions with status code `401` originating from Canvas.

I believe this was done originally because it seemed a more durable and programmatically safe way to identify the responses (rather than comparing string body messages which Canvas may opt to change). In addition, the user would have to hit the API directly to produce a `401` not related to an invalid token (for example, the UI won't let them request a different course that they don't have access to, but a direct API call would produce a `401`).

However, this became a problem for penetration testers, because they are hitting the API directly with lots of parameter values, and the token deletion and redirects block from a scan from continuing. Thus, for at least some use cases, we need a way to differentiate between `401` errors. We could allow the interceptor to be turned off or on, but we could also process the responses and only delete tokens when the response body has "Invalid access token." or "Insufficent scopes on access token.".

ssciolla:
The PR aims to resolve #305.

### DIFF
--- a/ccm_web/server/src/api/api.controller.ts
+++ b/ccm_web/server/src/api/api.controller.ts
@@ -7,6 +7,7 @@ import { ApiQuery, ApiSecurity } from '@nestjs/swagger'
 
 import { Globals, isAPIErrorData } from './api.interfaces'
 import { APIService } from './api.service'
+import { InvalidTokenInterceptor } from './invalid.token.interceptor'
 import { CourseNameDto } from './dtos/api.course.name.dto'
 import { CreateSectionsDto } from './dtos/api.create.sections.dto'
 import { GetSectionsAdminQueryDto } from './dtos/api.get.sections.admin.dto'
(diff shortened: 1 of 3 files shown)


pushyamig (approved):
Thanks for the Quick solution, in the midst of supporting the Penetration testing. 

A synopsis from my testing and understanding 
 
The problem lies in differentiating 3 different 401 Un authorized response for canvas 3 different scenarios  
1. When user revoked the API permission from his Canvas profile setting. Give a 'Invalid access token
2. When scope changes as part of API developer key and accessing existing token might result in "Insufficient scopes on access token
3. When user try to access an canvas resource (courses/section) that he doesn't access to (CCM doesn't allow that) giving "user not authorized to perform that action"

The proposed solution is resorting to checking the error msg since canvas is not differentiating with different status code or other header options. 

In case of err msg change by canvas ( for some odd reasons) for scenarios 1,2 they might be end up denying access in ccm. Further investigation look at the Canvas code base by @jonespm `"Invalid toke access" has been there for 8+ years and "Insufficient scopes" for 4+ years.` So we can patch a release to fix this and not have it as configuration option
So it is safe for now to make a check for invalid access token and insufficient scopes as configuration option where token renewal is needed.  With upcoming release we should always test scenarios 1,2  to see if the error message for 401 is changed or not.

PR #308, opened 2022-01-24
Disable `cors` in `LTIService` (#307)

### LINKED ISSUE #307: Disable `cors` in `lti/` URL space
The penetration testing team asked about a lenient CORS setting on `lti/*` routes. Based on the `ltijs` documentation and source code, this is a default, but it can be turned off. This is mentioned in #170. `ltijs` does seem to be doing some blocking of requests from unregistered platforms, so there seems to be some protection in place here. However, in my testing, I observed no disruption to the LTI handshake when setting the available `ltijs` setting to `false`. I think it makes sense to turn this off as an extra precuation, provided it isn't needed.

Resource(s):
- https://cvmcosta.me/ltijs/#/provider?id=cors
- https://github.com/Cvmcosta/ltijs/blob/c80cbd40cbbe18e39d7583d04bd3c289c880fff3/src/Utils/Server.js#L37-L45
- https://github.com/expressjs/cors
- https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

ssciolla:
The PR aims to resolve #307.

Notes:

I'm requesting reviews from the entire development team, as this is a minuscule code change and others may have expertise regarding LTI and web security to contribute. See issue for details.

### DIFF
--- a/ccm_web/server/src/lti/lti.service.ts
+++ b/ccm_web/server/src/lti/lti.service.ts
@@ -50,7 +50,8 @@ export class LTIService implements BeforeApplicationShutdown {
         keysetRoute: '/keys',
         // Set secure to true if the testing platform is in a different domain and https is being used
         // Set sameSite to 'None' if the testing platform is in a different domain and https is being used
-        cookies: { secure: true, sameSite: 'None' }
+        cookies: { secure: true, sameSite: 'None' },
+        cors: false
       }
     )
 


pushyamig (approved):

ssciolla:
Merging, so we can test with a deployment.

pushyamig:
I will QA this

Issue #103, opened 2021-06-09
Deal with CSP violation error when not launching from Canvas Vanity URL


pushyamig:
So we really don't know if we are going to implement Canvas Vanity URL in Prod and we are getting some errors related to Violation of Content Secure Policy(CSP) when launching CCM, We might need to deal with this in our NodeJS app.

Some research is needed how to do this in NodeJs. You can replicate the below error in the Canvas beta instance
![Screen Shot 2021-06-09 at 9 44 21 AM](https://user-images.githubusercontent.com/8579775/121366867-d1fe7380-c907-11eb-936b-3f5b40e3fad2.png)

pushyamig:
1. x-frame option = allow all

jonespm:
You used to be able to use X-Frame-Options ALLOW-FROM (and in some browsers ALLOWALL) to do this easily. This could be done with Helmet `helmet.frameguard(options)`. https://helmetjs.github.io/

However this is [marked as deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) by CSP frame-ancestors.

It looks like there's a helmet-csp package. https://www.npmjs.com/package/helmet-csp

I'm MyLA we're passing in instructure.com into `frame-src`. Many CSP policies have a fallback.

pushyamig:
This is Non-issue. Thanks @jonespm further insights on this and analyzing this is something coming from Canvas not from CCM. 

it's just a [new feature](https://umich.beta.instructure.com/accounts/1/settings#tab-security) that is only an issue on beta. The security tab is only in beta, in test you can enable it canvas test and you don't see this option in Prod at all

comments from @jonespm 
For some reason this domain isn't added by default but they say it should be. "Canvas and Instructure domains are included automatically and do not count against your 50 domain limit."

The policy is "report only" so it only will display the error but have no effect
we added  the beta domain to allowed list. You might see this when Prod sync beta over the beta. 

But this cannot be reproduce locally, test, umich dev. Now after adding beta canvas domain we don't see this error any more. Again not coming from CCM but from Canvas

zqian:
As Pushyami mentioned above, [The Security tab only displays in Account Settings](https://community.canvaslms.com/t5/Admin-Guide/How-do-I-manage-the-Content-Security-Policy-for-an-account/ta-p/149) if admin have enabled the Content Security Policy feature preview.

[...]

Issue #312, opened 2022-01-27
Error when launching CCM in Safari


jonespm:
When launching CCM from Safari on Canvas beta this error returns

`{"status":401,"error":"Unauthorized","details":{"description":"Error validating ltik or IdToken","message":"MISSING_VALIDATION_COOKIE"}}`

I believe this is a common issue with Safari if a user has the default Privacy option "Prevent cross-site tracking" checked. I saw it was [brought up as an issue in the library](https://github.com/Cvmcosta/ltijs/issues/53) but there was no resolution. The resultion by most other vendors is either turn this setting off or switch to a vanity URL, which we've been in the process of doing but isn't complete.

I tested on Canvas Test which does have a vanity URL and it works fine. However this will be the same in prod as it is in beta.

Our python apps that use pylti13 prompt to open a new window. That's because there's [javascript to check for this](https://github.com/dmitry-viskov/pylti1.3/blob/master/pylti1p3/cookies_allowed_check.py) in the library. Since the majority of our users are on Safari this might be something we need to address and at least prompt users to open in a new window or disable this option. 

IMS LTI team is working on a solution to avoid this by using postMessage rather than cookies but this isn't yet in the spec.

pushyamig:
@melindakraft , Can you suggest the user message telling about the error?

melindakraft:
`It seems your browser is blocking 3rd party cookies which are required for this application. To resolve this issue, please update your settings for <a href="https://support.google.com/chrome/answer/95647?hl=en" target="_blank"> Chrome</a>, <a href="https://support.apple.com/guide/safari/manage-cookies-and-website-data-sfri11471/12.0/mac/10.14" target="_blank"> Safari for MacOS</a>, <a href="https://www.whatismybrowser.com/guides/how-to-enable-cookies/safari-iphone" target="_blank"> Safari for iOS</a>, <a href="https://support.mozilla.org/en-US/kb/enhanced-tracking-protection-firefox-desktop?redirectslug=enable-and-disable-cookies-website-preferences&redirectlocale=en-US" target="_blank"> Firefox</a>, and <a href="https://www.whatismybrowser.com/guides/how-to-enable-cookies/edge" target="_blank"> Microsoft Edge</a> to allow 3rd party cookies.`

ssciolla:
@melindakraft, @pushyamig, working on this now. A thought occurred to me: should all these browser specific links be in the help docs or on this page? I could see arguments for doing it either way.

[...]

Issue #209, opened 2021-09-30
Tweak language shown to unauthorized student users and unauthenticated users


ssciolla:
**Background:** 
Currently, if you navigate to the application's URL directly, the user is shown "You were not properly authenticated to the application." @melindakraft thinks this language should be tweaked to be a bit more user-friendly and link to documentation.

Similarly, if an authenticated student user happens to launch the tool (going to the external tools URL directly), they should see a similar user-friendly unauthorized message and a link to documentation.

@melindakraft will think about and provide what the messages should say.

**Text changes:**

Access Denied.

To use U-M Course Manager you must have a course management role in this course.

If you believe this message is in error, please contact 4help@umich.edu(https://its.umich.edu/help)

melindakraft:
How about something like:

Access Denied. 

To use U-M Course Manager you must have a course management role in this course. 

If you believe this message is in error, please contact 4help@umich.edu (include link to ITS help page).

ssciolla:
@pushyamig, can you assess whether what to do about this issue now that @zqian added that code to block these users at the level of the LTI service. Maybe we just want to tweak the language there, or we could redirect them to the root as we suggested already, and change the "You were not properly authenticated page..."

pushyamig:
The ITS help link to use is https://its.umich.edu/help. as suggested by @melindakraft

PR #316, opened 2022-01-31
Allow for client-side routing, set up `/access-denied` (#48, #209)

### LINKED ISSUE #48: Allow for direct routing on client side
There is currently an issue with the server configuration where you cannot navigate directly to a specific page after `/` directly by URL. I think this has something to do with how `webpack-dev-middleware` works (see issue below).

https://stackoverflow.com/questions/43921770/webpack-dev-middleware-pass-through-for-all-routes

ssciolla:
The PR aims to resolve #48 and #209. This is somewhat experimental, but I believe it works well. I think it could help us handle some other issues, such as #253 and #312. It also is related to the closed issue #103.

Note(s):
- Switches from `app.useStaticAssets` to using `ServeStaticModule`, which is designed for single-page-applications.
- Because `ServeStaticModule` is now registered in `app.module.ts`, `helmet` and its Content Security Policy (CSP) settings now apply to the static assets served. This is likely a good thing, but means we need to do more CSP configuration (included here). Without them, the iframe was getting blocked in the Canvas test instance.
- Adds a higher-level instance of `Switch` in `client/index.tsx`, so we can have public routes for error messages and redirect effectively to them from the `server` and `client` (see changes to `lti.service.ts` and `App.tsx`).

### DIFF
--- a/ccm_web/client/src/index.tsx
+++ b/ccm_web/client/src/index.tsx
@@ -1,11 +1,12 @@
 import { SnackbarProvider } from 'notistack'
 import React from 'react'
 import ReactDOM from 'react-dom'
-import { BrowserRouter } from 'react-router-dom'
+import { BrowserRouter, Route, Switch } from 'react-router-dom'
 import { ThemeProvider } from '@material-ui/core'
 
 import ccmTheme from './theme'
 import App from './App'
+import AccessDenied from './pages/AccessDenied'
 import './index.css'
 
 ReactDOM.render(
@@ -13,7 +14,12 @@ ReactDOM.render(
    <ThemeProvider theme={ccmTheme}>
     <SnackbarProvider maxSnack={3}>
       <BrowserRouter>
-        <App />
+        <Switch>
+          <Route exact={true} path='/access-denied'>
+            <AccessDenied forHelpLink='https://its.umich.edu/help' />
+          </Route>
+          <Route><App /></Route>
+        </Switch>
       </BrowserRouter>
     </SnackbarProvider>
    </ThemeProvider>
--- a/ccm_web/client/src/index.tsx
+++ b/ccm_web/client/src/index.tsx
@@ -4,10 +4,10 @@ import ReactDOM from 'react-dom'
 import { BrowserRouter, Route, Switch } from 'react-router-dom'
 import { ThemeProvider } from '@material-ui/core'
 
-import ccmTheme from './theme'
 import App from './App'
-import AccessDenied from './pages/AccessDenied'
 import './index.css'
+import ccmTheme from './theme'
+import AccessDenied from './pages/AccessDenied'
 
 ReactDOM.render(
   <React.StrictMode>
--- a/ccm_web/client/src/index.tsx
+++ b/ccm_web/client/src/index.tsx
@@ -16,7 +16,7 @@ ReactDOM.render(
       <BrowserRouter>
         <Switch>
           <Route exact={true} path='/access-denied'>
-            <AccessDenied forHelpLink='https://its.umich.edu/help' />
+            <AccessDenied email='4help@umich.edu' helpLink='https://its.umich.edu/help' />
           </Route>
           <Route><App /></Route>
         </Switch>
--- a/ccm_web/client/src/pages/AccessDenied.tsx
+++ b/ccm_web/client/src/pages/AccessDenied.tsx
@@ -1,8 +1,7 @@
 import React from 'react'
+import { Link, makeStyles, Typography } from '@material-ui/core'
 
 import InlineErrorAlert from '../components/InlineErrorAlert'
-
-import { Link, makeStyles, Typography } from '@material-ui/core'
 
 const useStyles = makeStyles((theme) => ({
   container: {
--- a/ccm_web/client/src/pages/AccessDenied.tsx
+++ b/ccm_web/client/src/pages/AccessDenied.tsx
@@ -7,9 +7,6 @@ import { Link, makeStyles, Typography } from '@material-ui/core'
 const useStyles = makeStyles((theme) => ({
   container: {
     margin: theme.spacing(2)
-  },
-  link: {
-    textDecoration: 'underline'
   }
 }))
 
@@ -20,9 +17,9 @@ interface AccessDeniedProps {
 
 export default function AccessDenied (props: AccessDeniedProps): JSX.Element {
   const classes = useStyles()
-  const emailLink = <Link className={classes.link} href={`mailto:${props.email}`}>{props.email}</Link>
+  const emailLink = <Link href={`mailto:${props.email}`}>{props.email}</Link>
   const forHelpLink = (
-    <Link className={classes.link} href={props.helpLink} target='_blank' rel='noopener'>
+    <Link href={props.helpLink} target='_blank' rel='noopener'>
       ITS Help Page
     </Link>
   )
--- a/ccm_web/client/src/pages/AccessDenied.tsx
+++ b/ccm_web/client/src/pages/AccessDenied.tsx
@@ -1,0 +1,36 @@
+import React from 'react'
+
+import InlineErrorAlert from '../components/InlineErrorAlert'
+
+import { Link, makeStyles, Typography } from '@material-ui/core'
+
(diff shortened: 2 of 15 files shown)


ccm_web/client/src/index.tsx
```diff
@@ -1,19 +1,25 @@
...
 
 ReactDOM.render(
   <React.StrictMode>
    <ThemeProvider theme={ccmTheme}>
     <SnackbarProvider maxSnack={3}>
       <BrowserRouter>
-        <App />
+        <Switch>
+          <Route exact={true} path='/access-denied'>
+            <AccessDenied email='4help@umich.edu' helpLink='https://its.umich.edu/help' />
```
ssciolla, line 19: Is it okay to have these values be hard-coded? We could put them in the configuration, but I'm not sure how we would actually send to them to the client, since the user isn't authenticated and doesn't have access to `globals`. We'd need a public API endpoint with public data, but maybe that's more of a hassle than it's worth. Maybe it's possible to put the data in `index.html` and grab it (like we do in our Django projects), but we don't have a mechanism for putting configuration variables in that file yet.

ssciolla: Or maybe something like this `DefinePlugin` would work for setting a couple public variables. https://webpack.js.org/plugins/define-plugin/

ccm_web/package.json
```diff
@@ -41,7 +42,7 @@
     "express": "^4.17.1",
     "express-session": "^1.17.2",
     "got": "^11.8.2",
-    "helmet": "^4.6.0",
+    "helmet": "^5.0.2",
```
ssciolla, line 45: I chose to adopt the latest major version of `helmet` here, as I started to get confused when looking at the documentation differences between versions. Looks like some good changes to defaults were made in the latest releases: https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md

ccm_web/client/src/index.tsx
ssciolla: Let me know what you think. I can work on this now or as a separate issue. `DefinePlugin` seems reasonable, but would have to try using `config.ts` in the `webpack` artifacts.

pushyamig:
@ssciolla one thing I noticed is when you open CCM in it's own window and try some random URL, it not show access denied page, instead react code is loading and i see some error with Breadcrumbs missing data. I *feel* other than the routes described in the `FeatureUIData.tsx` should be routed to default no-page found

pushyamig:
May Be in the code below you might need add a default route? 
```
<Breadcrumbs {...{ features, pathnames }} />
      <Switch>
        <Route exact={true} path='/'>
          <Home globals={globals} course={course} setCourse={setCourse} getCourseError={getCourseError} />
        </Route>
        {features.map(feature => {
          return (
            <Route key={feature.data.id} path={feature.route}>
              <feature.component
                globals={globals}
                course={course}
                title={feature.data.title}
                helpURLEnding={feature.data.helpURLEnding}
              />
            </Route>
          )
        })}
        <Route><div><em>Under Construction</em></div></Route>
      </Switch>
```

ccm_web/client/src/index.tsx
pushyamig: I feel it is ok to keep this hard-coded until we see a possible collaboration with other schools interested in using CCM for their schools. Since lot of ourText still has `U-M` in the UI. Hope this will convince you not to worry about this now. You can create an issue for later, definitely something interesting to solve how to load static links based on configuration.

ssciolla:
> May Be in the code below you might need add a default route?

@pushyamig, nice catch! This is a bug, there are stack traces related to `Breadcrumbs` working on it now. I'll also add a "nothing here" page.

ssciolla:
@pushyamig, okay, this should be fixed. Please test again.

ccm_web/client/src/pages/AccessDenied.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export default function AccessDenied (props: AccessDeniedProps): JSX.Element {
+  const classes = useStyles()
+  const emailLink = <Link href={`mailto:${props.email}`}>{props.email}</Link>
+  const forHelpLink = (
+    <Link href={props.helpLink} target='_blank' rel='noopener'>
+      ITS Help Page
+    </Link>
+  )
```
pushyamig, line 24: What is the reason for having the email and link passed as props? I feel at some level you might be thinking to get them as configuration?

ssciolla: You had one of them as a prop I think, so I followed that pattern. At some point we may want to make it configurable, so I guess that prepares us a little for it.

ccm_web/server/src/app.module.ts
```diff
@@ -48,8 +64,18 @@ const logger = baseLogger.child({ filePath: __filename })
...
+  constructor (private readonly configService: ConfigService<Config, true>) {
+    this.frameDomain = this.configService.get('server.frameDomain', { infer: true })
+  }
+
   configure (consumer: MiddlewareConsumer): void {
-    consumer.apply(helmet())
+    consumer.apply(helmet({
+      contentSecurityPolicy: {
+        directives: { 'frame-ancestors': [this.frameDomain] }
+      }
```
pushyamig, line 77: Nice!

ccm_web/package.json
```diff
@@ -30,6 +30,7 @@
     "@nestjs/passport": "^8.0.1",
     "@nestjs/platform-express": "^8.2.0",
     "@nestjs/sequelize": "^8.0.0",
+    "@nestjs/serve-static": "^2.2.2",
```
pushyamig, line 33: Looks like this library defaults settings for caching is enabled to true. https://github.com/nestjs/serve-static/blob/master/lib/interfaces/serve-static-options.interface.ts

ccm_web/server/src/app.module.ts
```diff
@@ -24,6 +27,19 @@ const logger = baseLogger.child({ filePath: __filename })
...
+      imports: [ConfigModule],
+      inject: [ConfigService],
+      useFactory: (configService: ConfigService<Config, true>) => ([{
+        rootPath: (
+          path.join(
+            path.join(__dirname, '..', '..'),
+            configService.get('server.isDev', { infer: true }) ? path.join('dist', 'client') : 'client'
+          )
+        ),
+        exclude: ['/api*', '/canvas*', '/lti*', '/auth*']
```
pushyamig, line 40: does that mean that if we have a static react route with any of the names above might be conflicting? Like if we have react page as `https:/dff.ngrok.com/canvas-section` will be blocked. 

What you have done is cool and seems to working, just trying to wrap my head around with this PR

ssciolla: This is a good point, hm...

Yes, I think `canvas-section` would conflict. I think for `api`, `canvas`, and `auth`, it's probably safe to add `/` at the end of each of them. I'll have to check whether `/lti/*` introduces any weird behavior.

ssciolla: It also looks like if I remove this entirely, the existing API routes still work, they just all fall through to the Not Found page. Maybe this is fine behavior. Currently anything not defined after these route beginnings result in a standard JSON `express` error.

pushyamig: OK

pushyamig (approved):
The changes in this PR are progressive and working as expected. We just need to be little careful when we chose our **new** client-routes and not conflict with. lti,api,auth.

ccm_web/server/src/app.module.ts
ssciolla: I decided to add the `/` at the end of each of them so something like `canvas-sections` or `lti-error` would still fallthrough. I think excluding the routes is still a good practice, since anything after these prefixes **should** be an API route and show the appropriate type error instead of loading the static.

Interestingly, with the latest commit (), a route like `/canvas` would still be excluded. It looks like this is because there is an appended slash when the `ServeStaticModule` check whether a route is excluded: https://github.com/nestjs/serve-static/blob/f9c7be3365359d86395f33b6c8c7d8aa1269de9d/lib/utils/is-route-excluded.util.ts#L12-L16

PR #317, opened 2022-01-31
Add `aria-live` to `TablePagination`, remove it from `AddNonUMUsers` container (#258)

### LINKED ISSUE #258: Footer n-nn of n will not be announced when it changes.

ssciolla:
Per feedback received on #258 to use `aria-live` more granularly, this PR proposes two changes to `aria-live` usage in the application:

1) add `aria-live='polite'` as instructed to the container for the table pagination footer where the displayed rows are reported
2) remove `aria-live='polite'` that was added provisionally to the `AddNonUMUsers` container, as high-level usages like this are probably noisy and discouraged.

The PR seeks to resolve #258.

### DIFF
--- a/ccm_web/client/src/components/CustomTable.tsx
+++ b/ccm_web/client/src/components/CustomTable.tsx
@@ -82,6 +82,7 @@ function CustomTable<T extends TableEntity> (props: TableProps<T>): JSX.Element 
               }}
               onPageChange={handleChangePage}
               ActionsComponent={TablePaginationActions}
+              aria-live='polite'
             />
           </TableRow>
         </TableFooter>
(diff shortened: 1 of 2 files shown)


pushyamig:
@ssciolla I can see if you are paginating table, voice over is saying "your are on next page of the table or something similar". But it is still not reading the content of the table. should it?

ssciolla:
> @ssciolla I can see if you are paginating table, voice over is saying "your are on next page of the table or something similar". But it is still not reading the content of the table. should it?

That's probably a question for @gsilver. I would think that would have been asked for in the issue if it was a problem. We could put `aria-live='polite'` on the table container, but I think the recommendation was to keep it granular. Maybe the thought is that if the page range is updated, the user will know to have table re-read using a command. I'm not sure.

pushyamig (approved):
Accessibility Team think this is working as expected.

PR #309, opened 2022-01-26
I209 access denied msg

pushyamig:
Fixes #209

PR #321, opened 2022-02-02
#320 Disallow client to store api response

### LINKED ISSUE #320: Add a Cache Control header for not caching the API response
It is recommendation that don't cache any sort of API response on the Client side. So Adding a Cache-Control header is way to go about it. 

Looking at all the Cache Control header options `Cache-Control: no-store` fits the requirement.  While it is easily deceptive to think `Cache-Control: no-cache` it won't caching but what that tell the browser is cache the response in the local store and revalidate with server if this value changed before presenting to the user

https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#the_cache-control_header

pushyamig:
Fixes #320 

This PR is built on the dev-ccm instance. If you like to check it out here https://canvas-test.it.umich.edu/courses/403334/external_tools/43025

### DIFF
--- a/ccm_web/server/src/app.module.ts
+++ b/ccm_web/server/src/app.module.ts
@@ -1,8 +1,9 @@
 import helmet from 'helmet'
+import { NoCacheInterceptor } from './no.cache.interceptor'
 import { MiddlewareConsumer, Module, NestModule, RequestMethod } from '@nestjs/common'
+import { APP_INTERCEPTOR } from '@nestjs/core'
 import { ConfigModule, ConfigService } from '@nestjs/config'
 import { SequelizeModule } from '@nestjs/sequelize'
-import { APP_INTERCEPTOR } from '@nestjs/core'
 
 import { APIModule } from './api/api.module'
 import { AuthModule } from './auth/auth.module'
@@ -13,7 +14,6 @@ import { UserModule } from './user/user.module'
 import { User } from './user/user.model'
 import { UserService } from './user/user.service'
 
-import { CacheControlToHeaderInterceptor } from './no.cache.interceptor'
 import { Config, validateConfig } from './config'
 import baseLogger from './logger'
 
@@ -50,7 +50,7 @@ const logger = baseLogger.child({ filePath: __filename })
   providers: [
     {
       provide: APP_INTERCEPTOR,
-      useClass: CacheControlToHeaderInterceptor
+      useClass: NoCacheInterceptor
     },
     UserService
   ]
--- a/ccm_web/server/src/app.module.ts
+++ b/ccm_web/server/src/app.module.ts
@@ -2,6 +2,7 @@ import helmet from 'helmet'
 import { MiddlewareConsumer, Module, NestModule, RequestMethod } from '@nestjs/common'
 import { ConfigModule, ConfigService } from '@nestjs/config'
 import { SequelizeModule } from '@nestjs/sequelize'
+import { APP_INTERCEPTOR } from '@nestjs/core'
 
 import { APIModule } from './api/api.module'
 import { AuthModule } from './auth/auth.module'
@@ -12,6 +13,7 @@ import { UserModule } from './user/user.module'
 import { User } from './user/user.model'
 import { UserService } from './user/user.service'
 
+import { CacheControlToHeaderInterceptor } from './no.cache.interceptor'
 import { Config, validateConfig } from './config'
 import baseLogger from './logger'
 
@@ -45,7 +47,13 @@ const logger = baseLogger.child({ filePath: __filename })
     CanvasModule,
     APIModule
   ],
-  providers: [UserService]
+  providers: [
+    {
+      provide: APP_INTERCEPTOR,
+      useClass: CacheControlToHeaderInterceptor
+    },
+    UserService
+  ]
 })
 export class AppModule implements NestModule {
   configure (consumer: MiddlewareConsumer): void {
--- a/ccm_web/server/src/main.ts
+++ b/ccm_web/server/src/main.ts
@@ -14,7 +14,6 @@ import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger'
 
 import { AppModule } from './app.module'
 
-import { CacheControlToHeaderInterceptor } from './no.cache.interceptor'
 import { Config } from './config'
 import baseLogger from './logger'
 
@@ -69,7 +68,6 @@ async function bootstrap (): Promise<void> {
       }
     })
   )
-  app.useGlobalInterceptors(new CacheControlToHeaderInterceptor())
 
   app.useGlobalPipes(new ValidationPipe())
 
--- a/ccm_web/server/src/main.ts
+++ b/ccm_web/server/src/main.ts
@@ -14,6 +14,7 @@ import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger'
 
 import { AppModule } from './app.module'
 
+import { CacheControlToHeaderInterceptor } from './no.cache.interceptor'
 import { Config } from './config'
 import baseLogger from './logger'
 
@@ -68,6 +69,7 @@ async function bootstrap (): Promise<void> {
       }
     })
   )
+  app.useGlobalInterceptors(new CacheControlToHeaderInterceptor())
 
   app.useGlobalPipes(new ValidationPipe())
 
--- a/ccm_web/server/src/no.cache.interceptor.ts
+++ b/ccm_web/server/src/no.cache.interceptor.ts
@@ -1,0 +1,15 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common'
+import { Observable } from 'rxjs'
+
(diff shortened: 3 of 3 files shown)


ccm_web/server/src/no.cache.interceptor.ts
```diff
@@ -0,0 +1,15 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common'
```
pushyamig, line 1: Since I am adding the cache control globally in express, I felt having the no cache interceptor at the root level is logical. we can create a new folder `global interceptor` and move it there. I am open to ideas

ccm_web/server/src/main.ts
```diff
@@ -68,6 +69,7 @@ async function bootstrap (): Promise<void> {
       }
     })
   )
+  app.useGlobalInterceptors(new CacheControlToHeaderInterceptor())
```
pushyamig, line 72: I Felt adding it globally is a best way than adding a Decorator at the API level  https://docs.nestjs.com/controllers#headers

ssciolla: My main question is whether adding it globally here will affect the static routes. My guess is no currently, but it may once we make a change like in #316, because the static serving will be installed at the module level, not globally. Hm...

pushyamig: Then I can add this interceptor at the API level, if you feel it is going to disrupt the static routes with #316. Let me pull your change from PR  #316 and combine mine and see if anything changes.

As of now the static assets like the `main-sdsd.js` file have always caching on it.

ccm_web/server/src/no.cache.interceptor.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { Response as ExpressResponse } from 'express'
+
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#the_cache-control_header
+
+@Injectable()
+export class CacheControlToHeaderInterceptor implements NestInterceptor {
+  intercept (context: ExecutionContext, next: CallHandler): Observable<unknown> {
+    const ResponseObj: ExpressResponse = context.switchToHttp().getResponse()
+    ResponseObj.setHeader('Cache-Control', 'no-store')
```
jonespm, line 12: I don't know too much about what headers are appropriate here either.

I'm looking at Canvas and they just set
`max-age=0`

But they're also using this if-none-match and etag to improve performance. I'd be a little concerned about the performance of just setting it to `no-store` without considering etags. I don't know how much effort that would take in a nestjs app though. Maybe this is related? https://wanago.io/2022/01/17/api-nestjs-etag-cache/

pushyamig: > max-age=0

I just noticed it my self just now. I think this will bad. I am going to change it back not in the `main.ts` and put in the `api.controller.ts` and see if that fixes that.

pushyamig: I tested again on canvas beta server for Cache-control header for the static file `main.ts`. I do see randomly it introduces max-age =0. So I was thinking it might be my code doing it, but probably not. Express/NestJS handling it some how. Chatting with @jonespm said Due to Etag this might be controlled and *might* not be an issue. But need to talk to Ross about it

pushyamig: @ssciolla and @jonespm , Ross said he will test and let us know if we **really** need to add cache headers or default express Etag step is good enough.

ccm_web/server/src/main.ts
pushyamig: @ssciolla answers from these https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/321#discussion_r797989387 will help define it.

ccm_web/server/src/no.cache.interceptor.ts
pushyamig: After our discussion in today's meeting. I did testing in 4 scenarios looking for Status code response time for a browser fetch (GET) for static assets like (`main.js`, I only looked at this) and an API response (like  get sections)

Scenarios tested:
1. Code change from this branch/PR, Local run, Configured on Canvas Dev
2. Code Change mixing changes from this PR and Sam's PR #309, Local run, Configured on Canvas Dev
3. CCM dev instance configured on Canvas test, running latest from git `Main`
4. CCM Test instance configured on Canvas Beta, running latest from git `Main`

My finding are recored [here](https://docs.google.com/spreadsheets/d/1Wu085vyFmuDBuVpYaxmz6vKazQoonAa1oevy3ZEVicE/edit#gid=0)

@ssciolla and @jonespm Let me know your thoughts. My conclusion is adding this change won't be disruptive.

ssciolla: Okay sounds like there will be no issue with what you are doing. I will do some local testing this afternoon and then approve if I don't see any issues.

pushyamig: @ssciolla I want you to pull this changes in your #316 and test it.

ccm_web/server/src/main.ts
ssciolla: Looks like the global interceptor doesn't really affect the static. It may be slightly clearer to install the interceptor in `app.module.ts`, using the latter pattern here: https://docs.nestjs.com/interceptors#binding-interceptors. I'll leave that up to you. I like that we wouldn't have to import any Nest `Injectable`s except `app.module.ts` in `main.ts`, and that it would automatically be included in any test modules if we ever get to building out the end-to-end tests.

ssciolla: Looks like the global interceptor doesn't really affect the static. It may be slightly clearer to install the interceptor in `app.module.ts`, using the latter pattern here: https://docs.nestjs.com/interceptors#binding-interceptors. I'll leave that up to you. I like that we wouldn't have to import any Nest `Injectable`s except `app.module.ts` in `main.ts`, and that it would automatically be included in any test modules if we ever get to building out the end-to-end tests.

ccm_web/server/src/app.module.ts
```diff
@@ -45,7 +47,13 @@ const logger = baseLogger.child({ filePath: __filename })
     CanvasModule,
     APIModule
   ],
-  providers: [UserService]
+  providers: [
+    {
+      provide: APP_INTERCEPTOR,
+      useClass: CacheControlToHeaderInterceptor
+    },
+    UserService
```
pushyamig, line 55: @ssciolla I add the interceptor at in the app module instead of `main.ts`. I also feel this is clean. I tested I don't see any difference by changing the location. I still see cache header and statics are not effected

For some reason, your comment on this is not appearing in the PR only saw via Github email notification.

ssciolla: Yeah, sorry, I meant to start a review, so I deleted it and re-posted it as pending. Thanks for doing that

ccm_web/server/src/no.cache.interceptor.ts
ssciolla, line 12:
```suggestion
    const responseObj: ExpressResponse = context.switchToHttp().getResponse()
    responseObj.setHeader('Cache-Control', 'no-store')
```

ssciolla (approved):
I think this looks good! I guess time will tell if this is a significant performance hit. My gut says it's not.

For what it's worth, I saw that you can apply a `@Header` decorator to specific controller methods. I don't know if there's a way to do it for an entire controller or all controllers. https://docs.nestjs.com/controllers#headers

ccm_web/server/src/no.cache.interceptor.ts
```diff
@@ -0,0 +1,15 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common'
+import { Observable } from 'rxjs'
+
+import { Response as ExpressResponse } from 'express'
+
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#the_cache-control_header
+
+@Injectable()
+export class CacheControlToHeaderInterceptor implements NestInterceptor {
```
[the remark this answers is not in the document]
pushyamig: Done

ccm_web/server/src/app.module.ts
```diff
@@ -12,6 +13,7 @@ import { UserModule } from './user/user.module'
 import { User } from './user/user.model'
 import { UserService } from './user/user.service'
 
+import { CacheControlToHeaderInterceptor } from './no.cache.interceptor'
```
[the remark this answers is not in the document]
pushyamig: Done

ccm_web/server/src/app.module.ts
```diff
@@ -2,6 +2,7 @@ import helmet from 'helmet'
 import { MiddlewareConsumer, Module, NestModule, RequestMethod } from '@nestjs/common'
 import { ConfigModule, ConfigService } from '@nestjs/config'
 import { SequelizeModule } from '@nestjs/sequelize'
+import { APP_INTERCEPTOR } from '@nestjs/core'
```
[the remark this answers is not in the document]
pushyamig: Done

Issue #293, opened 2022-01-06
Change button disabling/incomplete error logic across client


ssciolla:
While working on #270, we've discussed the pros/cons of using button disabling to indicate whether the user still needs to complete something before they can proceed. We seem to have settled on the position that only disabling can be confusing and less clear to the user than allowing the button to be clicked and then showing error alerts where action needs to be taken.

It's probably best to take this approach everywhere for consistency. The main places where disabling currently happens are with "Select Section" workflow steps, and in the "Merge Sections" feature.

<img width="500" alt="Screen Shot 2022-01-06 at 12 11 32 PM" src="https://user-images.githubusercontent.com/35741256/148423297-423493b8-39c8-42f1-974f-c5d910a7769b.png">

pushyamig:
@ssciolla As understood from our discussion, this is Icebox issue?

ssciolla:
@pushyamig, that's your call.

PR #325, opened 2022-02-03
#324 non-umich text changes

### LINKED ISSUE #324: Add Non-UM user text changes
Text changes suggested as per usability [review](https://docs.google.com/document/d/1v1TW2YUgiERAv_7QYSDT2d4W1e_u1A6XduSwKwAVtT8/edit#heading=h.4baijvfc8vi8)

1. "first_name" with the user's first name (between 1 and 255 characters in length);
2. "last_name" with the user's last name (between 1 and 255 characters in length).
3. This tool will enroll non-UM users in the selected section. If they do not have an account in Canvas, they will be sent an email invitation to choose a login method and added to Canvas before they are enrolled in the section.
4. Home page of Enroll user ( UM and Non-UM) should have text `(Required)`

pushyamig:
Fixes #324

### DIFF
--- a/ccm_web/client/src/components/MethodSelect.tsx
+++ b/ccm_web/client/src/components/MethodSelect.tsx
@@ -38,7 +38,7 @@ export default function UserInputMethodSelect<T extends string> (props: MethodSe
     <>
     <div className={classes.spacing}>
       <FormControl component='fieldset'>
-        <FormLabel required className={classes.spacing}>{props.label}</FormLabel>
+        <FormLabel className={classes.spacing}>{props.label} (Required)</FormLabel>
         <RadioGroup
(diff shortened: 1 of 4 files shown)


ssciolla:
There are other places in the UI where a selection or entry is "required". Should those also be changed? This is somewhat related to #293.

pushyamig:
> There are other places in the UI where a selection or entry is "required". Should those also be changed? This is somewhat related to https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/293.

added few more.

ssciolla (approved):
Looks fine. We may need to adjust some section selection text as well, but that could be resolved with #293.

Issue #272, opened 2021-12-13
Snack bar error message should show the API response reasoning for failure. 


pushyamig:
Currently for many features that use snack bar for reporting success/failure. The error message just mention "Course rename is failed" Or Section creation has errors" instead show the message sent from the API response. 

There might be more places where this pattern might occurring. This issue should address all of then.

pushyamig:
@ssciolla Can you give me an estimate on this issue?

melindakraft:
I'm not testing for every error to validate all of the snack bar messages, however, I have seen this manifest and it appears messaging is working as expected.

PR #327, opened 2022-02-03
Introduce `createLimitedPromises` wrapper for `p-limit` (#326)

### LINKED ISSUE #326: Support upload of 400 users for Add UM user
Current testing indicate that Ad UM user unable to bulk upload 400 user successfully and the last tested the number of users added was about 289. We should come up with a solution that consistently add 400 user, if not come up with a suggest X number user for adding. 

The reason for errors due to Rate limiting 403 all the times

Look for testing outcomes and results from [here](https://docs.google.com/spreadsheets/d/1pNOlke-QLW-IqArn-2e_iF0Vp0uEKxzUds8bZRUKRJQ/edit#gid=1224470874)

ssciolla:
The PR aims to resolve #326.

(I'm pretty sure this does the trick, and it kinda feels like magic 🪄 .)

Resource(s):
- https://github.com/sindresorhus/p-limit
- https://stackoverflow.com/a/52262024

### DIFF
--- a/ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
+++ b/ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
@@ -273,9 +273,9 @@ export default function SingleSectionEnrollmentWorkflow (props: SingleSectionEnr
         <Backdrop className={classes.backdrop} open={isAddEnrollmentsLoading}>
           <Grid container>
             <Grid item xs={12}><CircularProgress color='inherit' /></Grid>
-            <Grid item xs={12}>Loading...</Grid>
+            <Grid item xs={12}>Sending enrollment data...</Grid>
             <Grid item xs={12}>
-              Please stay on the page. This may take up to a couple of minutes for larger files.
+              Please stay on the page. This process may take several seconds for larger files.
             </Grid>
           </Grid>
         </Backdrop>
--- a/ccm_web/package.json
+++ b/ccm_web/package.json
@@ -50,6 +50,7 @@
     "morgan": "^1.10.0",
     "mysql2": "^2.3.2",
     "notistack": "^1.0.10",
+    "p-limit": "^3.1.0",
     "papaparse": "^5.3.1",
     "passport": "^0.4.1",
     "passport-jwt": "^4.0.0",
--- a/ccm_web/server/src/api/api.admin.handler.ts
+++ b/ccm_web/server/src/api/api.admin.handler.ts
@@ -2,7 +2,7 @@ import CanvasRequestor from '@kth/canvas-api'
 
 import { CourseApiHandler } from './api.course.handler'
 import { APIErrorData, isAPIErrorData } from './api.interfaces'
-import { handleAPIError, makeResponse } from './api.utils'
+import { handleAPIError, limitPromises, makeResponse } from './api.utils'
 import { CanvasAccount, CanvasCourse, CourseWithSections, CourseWorkflowState } from '../canvas/canvas.interfaces'
 
 import baseLogger from '../logger'
@@ -84,7 +84,9 @@ export class AdminApiHandler {
     }
     if (instructor !== undefined) queryParams.by_teachers = ['sis_login_id:' + instructor]
     if (courseName !== undefined) queryParams.search_term = courseName
-    const coursesApiPromises = accountIds.map(async (a) => await this.getAccountCourses(a, queryParams))
+    const coursesApiPromises = limitPromises<CanvasCourse[] | APIErrorData>(
+      accountIds.map(a => async () => await this.getAccountCourses(a, queryParams))
+    )
     const coursesResponses = await Promise.all(coursesApiPromises)
     const result = makeResponse<CanvasCourse[]>(coursesResponses)
     if (isAPIErrorData(result)) return result
@@ -92,10 +94,12 @@ export class AdminApiHandler {
     result.map(cs => allCourses.push(...cs))
 
     // Get sections for those courses
-    const coursesWithSectionsApiPromises = allCourses.map(async c => {
-      const courseHandler = new CourseApiHandler(this.requestor, c.id)
-      return await courseHandler.getSectionsWithCourse(CourseApiHandler.slimCourse(c))
-    })
+    const coursesWithSectionsApiPromises = limitPromises(
+      allCourses.map(c => async () => {
+        const courseHandler = new CourseApiHandler(this.requestor, c.id)
+        return await courseHandler.getSectionsWithCourse(CourseApiHandler.slimCourse(c))
+      })
+    )
     const coursesWithSectionsResult = await Promise.all(coursesWithSectionsApiPromises)
     const finalResult = makeResponse<CourseWithSections>(coursesWithSectionsResult)
 
--- a/ccm_web/server/src/api/api.admin.handler.ts
+++ b/ccm_web/server/src/api/api.admin.handler.ts
@@ -84,9 +84,7 @@ export class AdminApiHandler {
     }
     if (instructor !== undefined) queryParams.by_teachers = ['sis_login_id:' + instructor]
     if (courseName !== undefined) queryParams.search_term = courseName
-    const coursesApiPromises = createLimitedPromises<CanvasCourse[] | APIErrorData>(
-      accountIds.map(a => async () => await this.getAccountCourses(a, queryParams))
-    )
+    const coursesApiPromises = accountIds.map(async (a) => await this.getAccountCourses(a, queryParams))
     const coursesResponses = await Promise.all(coursesApiPromises)
     const result = makeResponse<CanvasCourse[]>(coursesResponses)
     if (isAPIErrorData(result)) return result
--- a/ccm_web/server/src/api/api.admin.handler.ts
+++ b/ccm_web/server/src/api/api.admin.handler.ts
@@ -2,7 +2,7 @@ import CanvasRequestor from '@kth/canvas-api'
 
 import { CourseApiHandler } from './api.course.handler'
 import { APIErrorData, isAPIErrorData } from './api.interfaces'
-import { handleAPIError, limitPromises, makeResponse } from './api.utils'
+import { createLimitedPromises, handleAPIError, makeResponse } from './api.utils'
 import { CanvasAccount, CanvasCourse, CourseWithSections, CourseWorkflowState } from '../canvas/canvas.interfaces'
 
 import baseLogger from '../logger'
@@ -84,7 +84,7 @@ export class AdminApiHandler {
     }
     if (instructor !== undefined) queryParams.by_teachers = ['sis_login_id:' + instructor]
     if (courseName !== undefined) queryParams.search_term = courseName
-    const coursesApiPromises = limitPromises<CanvasCourse[] | APIErrorData>(
+    const coursesApiPromises = createLimitedPromises<CanvasCourse[] | APIErrorData>(
       accountIds.map(a => async () => await this.getAccountCourses(a, queryParams))
     )
     const coursesResponses = await Promise.all(coursesApiPromises)
@@ -94,7 +94,7 @@ export class AdminApiHandler {
     result.map(cs => allCourses.push(...cs))
 
     // Get sections for those courses
-    const coursesWithSectionsApiPromises = limitPromises(
+    const coursesWithSectionsApiPromises = createLimitedPromises(
       allCourses.map(c => async () => {
         const courseHandler = new CourseApiHandler(this.requestor, c.id)
         return await courseHandler.getSectionsWithCourse(CourseApiHandler.slimCourse(c))
--- a/ccm_web/server/src/api/api.course.handler.ts
+++ b/ccm_web/server/src/api/api.course.handler.ts
@@ -2,7 +2,7 @@ import CanvasRequestor from '@kth/canvas-api'
 
 import { APIErrorData, isAPIErrorData } from './api.interfaces'
 import { SectionApiHandler } from './api.section.handler'
-import { handleAPIError, limitPromises, HttpMethod, makeResponse } from './api.utils'
+import { createLimitedPromises, handleAPIError, HttpMethod, makeResponse } from './api.utils'
 import {
   CanvasCourse, CanvasCourseBase, CanvasCourseInput, CanvasCourseSection, CanvasCourseSectionBase,
   CourseWithSections
@@ -107,7 +107,7 @@ export class CourseApiHandler {
 
   async createSections (sections: string[]): Promise<CanvasCourseSection[] | APIErrorData> {
     const start = process.hrtime()
-    const apiPromises = limitPromises<CanvasCourseSection | APIErrorData>(
+    const apiPromises = createLimitedPromises<CanvasCourseSection | APIErrorData>(
       sections.map(section => async () => await this.createSection(section))
     )
     const sectionsOrErrorDataObjs = await Promise.all(apiPromises)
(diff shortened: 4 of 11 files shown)


ccm_web/package.json
```diff
@@ -49,6 +49,7 @@
     "morgan": "^1.10.0",
     "mysql2": "^2.3.2",
     "notistack": "^1.0.10",
+    "p-limit": "^3.1.0",
```
ssciolla, line 52: There is a more recent version of `p-limit`, version `4`, but it uses ESM, which would require some tinkering with `package.json` and `tsconfig`s, and maybe more dramatic import changes across the project. This should probably be handled separately, and it may not be urgent. I need to do more reading about it. See https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c and https://www.typescriptlang.org/docs/handbook/esm-node.html

pushyamig: > but it uses ESM
What are we currently supporting now if it is not ECMA Script module? I felt we are Pardon me for my ignorance.

ssciolla: I'm not an expert on this, but I believe CommonJS is the answer. There is some interoperability between module systems, but I got an error when trying to run `4.0.0` in the backend. The resources I linked to should provide more info. I will try this separately, I believe @zqian also commented on this at one point. While NodeJS seems to have implemented ESM, TypeScript seems to be flagging their support as experimental.

pushyamig:
I will review this some time in late afternoon.

pushyamig:
I did some testing with 400 user enrolling 4 times and in all the time i saw the success. [here](https://docs.google.com/spreadsheets/d/1JcwQbLHhhgnrzIntVeQb7-1oSTXvDlUFzBEj4MSEr10/edit#gid=0) is the report from that. I might so testing on Monday.

ccm_web/package.json
ssciolla: See #330.

ccm_web/server/src/api/api.utils.ts
```diff
@@ -63,3 +64,19 @@ export function makeResponse<T> (multipleResults: Array<APIErrorData | T>): T[]
     }
   }
 }
+
+/*
+Convenience wrapper for p-limit: see https://github.com/sindresorhus/p-limit
+*/
+export function createLimitedPromises<T> (
+  funcs: Array<() => Promise<T>>, maxConcurrentPromises = 20
```
ssciolla, line 72: I just changed the default here from `5` to `20`. I've been trying a few different values, and observing how the `x-rate-limit-remaining` value changes. I tried `10`, `20`, and `40`. With some repeated batches of 400, `40` was triggering some retries. `20` still seemed to be super fast, with runs taking 12, 13, 18 seconds, and I never saw retries, even with batches in quick succession. The `x-rate-limit-remaining` seems to have a maximum of `700`, and I never saw it go below `400`. It certainly needs more testing, but this feels pretty good to me. If the results are this consistent, we may want to change the loading message so it reflects the typical waiting time.

pushyamig:
I will review some more by end of day

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -84,18 +84,22 @@ export class AdminApiHandler {
     }
     if (instructor !== undefined) queryParams.by_teachers = ['sis_login_id:' + instructor]
     if (courseName !== undefined) queryParams.search_term = courseName
-    const coursesApiPromises = accountIds.map(async (a) => await this.getAccountCourses(a, queryParams))
+    const coursesApiPromises = createLimitedPromises<CanvasCourse[] | APIErrorData>(
+      accountIds.map(a => async () => await this.getAccountCourses(a, queryParams))
+    )
```
ssciolla, line 89: We probably don't have to limit one, since we're just making one (albeit potentially rather large paginated request) for each parent account the user has. It can't be that many, even if they're a sub account admin in several accounts. Especially if we're at `20` for a limit, this won't make much of a difference.

ccm_web/server/src/api/api.course.handler.ts
```diff
@@ -107,7 +107,9 @@ export class CourseApiHandler {
 
   async createSections (sections: string[]): Promise<CanvasCourseSection[] | APIErrorData> {
     const start = process.hrtime()
-    const apiPromises = sections.map(async (section) => await this.createSection(section))
+    const apiPromises = createLimitedPromises<CanvasCourseSection | APIErrorData>(
+      sections.map(section => async () => await this.createSection(section))
+    )
```
ssciolla, line 112: Probably okay without this limiting here also, since the limit is 60, but it won't hurt anything either.

ccm_web/server/src/api/api.course.handler.ts
```diff
@@ -118,10 +120,12 @@ export class CourseApiHandler {
...
-    const apiPromises = sectionIds.map(async (si) => {
-      const sectionHandler = new SectionApiHandler(this.requestor, si)
-      return await sectionHandler.mergeSection(this.courseId)
-    })
+    const apiPromises = createLimitedPromises<CanvasCourseSectionBase | APIErrorData>(
+      sectionIds.map(si => async () => {
+        const sectionHandler = new SectionApiHandler(this.requestor, si)
+        return await sectionHandler.mergeSection(this.courseId)
+      })
+    )
```
ssciolla, line 128: I'd probably keep this here, since theoretically we allow up to 250 sections here, but the frontend isn't doing that (I think it's only ever one at a time, actually).

ccm_web/server/src/api/api.service.ts
```diff
@@ -142,10 +146,12 @@ export class APIService {
...
-    const apiPromises = sectionIds.map(async (si) => {
-      const sectionHandler = new SectionApiHandler(requestor, si)
-      return await sectionHandler.unmergeSection()
-    })
+    const apiPromises = createLimitedPromises<CanvasCourseSectionBase | APIErrorData>(
+      sectionIds.map(si => async () => {
+        const sectionHandler = new SectionApiHandler(requestor, si)
+        return await sectionHandler.unmergeSection()
+      })
+    )
```
ssciolla, line 154: See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/327/files#r803102325, same logic applies.

ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -23,11 +23,11 @@ type SupportedAPIEndpoint = '/api/v1/' | '/api/graphql/'
 
 const requestorOptions: GotOptions = {
   retry: {
-    limit: 4,
+    limit: 3,
     methods: ['POST', 'GET', 'PUT', 'DELETE'],
     statusCodes: got.defaults.options.retry.statusCodes.concat([403]),
     calculateDelay: ({ attemptCount, retryOptions, error, computedValue }) => {
-      const delay = computedValue === 0 ? 0 : 5000
+      const delay = computedValue === 0 ? 0 : 2000
```
ssciolla, line 30: I just did this. My thinking is that we initially set these so high as a way to back off from the rate limit, and get all the request through, but that shouldn't be necessary anymore. We still want retrying behavior for intermittent Canvas errors and such, but lower levels like this should be sufficient.

ssciolla:
@pushyamig and @lsloan, I think this is ready for review. Please let me know if any of the cases where I introduced limiting you think it should be removed (fine doing that, just want your input; I weighed in on the cases where it seemed less useful).

ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
```diff
@@ -273,9 +273,9 @@ export default function SingleSectionEnrollmentWorkflow (props: SingleSectionEnr
         <Backdrop className={classes.backdrop} open={isAddEnrollmentsLoading}>
           <Grid container>
             <Grid item xs={12}><CircularProgress color='inherit' /></Grid>
-            <Grid item xs={12}>Loading...</Grid>
+            <Grid item xs={12}>Sending enrollment data...</Grid>
             <Grid item xs={12}>
-              Please stay on the page. This may take up to a couple of minutes for larger files.
+              Please stay on the page. This process may take several seconds for larger files.
```
pushyamig, line 278: Thx for updating these user message, often forgotten!

ccm_web/server/src/api/api.course.handler.ts
pushyamig: Sure, Now that I think about it. Doesn't hurt to have it on create section as well

ccm_web/server/src/api/api.utils.ts
pushyamig: can the `maxConcurrentPromises` be a configurable option?

ccm_web/server/src/api/api.admin.handler.ts
pushyamig: Good thinking on wrapping the concurrency code to Getting courses from an account, I was bit worried if user search by course name `Bio` then it might fetch lot of courses and will end up with rate limiting errors, this should take care it

ccm_web/server/src/api/api.course.handler.ts
pushyamig: I don't think we will ever merge 250 sections. Not a practical use case. I am fine if FE is not validating that and BE is is type validating/guarding it.

pushyamig:
The change seems to be straight forward but affects most features. I did some preliminary testing and review, things looks good! I will go one more round tomorrow and will approve.

ccm_web/server/src/api/api.utils.ts
ssciolla: We certainly coould, but in what situation can you imagine where we would want to be able to change this in production? I feel like it's better to just test to make sure it performs well with a certain number. I'm just worried about making the configuration too complicated (an issue we've had with other projects).

pushyamig: We got too many configuration value, I am ok to leave it. Since if this works consistently we never going to change it and I am not anticipating we will be supporting 700 users enrollment.

ssciolla: I guess it's possible Canvas could adjust their rate limit, and we may need to respond to that. Hm... it's possible to make it configurable, but as it is now, we'd have to pass in a configuration variable all the time (since this function lives in `api.utils`, and I can't simply inject `ConfigService` here. Maybe I can rework this so that isn't necessary...

pushyamig: @ssciolla Sorry for not commenting before. I think it is ok not make the concurrency configurable. We can think about it after the initial release

ccm_web/server/src/api/api.service.ts
pushyamig: Why unmerge has to be wrapped with the concurrency call? since we mostly ever going to do 1 section unmerge at a time

ssciolla: That's fine, I can remove it if you want, I was just looking for feedback on which to remove it from.

ssciolla: Theoretically, someone could do up to 250 using the API, but not likely to ever happen.

pushyamig: Since UI doesn't support bulk unmerge and only one at a time, I believe we can remove it

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -84,18 +84,22 @@ export class AdminApiHandler {
...
 
     // Get sections for those courses
-    const coursesWithSectionsApiPromises = allCourses.map(async c => {
-      const courseHandler = new CourseApiHandler(this.requestor, c.id)
-      return await courseHandler.getSectionsWithCourse(CourseApiHandler.slimCourse(c))
-    })
+    const coursesWithSectionsApiPromises = createLimitedPromises(
+      allCourses.map(c => async () => {
+        const courseHandler = new CourseApiHandler(this.requestor, c.id)
+        return await courseHandler.getSectionsWithCourse(CourseApiHandler.slimCourse(c))
```
pushyamig, line 100: Also good idea to add to get sections call, I am not anticipating if a course will ever have more than 100 sections, but it is good to add.

pushyamig:
Over all this is a positive change and huge performance improvement and increase customer satisfaction rate. This also means that for Add Non-UM user we could support more that 200 (based on testing).

I have my testing report across [features](https://docs.google.com/spreadsheets/d/1JcwQbLHhhgnrzIntVeQb7-1oSTXvDlUFzBEj4MSEr10/edit#gid=0)

This is a go from me, I think just need to resolve some merge conflicts and probably remove the concurrency call with un merge sections

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: Just to clarify, this is limiting the number of promises for calls of `this.getAccountCourses`, so if there were three accounts that someone was a sub-account admin for, that would be three promises. The `@kth/canvas-api` library is doing the paging, so that may actually involve multiple calls behind the scenes (probably in sequence?), but for us it's just one promise. Does that make sense? That's why I was saying it was minimal improvement here.

pushyamig: Got it.

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: Similar to the comment about `getAccountCourses`, here we're limiting promises of calls to `courseHandler.getSectionsWithCourse`, so one promise per course. @kth/canvas-api handles the paging for getting sections. It still makes sense to me here because if there are lots of courses returned, we want to limit that.

ccm_web/server/src/api/api.course.handler.ts
ssciolla: Mmm, I was wrong about this. The "Go Merge" could trigger multiple unmerge calls at once. Unlikely to be very many, but the hard limit is 250. I suppose it's possible the user could hit that in the UI. Hopefully we have a way of representing that error.

ccm_web/server/src/api/api.service.ts
ssciolla: So you're not worried about direct API calls? This is the DTO for both merge and unmerge: https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/main/ccm_web/server/src/api/dtos/api.section.ids.dto.ts

ccm_web/server/src/api/api.admin.handler.ts
pushyamig: OK. I will go by your recommendation.

ccm_web/server/src/api/api.course.handler.ts
pushyamig: > Hopefully we have a way of representing that error.

I feel this should be addressed this as any other error during this API call. This case shouldn't be any exception. I think either way this issue is separate from this. You can create an separate issue but I won't necessary want this fix for initial release. since I am not anticipating people will ever merge 250 section but can be handled later.

ccm_web/server/src/api/api.service.ts
pushyamig: Ok I checked again, even though this is list ` sectionIds.map` but that list will always have 1 section for unmerge section. So the wrapper for p-limit strictly it is not needed since we are queuing one call. But I will leave this up to you

pushyamig (approved):
Nice Work @ssciolla

ccm_web/server/src/api/api.utils.ts
ssciolla: Okay, we can wait on this.

ccm_web/server/src/api/api.service.ts
ssciolla: I'm gonna leave it. My thinking is that the backend logic should reflect what we support in the API, since either the frontend could change, or we may want to use the API directly in some strange customer support scenario. In all other cases where we could have >20 calls, we're using this, so it makes sense to be consistent.

ccm_web/server/src/api/api.course.handler.ts
ssciolla: Yeah, I can do some testing as part of #272. It's an edge case though.

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: I elected to remove the limiting here. I think it's highly unlikely that anyone would be an admin in >20 sub-accounts. We can reassess if we ever see an edge case scenario.

ssciolla:
@pushyamig, thanks for the thorough review!

ccm_web/server/src/api/api.admin.handler.ts
pushyamig: I agree

PR #332, opened 2022-02-04
Handle LTI launch error due to browser cookie settings (#312)

### LINKED ISSUE #312: Error when launching CCM in Safari

ssciolla:
The PR aims to resolve #312.

Screenshot of error page

Screen Shot 2022-03-09 at 3 59 39 PM

### DIFF
--- a/ccm_web/client/src/pages/LaunchError.tsx
+++ b/ccm_web/client/src/pages/LaunchError.tsx
@@ -2,10 +2,13 @@ import React from 'react'
 
 import { Typography } from '@material-ui/core'
 
+import HelpContact, { HelpContactProps } from '../components/HelpContact'
 import InlineErrorAlert from '../components/InlineErrorAlert'
 import Layout from '../components/Layout'
 
-export default function LaunchError (): JSX.Element {
+interface LaunchErrorProps extends HelpContactProps {}
+
+export default function LaunchError (props: LaunchErrorProps): JSX.Element {
   return (
     <Layout>
       <InlineErrorAlert>
@@ -14,6 +17,7 @@ export default function LaunchError (): JSX.Element {
           The tool could not be launched. It is likely your browser is blocking third-party cookies,
           which are required for this application.
         </Typography>
+        <HelpContact {...props} />
       </InlineErrorAlert>
     </Layout>
   )
--- a/ccm_web/client/src/pages/LaunchError.tsx
+++ b/ccm_web/client/src/pages/LaunchError.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { Typography } from '@material-ui/core'
+import { Link, Typography } from '@material-ui/core'
 
 import HelpContact, { HelpContactProps } from '../components/HelpContact'
 import InlineErrorAlert from '../components/InlineErrorAlert'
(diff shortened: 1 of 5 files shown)


ccm_web/server/src/lti/lti.service.ts
```diff
@@ -113,6 +113,10 @@ export class LTIService implements BeforeApplicationShutdown {
       })
     })
 
+    provider.onInvalidToken(async (req: Request, res: Response) => {
+      return res.redirect('/launch-error')
```
pushyamig, line 117: Can you tell the difference between launch from safari vs student launching CCM (by somehow grabbing url)? doesn't it both invalid token launches. 

What you have right now will work and not asking to detect safari, just want to know if there is a way LTIJS is giving that info.

ssciolla: I think if it's a student, the launch still works, we just don't log them in and redirect them to `access-denied`. See `onConnect` hook.

ccm_web/client/src/pages/LaunchError.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default function LaunchError (): JSX.Element {
+  const classes = useStyles()
+
+  return (
+    <div className={classes.container}>
+      <InlineErrorAlert>
+        <Typography variant='subtitle1' component='h1'>Launch Error</Typography>
+        <Typography>
+          The tool could not be launched. This is likely due to settings related to cookies on your Internet browser.
```
pushyamig, line 21: Should the text be updated as per this https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/312#issuecomment-1055860902

ssciolla: Okay, I'll do this today.

ccm_web/server/src/lti/lti.service.ts
pushyamig: I looked it up a bit more. I see that `LTI login/authorize/Launch` is happening if 3rd-party cookies are disabled. some how the code block `onInvalidToken` is detecting as cookies are disabled and tool launch is failed. So I did not see from the code either (documentation of LTIJS is not much about this) much description about this function. 

I would just go with an understanding that **onInvalidToken** is written for seeing cookies disable or enabled. IF that is not true then we can find out based on the error case that comes to us in future.

ccm_web/client/src/pages/LaunchError.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <Typography>
+          The tool could not be launched. It is likely your browser is blocking third-party cookies,
+          which are required for this application.
+        </Typography>
+        <Typography>
+          To resolve this issue, please update your settings
+          for <Link href={chromeLink}>Chrome</Link>, <Link href={macOSLink}>Safari for MacOS</Link>
+          , <Link href={iOSLink}>Safari for iOS</Link>, <Link href={firefoxLink}>Firefox</Link>
+          , <Link href={edgeLink}>Microsoft Edge</Link>,
+          or another preferred browser to allow third-party cookies.
```
pushyamig, line 31: So far no objection from Melinda about this text and on that note this is ok to merge

pushyamig (approved):

ssciolla:
Thanks, @pushyamig, I'll merge and we can always adjust text later.

PR #334, opened 2022-02-07
Change list structure and otherwise improve `SectionSelectorWidget` (#230, #260)

### LINKED ISSUE #230: Fix alignment of Select All checkbox in `SectionSelectorWidget`
Optional : Select All is strictly not centered, I feel it should be left align
Screen Shot 2021-10-28 at 5 11 11 PM
ed

_Originally posted by @pushyamig in https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/223#issuecomment-954221928_

### LINKED ISSUE #260: Section selection needs better instrumentation
Location: Format Third-Party Gradebook, section selection
Impact Blocker/High
Type: Semantics and structure

Issue
Section selection needs better instrumentation

Impact
Screen reader users will be able to tell what sections are selected only with difficulty or not at all.

Solution
List of sections is coded as a `list <ul> but the list children are div[role="button"] `- this is might cause parsing errors in user agens and assistive technology. 
The state of the button is not clear. It is not set on enter (aria-pressed="false"), it does not change on activation (to aria-pressed="true"), or upon unselecting (to aria-pressed="false"), making it hard to tell which have been selected. 

See:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

These issues are similar as elsewhere, but instead of a table it is a list, and instead of checkbox roles we have buttons.

ssciolla:
This PR aims to resolve #230 and #260.

Change(s):
1. Rather than relying on the provided `button` prop on `ListItem`, a `ButtonBase` was placed inside each `ListItem`. This follows the preferred approach identified by the accessibility team meant to resolve #260.
2.  Modified the `usePromise` hook for unmerging sections so an intermediate state variable (`sectionsToUnmerge`) and `useEffect` hook could be removed.
3. Made the text of the course name in a selected section blue, which I believe was the original intention, but wasn't happening before (reason...)
4. Used conditional rendering in a handful of instances (`props.multiSelect`, `props.search`, `props.header`), instead of using hidden styles (this resulted in some minor visual changes). I think this also had the side effect of fixing #230.
5. Broke up long lines for readability.

### DIFF
--- a/ccm_web/client/src/components/SectionSelectorWidget.tsx
+++ b/ccm_web/client/src/components/SectionSelectorWidget.tsx
@@ -35,12 +35,21 @@ const useStyles = makeStyles((theme) => ({
       }
     }
   },
+  listItemRoot: {
+    paddingTop: '0px',
+    paddingBottom: '0px'
+  },
   listButton: {
     width: '100%',
     height: '100%',
     textAlign: 'left',
     paddingLeft: theme.spacing(2),
-    paddingRight: theme.spacing(2)
+    paddingRight: theme.spacing(2),
+    paddingTop: theme.spacing(1),
+    paddingBottom: theme.spacing(1),
+    '&:hover': {
+      backgroundColor: theme.palette.action.hover
+    }
   },
   buttonFocusVisible: {
     backgroundColor: theme.palette.action.focus
@@ -513,9 +522,10 @@ function SectionSelectorWidget (props: ISectionSelectorWidgetProps): JSX.Element
             const isSelected = isSectionSelected(section.id)
             return (
               <ListItem
+                key={section.id}
                 divider
                 disableGutters
-                key={section.id}
+                classes={{ root: classes.listItemRoot }}
                 disabled={section.locked}
                 selected={isSelected}
                 className={(section.locked !== true && props.highlightUnlocked === true) ? classes.highlighted : ''}
--- a/ccm_web/client/src/components/SectionSelectorWidget.tsx
+++ b/ccm_web/client/src/components/SectionSelectorWidget.tsx
@@ -11,7 +11,7 @@ import { useDebounce } from '@react-hook/debounce'
 
 import { unmergeSections } from '../api'
 import usePromise from '../hooks/usePromise'
-import { CanvasCourseSectionWithCourseName, ICanvasCourseSectionSort } from '../models/canvas'
+import { CanvasCourseSectionBase, CanvasCourseSectionWithCourseName, ICanvasCourseSectionSort } from '../models/canvas'
 import { ISectionSearcher } from '../utils/SectionSearcher'
 
 const useStyles = makeStyles((theme) => ({
@@ -112,7 +112,7 @@ interface ISectionSelectorWidgetProps {
     sort?: { sorters: Array<{ func: ICanvasCourseSectionSort, text: string}>, sortChanged: (currentSort: ICanvasCourseSectionSort) => void }
   }
   canUnmerge: boolean
-  sectionsRemoved?: (sections: SelectableCanvasCourseSection[]) => void
+  sectionsRemoved?: (sections: CanvasCourseSectionBase[]) => void
   highlightUnlocked?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


ccm_web/client/src/components/SectionSelectorWidget.tsx
```diff
@@ -123,32 +138,23 @@ function SectionSelectorWidget (props: ISectionSelectorWidgetProps): JSX.Element
...
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      doUnmerge().then(() => {
-        setInternalSections(internalSections.filter(section => { return !sectionsToUnmerge.map(s2u => { return s2u.id }).includes(section.id) }))
-        if (props.sectionsRemoved !== undefined) {
-          props.sectionsRemoved(sectionsToUnmerge)
-        }
-        setSectionsToUnmerge([])
-      })
-    }
-  }, [sectionsToUnmerge])
```
ssciolla, line 151: See 1 in PR description.

ccm_web/client/src/components/SectionSelectorWidget.tsx
```diff
@@ -303,54 +309,60 @@ function SectionSelectorWidget (props: ISectionSelectorWidgetProps): JSX.Element
...
+          onClick={async (e) => {
+            e.stopPropagation()
+            await doUnmerge([section])
+          }}
+        >
+          Unmerge
+        </Button>
+      )
     }
   }
```
ssciolla, line 328: See 1 in PR description.

ccm_web/client/src/components/SectionSelectorWidget.tsx
```diff
@@ -391,101 +403,140 @@ function SectionSelectorWidget (props: ISectionSelectorWidgetProps): JSX.Element
...
+                      control={
+                        <Checkbox
+                          checked={isSelectAllChecked}
+                          onChange={handleSelectAllClicked}
+                          name='selectAllUnstagedCB'
+                          color='primary'
+                        />
+                      }
+                      disabled={selectableSections().length === 0}
+                      label='Select All'
```
ssciolla, line 500: See item 4 in PR description (conditional rendering).

ssciolla:
```
 // TODO display:none, visibility:hidden.. find a better way
```
@chrisrrowland, tagging you since I thought you might be somewhat interested in what I did here (as well as in the accessibility changes).

ccm_web/client/src/components/SectionSelectorWidget.tsx
```diff
@@ -391,101 +403,140 @@ function SectionSelectorWidget (props: ISectionSelectorWidgetProps): JSX.Element
...
+                <ButtonBase
+                  className={classes.listButton}
+                  onClick={() => handleListItemClick(section.id)}
+                  disabled={section.locked}
+                  aria-pressed={isSelected}
+                  focusVisibleClassName={classes.buttonFocusVisible}
+                >
+                  {listItemText(section)}
+                </ButtonBase>
+              </ListItem>
```
ssciolla, line 532: See Item 1 in PR description.

ccm_web/client/src/components/SectionSelectorWidget.tsx
pushyamig: Is this change needed as part of accessibility fix #260? or you felt this is simple/improved way of handing the click event so did it?

ccm_web/client/src/components/SectionSelectorWidget.tsx
pushyamig: @ssciolla, what do you mean by TODO here?

ccm_web/client/src/components/SectionSelectorWidget.tsx
ssciolla: Not part of the accessibility fix, just something that made sense to me.

ccm_web/client/src/components/SectionSelectorWidget.tsx
ssciolla: Chris had made this comment, which I'm now removing because the issue is resolved. Item 4 explains the topic. Basically Chris had been using `display: none` and `visilibity: hidden` CSS to hide parts of the component that were only sometimes used. Instead, I'm just not rendering those elements at all if the settings aren't passed to the component.

pushyamig:
@ssciolla This fix seem to be working fine. I know for the structure for the table component for screen readers, you shared this PR with Accessibility team and I guess they seems to like it. I don't have anymore feedback on that. 

This change seems knowing and unknowingly had few other fixes and code refactoring. Smoke testing all of them working fine. I tested the section widget component with Merge section and add UM user. The items on there `header`, `selectAll checkbox` and `search` is working. I think exhaustive testing might be not needed, just went by code logic.

pushyamig (approved):

ssciolla:
Thanks for the review, @pushyamig. I was looking over the code one more and time doing more testing, and I found a couple other things worth changing, as well as a margin inconsistency (extra margin around the `button` inside `li` with the component in Merge Sections because of a CSS class set in the `MergeSections.tsx` page. I moved things around a little and resolved it. If you could take a look at my latest commits, I'd appreciate it!

PR #337, opened 2022-02-09
Add table captions (#257)

### LINKED ISSUE #257: Table containing the verification rows does not have a caption.
Category: Semantics and Structure. Impact: major

Location: Confirmation screen: Gradebook formatter, Add UM user, Create sections

Issue
Table containing the verification rows does not have a caption.

Impact
Given the SPA nature of CCM, lack of caption will add cognitive burden. What happened? 

Solution
Provide a caption. Indicating the number of items returned would be valuable. Caption can be visually hidden if needed for design considerations

ssciolla:
The PR aims to resolve #257.

(I only did minimal testing, but I will do some more. This should do the trick.)

### DIFF
--- a/ccm_web/client/src/components/TableCaption.tsx
+++ b/ccm_web/client/src/components/TableCaption.tsx
@@ -1,0 +1,24 @@
+import React from 'react'
+import { makeStyles } from '@material-ui/core'
+
+const useStyles = makeStyles(() => ({
+  srOnly: {
+    position: 'absolute',
+    width: '1px',
+    height: '1px',
+    padding: '0',
+    margin: '-1px',
+    overflow: 'hidden',
+    clip: 'rect(0,0,0,0)',
+    border: '0'
+  }
+}))
+
+interface TableCaptionProps {
+  text: string
+}
+
+export default function TableCaption (props: TableCaptionProps): JSX.Element {
+  const classes = useStyles()
+  return <caption className={classes.srOnly}>{props.text}</caption>
+}
--- a/ccm_web/client/src/components/ValidationErrorTable.tsx
+++ b/ccm_web/client/src/components/ValidationErrorTable.tsx
@@ -50,7 +50,8 @@ function ErrorTable (props: ErrorTableProps): JSX.Element {
 
     return (
       <TableContainer component={Paper}>
-        <Table stickyHeader aria-label="custom pagination table">
+        <Table stickyHeader aria-label='custom pagination table'>
+          <caption>{props.invalidations.length} validation errors occurred.</caption>
           <TableHead>
               <TableRow>
                 {columns.map((column) => (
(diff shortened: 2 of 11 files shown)


ccm_web/client/src/components/TableCaption.tsx
```diff
@@ -0,0 +1,24 @@
...
+    position: 'absolute',
+    width: '1px',
+    height: '1px',
+    padding: '0',
+    margin: '-1px',
+    overflow: 'hidden',
+    clip: 'rect(0,0,0,0)',
+    border: '0'
+  }
+}))
```
ssciolla, line 15: I pulled this from `SectionSelectorWidget`. Maybe we can unify these, but the types of elements they are applied to are different (`caption` versus `span`).

ccm_web/client/src/components/ValidationErrorTable.tsx
```diff
@@ -50,7 +54,8 @@ function ErrorTable (props: ErrorTableProps): JSX.Element {
 
     return (
       <TableContainer component={Paper}>
-        <Table stickyHeader aria-label="custom pagination table">
+        <Table stickyHeader aria-label='custom pagination table'>
+          <TableCaption text={`${props.invalidations.length} validation errors occurred.`} />
           <TableHead>
               <TableRow>
                 {columns.map((column) => (
```
ssciolla, line 61: `ValidationErrorTable` could use `CustomTable`. This was a refactor discussed at one point but did not happen. It should be relatively straightforward if you want me to do it.

pushyamig: can the error message be something like `5 validation errors occurred on row 2,3,4,5,9`? 
I am not asking for the change if accessibility team suggested `5 validation errors occurred` is good info for table caption.

pushyamig: Sure

ssciolla:
> A generic question, again not asking for a change.
> 
> I read the problem statement, solution discussion with accessibility team. This fix is right one. When I turn on Voice over I don't see the table caption is being read. Is that expected?

Are you using Control + Option + A to read all the content on the page? I did test that it worked, but I could have missed something. My current understanding is that screenreader stuff doesn't need to be read automatically, just be present when the screenreader is told to read all the content.

pushyamig:
>  just be present when the screenreader is told to read all the content.
Probably I don't know how to trigger the Screen reader to read. I tried the Control + Option + A after you suggestion. But no luck. I am sure I don't now how to use it that's all. 

But on other note, I see captions are added as expected. I tested a few, but based on the file changes I can say it will work.

ssciolla:
Hi @pushyamig, I am able to get the table captions to be read by VoiceOver if I use the Control + Option + Command + T combination to locate the table. Please let me know if this works for you. I am following the recommendations of [Material UI](https://mui.com/components/tables/#caption) and [W3C](https://www.w3.org/WAI/tutorials/tables/caption-summary/) in terms of placement, and the hiding CSS should not affect whether it is available to screenreaders.

pushyamig:
> I am able to get the table captions to be read by VoiceOver if I use the Control + Option + Command + T combination to locate the table.

Now I can, thx for the tip on the key options to press. Working as expected.

pushyamig (approved):
Nice Work on adding Table Caption. I really like the design of on table captions and surprised screen reader have different key option to read the table.

ccm_web/client/src/components/ValidationErrorTable.tsx
ssciolla: I'll do this separately.

PR #340, opened 2022-02-14
Report API errors consistently, with and without snackbar (#272)

### LINKED ISSUE #272: Snack bar error message should show the API response reasoning for failure.

ssciolla:
This PR introduces a component called `APIErrorMessage` that is responsible for handling the error parsing and text generation for all errors originating from the API (both our application and Canvas). Other changes were made to further refine the error handling strategy in the client (see Notes below). The PR aims to resolve #272.

**Notes**
- `handleErrors.ts` was tweaked and expanded to include a `BadRequestError`, handle non-JSON errors (often HTML, like those from `ngrok` when the server isn't running), and reduce some code duplication.
- `APIErrorWithContext` was introduced for simplifying reporting of errors in components/pages with multiple possible errors sources. (The interface is identical to `APIErrorMessageProps`, which makes passing data for presentation simple.)
- Existing error reporting in `App.tsx` for `api.getCourse` was replaced with usage of `ErrorAlert` + `APIErrorMessage`.


ccm_web/client/src/components/CreateSectionWidget.tsx
```diff
@@ -54,18 +56,26 @@ function CreateSectionWidget (props: CreateSectionWidgetProps): JSX.Element {
...
             enqueueSnackbar('Section was created!', { variant: 'success' })
-          }).catch(() => {
-            enqueueSnackbar('Error adding section', {
-              variant: 'error'
-            })
+          }).catch((error: unknown) => {
+            enqueueSnackbar(
+              <APIErrorMessage
+                context='creating the section'
+                error={error instanceof Error ? error : undefined}
```
ssciolla, line 63: This will almost always be an `Error`. In the case where it isn't, the user will see a generic message, and we can ask them to look in the Web console for more information if we need to investigate.

ccm_web/client/src/pages/FormatThirdPartyGradebook.tsx
```diff
@@ -132,12 +133,6 @@ export default function FormatThirdPartyGradebook (props: FormatThirdPartyGradeb
     )
   }
 
-  useEffect(() => {
-    if (file !== undefined) {
-      parseFile(file)
-    }
-  }, [file])
-
```
ssciolla, line 140: This isn't related to the issue, but this hook isn't really needed, and it's something I've factored out in other pages/components. We can just use the `onUploadComplete` to set the file in state and call `parseFile`.

ccm_web/client/src/components/CourseSectionList.tsx
```diff
@@ -1,9 +1,15 @@
-import { Backdrop, Button, CircularProgress, Grid, List, ListItem, ListItemText, makeStyles, Typography } from '@material-ui/core'
+import { useSnackbar } from 'notistack'
 import React, { useEffect, useState } from 'react'
+import {
+  Backdrop, Button, CircularProgress, Grid, List, ListItem, ListItemText, makeStyles, Typography
+} from '@material-ui/core'
+
+import APIErrorMessage from './APIErrorMessage'
```
pushyamig, line 7: The imported module is actually names like this `ApiErrorMessage`. Here it is pulled as `./APIErrorMessage`. Currently there is compilation error because of this. 
```
[tsl] ERROR in /base/client/src/App.tsx(7,29)
ccm_web     |       TS1261: Already included file name '/base/client/src/components/APIErrorMessage.tsx' differs from file name '/base/client/src/components/ApiErrorMessage.tsx' only in casing.
ccm_web     |   The file is in the program because:
ccm_web     |     Imported via './components/APIErrorMessage' from file '/base/client/src/App.tsx'
ccm_web     |     Imported via './APIErrorMessage' from file '/base/client/src/components/MultipleUserEnrollmentWorkflow.tsx'
ccm_web     |     Imported via './APIErrorMessage' from file '/base/client/src/components/CreateSectionWidget.tsx'
```
I tried to rename the file and error went away. But I wanted to Open a PR but somehow the `git status` don't show as a change, which is strange. So I think renaming of the file consistent with imported file is needed.

ssciolla: Ah, might have missed that. I'll take a look. I tried a few different names.

pushyamig:
I will finish the review tomorrow. This branch has merge conflicts if you can resolve that would be great.

ccm_web/client/src/components/CourseSectionList.tsx
ssciolla: Weird, it looks like I did change the file name locally, but it wasn't reflected in Git. Looks like something related to this, which I haven't come across before: https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git. The first solution worked for me.

ssciolla:
> I will finish the review tomorrow. This branch has merge conflicts if you can resolve that would be great.

Done.

ccm_web/client/src/components/CourseSectionList.tsx
pushyamig: Yes, That's what I did in past I renamed the file in the Github. Never tried the git console command.

ccm_web/client/src/components/UserEnrollmentForm.tsx
```diff
@@ -261,19 +257,7 @@ export default function UserEnrollmentForm (props: UserEnrollmentFormProps): JSX
...
-        message = (
-          <Typography>
-            {prefix}: {error.errors.length > 0 ? error.errors[0].message : 'unknown API error from Canvas.'}
-          </Typography>
-        )
-      } else {
-        message = <Typography key={0}>{prefix}: {error.message}.</Typography>
-      }
-      return <ErrorAlert messages={[message]} tryAgain={resetAll} />
+      return <ErrorAlert messages={[<APIErrorMessage key={0} {...errorsWithContext[0]} />]} tryAgain={resetAll} />
```
pushyamig, line 260: Do you think, you need to revisit this when you integrate it checking for users exists or not for Non-UM feature?  there might be some change in Error response you were designing with this feature.

ccm_web/client/src/components/APIErrorMessage.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface APIErrorMessageProps {
+  context: string
+  error: Error | undefined
+}
+
+const extractErrorText = (error: Error): string[] => {
+  if (error instanceof CanvasError) {
+    return error.errors.map(e => e.message)
+  } else {
+    return [error.message]
```
pushyamig, line 15: This logic seems to be checking for various error format from backend, Canvas. Do you think this needs to be revisited again from the merging of add Non-UM user feature. Since The Backend error Object **Might** changed?

ccm_web/client/src/pages/FormatThirdPartyGradebook.tsx
pushyamig: OK, I test this and things are working as expected.

ccm_web/client/src/utils/handleErrors.ts
```diff
@@ -52,31 +59,31 @@ const handleErrors = async (resp: Response): Promise<void> => {
...
+  }
+  if (isCanvasAPIErrorData(errorBody)) throw new CanvasError(errorBody.errors)
 
+  const apiErrorMessage = Array.isArray(errorBody.message) ? errorBody.message.join(' ') : errorBody.message
   switch (resp.status) {
+    case 400:
+      throw new BadRequestError(apiErrorMessage)
     case 401:
-      if (errorBody.redirect === true) redirect('/')
       throw new UnauthorizedError()
```
pushyamig, line 80: is there any reason why you moved this code above L69-72? I felt having that code here is clean.

ccm_web/client/src/utils/handleErrors.ts
```diff
@@ -52,31 +59,31 @@ const handleErrors = async (resp: Response): Promise<void> => {
...
-  const errorBody: APIErrorData = JSON.parse(text)
+  let errorBody: APIErrorData
+  try {
+    errorBody = JSON.parse(text)
+  } catch (error) {
+    throw new Error(`Non-JSON error encountered with status code ${resp.status}`)
+  }
+
+  if (resp.status === 401 && errorBody.redirect === true) {
+    redirect('/')
```
pushyamig, line 70: Does redirect to root, will make LTI launch happen?

ccm_web/client/src/components/ApiErrorMessage.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export default function APIErrorMessage (props: APIErrorMessageProps): JSX.Element {
+  const errorMessages = props.error !== undefined ? extractErrorText(props.error) : []
+  const preface = (
+    `Error${errorMessages.length > 1 ? 's' : ''} occurred ` +
+    `while ${props.context}${props.error !== undefined ? ':' : '.'}`
+  )
+  return <Typography>{preface}&nbsp;{errorMessages.join('; ')}</Typography>
+}
```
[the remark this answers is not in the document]
pushyamig: Sure. Thanks for adding context for your design `APIErrorMessage`

ccm_web/client/src/components/UserEnrollmentForm.tsx
ssciolla: Mmm... probably. That's the biggest question mark right now, is how do I handle the irregular data shape from errors. I think adding a type guard and creating a new error subclass (`ExternalUserError`?) would probably work, but I'll have to give it some thought. The internals of `APIErrorMessage` would probably need to be changed then, but at least that's just a change in one place.

pushyamig:
I did test couple of error responses from this change, and not the full list of changes made per this PR. I have few comment for clarification but nothing to stop this from getting merged.

ccm_web/client/src/components/UserEnrollmentForm.tsx
pushyamig: Ok, I won't vet on this now. If you want to revisit this when you are integrating the Non-UM user then I am fine with it.

ccm_web/client/src/utils/handleErrors.ts
ssciolla: Hm, that might work. I don't quite remember why I did this.

ccm_web/client/src/components/UserEnrollmentForm.tsx
pushyamig: > I think adding a type guard and creating a new error subclass (ExternalUserError?)

This is good idea, since you know that the external user case is slightly different and easier for you to write code based on type of  error object.

ccm_web/client/src/utils/handleErrors.ts
pushyamig: It won't be an LTI launch. It's just a redirect to Home page.

ccm_web/client/src/components/APIErrorMessage.tsx
pushyamig: I think we can merge this code as is but when it is time to integrate with Non-UM UI with Backend, this component  can be adjusted. I don't think I want to put that as criteria for merging this.

ccm_web/client/src/utils/handleErrors.ts
ssciolla: Redirecting to root will basically reload the application, and then when `globals` shows they have no canvas token, they are shown the authorize prompt. We could change this logic somewhat and redirect to a new authorize page directly.

ccm_web/client/src/utils/handleErrors.ts
ssciolla: Okay, I moved it back: https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/340/commits/64c22681557910fa994d721dafc7e38419aa5f91

pushyamig (approved):

ccm_web/client/src/components/APIErrorMessage.tsx
ssciolla: As I told you in the meeting, I may not use this component, or only partially use it for the Add Non UM Users integration; we'll see how it shakes out.

PR #342, opened 2022-02-15
#341 alter user that first/last name are required

### LINKED ISSUE #341: Alert user that FirstName and LastName fields are required
Currently after the user types in email address and if the user doesn't exist then first name, last name elements shown, if user hit search button with out filing then no error alert is given to the user as them a required fields

pushyamig:
Fixes #341 

First name and last name are required by API endpoint, so I felt adding the alter saying it is required

### DIFF
--- a/ccm_web/client/src/components/UserEnrollmentForm.tsx
+++ b/ccm_web/client/src/components/UserEnrollmentForm.tsx
@@ -138,7 +138,7 @@ export default function UserEnrollmentForm (props: UserEnrollmentFormProps): JSX
     if (
       userExists === undefined ||
       role === undefined ||
-      selectedSection === undefined
+      selectedSection === undefined || firstName === undefined || lastName === undefined
     ) {
       return setShowIncompleteAlerts(true)
     }
@@ -215,6 +215,10 @@ export default function UserEnrollmentForm (props: UserEnrollmentFormProps): JSX
 
   const nameInput = (
     <>
+    {
+        showIncompleteAlerts && (firstName === undefined || lastName === undefined) &&
+          <InlineErrorAlert>You must fill in firstname or lastName.</InlineErrorAlert>
+    }
     <Typography className={classes.spacing} gutterBottom>
       The email you entered is not associated with an account in Canvas.
       Please provide a first and last name, and we will send them an email invitation


ccm_web/client/src/components/UserEnrollmentForm.tsx
```diff
@@ -215,6 +215,10 @@ export default function UserEnrollmentForm (props: UserEnrollmentFormProps): JSX
 
   const nameInput = (
     <>
+    {
+        showIncompleteAlerts && (firstName === undefined || lastName === undefined) &&
+          <InlineErrorAlert>You must fill in firstname or lastName.</InlineErrorAlert>
+    }
```
ssciolla, line 221: I feel like this is already kinda handled by the field validation? If you try to hit enter with these, it should tell you that you're not allowed to leave it blank. I understand that's not the same pattern, but it does effectively do this.

ccm_web/client/src/components/UserEnrollmentForm.tsx
```diff
@@ -138,7 +138,7 @@ export default function UserEnrollmentForm (props: UserEnrollmentFormProps): JSX
     if (
       userExists === undefined ||
       role === undefined ||
-      selectedSection === undefined
+      selectedSection === undefined || firstName === undefined || lastName === undefined
```
ssciolla, line 141: This might work functionally, but it's not quite logical, as `firstName` and `lastName` aren't required in all cases (and we don't know if they will be required, since `userExists` hasn't been populated).

ssciolla (commented):
Can you give some context to this change? Was it requested by the UX team? Leaving a couple comments for now.

pushyamig:
> Can you give some context to this change? Was it requested by the UX team? Leaving a couple comments for now.

I felt this was bug so I opened the change.

ccm_web/client/src/components/UserEnrollmentForm.tsx
pushyamig: When you search for an email address and if it doesn't exists then First/Last Name component appear and at that point if user hits the Submit button, role and section section widget will have an error alert but not the first name and last name component. So I felt it was missed so i fixed it

ssciolla: Okay, let me do some testing. I may suggest a slight change.

pushyamig: As long as the first name and last name component is not displayed(in case of user already exists) the alert won't be appearing. So I feel this validation is still valid.

ccm_web/client/src/components/UserEnrollmentForm.tsx
ssciolla: So this is actually a new bug. If the user does exist, this addition actually prevents the user from submitting the form, even though they have completed it and nothing on the page indicates they still need to something. I believe this needs to be changed to this, so that this condition fails if `userExists === true` and everything else (but the hidden first and last name) are finished.
```
    if (
      userExists === undefined ||
      (!userExists && firstName === undefined && lastName === undefined) ||
      role === undefined ||
      selectedSection === undefined
    ) {
      return setShowIncompleteAlerts(true)
    }
    setShowIncompleteAlerts(false)
```

ccm_web/client/src/components/UserEnrollmentForm.tsx
ssciolla: Okay, I agree that that is a good addition/fix to the flow. Just need to correct the logic.

ssciolla: You had an extra indentation, and I think I've improved the language.
```suggestion
    {
      showIncompleteAlerts && (firstName === undefined || lastName === undefined) &&
        <InlineErrorAlert>You must enter a first name and last name.</InlineErrorAlert>
    }
```

ssciolla (changes requested):

ccm_web/client/src/components/UserEnrollmentForm.tsx
pushyamig: OK, I have changed the logic. I did not gave a deep thought if the logic is logical :) but it was working so there I go. 
This will help code readability!

ssciolla (approved):

PR #345, opened 2022-02-17
Add root route and database health checks using `@nestjs/terminus` (#343)

### LINKED ISSUE #343: Add health check mechanism for use in production
With our other T&L applications, we have set up a route using something like [`django-watchman`](https://pypi.org/project/django-watchman/) which indicates the health of the application and its dependencies (like a database). This can then be used by another service to monitor the application's status. We need some equivalent from NodeJS land. 

It looks NestJS offers a package and recipe for this purpose: https://docs.nestjs.com/recipes/terminus#setting-up-a-healthcheck

ssciolla:
This PR aims to resolve #343.

Resource(s):
- https://docs.nestjs.com/recipes/terminus#healthchecks-terminus
- https://github.com/nestjs/terminus

### DIFF
--- a/ccm_web/server/src/health/health.controller.ts
+++ b/ccm_web/server/src/health/health.controller.ts
@@ -1,0 +1,27 @@
+import { Controller, Get } from '@nestjs/common'
+import { ConfigService } from '@nestjs/config'
+import {
+  HealthCheck, HealthCheckResult, HealthCheckService, HttpHealthIndicator, SequelizeHealthIndicator
+} from '@nestjs/terminus'
+
+import { Config } from '../config'
+
+@Controller('health')
+export class HealthController {
+  constructor (
+    private readonly configService: ConfigService<Config, true>,
+    private readonly health: HealthCheckService,
+    private readonly http: HttpHealthIndicator,
+    private readonly db: SequelizeHealthIndicator
+  ) {}
+
+  @Get()
+  @HealthCheck()
+  async check (): Promise<HealthCheckResult> {
+    const appDomain = this.configService.get('server.domain', { infer: true })
+    return await this.health.check([
+      async () => await this.http.pingCheck('root', `https://${appDomain}`),
+      async () => await this.db.pingCheck('database', { timeout: 3000 })
+    ])
+  }
+}
--- a/ccm_web/server/src/health/health.module.ts
+++ b/ccm_web/server/src/health/health.module.ts
@@ -1,0 +1,11 @@
+import { Module } from '@nestjs/common'
+import { ConfigModule } from '@nestjs/config'
+import { TerminusModule } from '@nestjs/terminus'
+
+import { HealthController } from './health.controller'
+
+@Module({
+  controllers: [HealthController],
+  imports: [ConfigModule, TerminusModule]
+})
+export class HealthModule {}
--- a/ccm_web/package-lock.json
+++ b/ccm_web/package-lock.json
@@ -22,6 +22,7 @@
         "@nestjs/sequelize": "^8.0.0",
         "@nestjs/serve-static": "^2.2.2",
         "@nestjs/swagger": "^5.1.4",
+        "@nestjs/terminus": "^8.0.4",
         "@react-hook/debounce": "^4.0.0",
         "axios": "^0.24.0",
         "class-transformer": "^0.4.0",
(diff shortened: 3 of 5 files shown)


ccm_web/server/src/health/health.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+    private readonly db: SequelizeHealthIndicator
+  ) {}
+
+  @Get()
+  @HealthCheck()
+  async check (): Promise<HealthCheckResult> {
+    const appDomain = this.configService.get('server.domain', { infer: true })
+    return await this.health.check([
+      async () => await this.http.pingCheck('root', `https://${appDomain}`),
+      async () => await this.db.pingCheck('database', { timeout: 3000 })
```
pushyamig, line 24: I don't think the DB connection is established to make this ping work. The DB connection needs to be give to  make that check. So you usually do it at `health.module.ts` but as of now DB connection is not established. 
More something like this should be done https://github.com/nestjs/terminus/blob/master/sample/007-sequelize-app/src/health/health.module.ts

ccm_web/server/src/health/health.module.ts
```diff
@@ -0,0 +1,11 @@
+import { Module } from '@nestjs/common'
+import { ConfigModule } from '@nestjs/config'
+import { TerminusModule } from '@nestjs/terminus'
+
+import { HealthController } from './health.controller'
+
+@Module({
+  controllers: [HealthController],
+  imports: [ConfigModule, TerminusModule]
```
pushyamig, line 9: the Sequelize module needs to be imported in order to make this work. 
https://github.com/nestjs/terminus/blob/master/sample/007-sequelize-app/src/health/health.module.ts

ccm_web/server/src/health/health.controller.ts
ssciolla: @pushyamig, I'm not sure about that. Based on my understanding of the docs on the `SequelizeModule` (https://docs.nestjs.com/techniques/database#sequelize-integration), you shouldn't need to import the `SequelizeModule` in other modules if it's in the `AppModule`.

Furthermore, I am seeing SQL emitted from `sequelize` (`SELECT 1`) when I hit this endpoint.

ssciolla: Yeah, I saw this too. I'm not sure what to make of it, or how else to test the database being down.

ccm_web/server/src/health/health.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+    private readonly http: HttpHealthIndicator,
+    private readonly db: SequelizeHealthIndicator
+  ) {}
+
+  @Get()
+  @HealthCheck()
+  async check (): Promise<HealthCheckResult> {
+    const appDomain = this.configService.get('server.domain', { infer: true })
+    return await this.health.check([
+      async () => await this.http.pingCheck('root', `https://${appDomain}`),
```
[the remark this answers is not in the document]
ssciolla: Mmm, maybe. It may be possible that just the static serving goes down, but I think in most cases you are correct.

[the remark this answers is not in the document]
pushyamig: I think this is ok, Since nagios might be looking at server URL not via API endpoint.

ccm_web/server/src/health/health.controller.ts
pushyamig: @jonespm The test I did is a valid way of looking if DB up/down?

pushyamig:
@ssciolla At this point I feel we can create a separate issue when DB is shutdown why the error is throws instead of stating up/down. At this moment I don't think we can do much if the library is not supporting it. The integration is working as per the recipes by the Author.  
I think we can merge this issue and get the Nagios configuration process by Rick.

pushyamig (approved):

ssciolla:
@pushyamig, we can merge this if you want to. I'm still seeing bizarre behavior though. Specifically, I'm seeing errors like this one, which are often masking a `SequelizeConnectionError`:

```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:371:5)
    at ServerResponse.setHeader (node:_http_outgoing:576:11)
    at ServerResponse.header (/base/node_modules/express/lib/response.js:771:10)
    at ServerResponse.send (/base/node_modules/express/lib/response.js:170:12)
    at ServerResponse.json (/base/node_modules/express/lib/response.js:267:15)
    at ExpressAdapter.reply (/base/node_modules/@nestjs/platform-express/adapters/express-adapter.js:36:57)
    at CSRFExceptionFilter.handleUnknownError (/base/node_modules/@nestjs/core/exceptions/base-exception-filter.js:38:24)
    at CSRFExceptionFilter.catch (/base/node_modules/@nestjs/core/exceptions/base-exception-filter.js:17:25)
    at CSRFExceptionFilter.catch (/base/server/src/auth/csrf.exception.filter.ts:32:18)
    at ExceptionsHandler.invokeCustomFilters (/base/node_modules/@nestjs/core/exceptions/exceptions-handler.js:33:26) {
  code: 'ERR_HTTP_HEADERS_SENT'
```

This ~~might~~ be a bug somewhere else in the application, but I haven't been able to figure out where.

That said, I was able to reproduce this on the `main` branch, so it's not exactly a new issue.

pushyamig:
>Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client

I sometime see this error when I am debugging in VSCode when I hit a break point and resume. I never saw this during the normal flow of execution. 

I think we can merge this for now.

PR #355, opened 2022-02-23
#354,346 expand assistant privileges and csv download improvements

pushyamig:
Fixes #354 #346

PR #364, opened 2022-03-07
I346 csv indentation

### LINKED ISSUE #346: Bulk Add User Sample CSV improvements
This probably applies to both non-umich and umich user sample CSVs.

The existing CSV for UM users has the sample data is reversed (under the wrong headers). I think it would be fine to have just one sample data row for a user with student role rather than all the roles. Users of this tool can reference the documentation should they want to know what other roles are available for use with this tool.

The same can be true for non-umich users sample download file.

pushyamig:
Fixes #346 

This PR is fixing the indentation problem that was not fixed PR #355

### DIFF
--- a/ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
+++ b/ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
@@ -214,7 +214,7 @@ export default function SingleSectionEnrollmentWorkflow (props: SingleSectionEnr
 
     const fileData =
     `${REQUIRED_ENROLLMENT_HEADERS.join(',')}\n` +
-    'studentone, student\n'
+    'studentone,student\n'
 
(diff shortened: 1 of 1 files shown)


ccm_web/client/src/components/SingleSectionEnrollmentWorkflow.tsx
```diff
@@ -213,8 +213,8 @@ export default function SingleSectionEnrollmentWorkflow (props: SingleSectionEnr
     if (rowInvalidations !== undefined) return renderRowValidationErrors(rowInvalidations)
 
     const fileData =
-    `${REQUIRED_ENROLLMENT_HEADERS.join(',')}
-    studentone, student`
+    `${REQUIRED_ENROLLMENT_HEADERS.join(',')}\n` +
+    'studentone, student\n'
```
ssciolla, line 217:
```suggestion
    'studentone,student\n'
```

ssciolla (approved):

PR #366, opened 2022-03-23
Removed the extra ping with App

### LINKED ISSUE #343: Add health check mechanism for use in production

pushyamig:
Fixes #343 

This issue is removing the redundant check against the APP URL since Nagios has there own way to check app up or down.

I have the PR on ccm-dev instance and Rick confirmed this change don't impact the Nagios configuration. 

I did not create an issue but tagging it to original issue with Health check

### DIFF
--- a/ccm_web/server/src/health/health.controller.ts
+++ b/ccm_web/server/src/health/health.controller.ts
@@ -1,26 +1,19 @@
 import { Controller, Get } from '@nestjs/common'
-import { ConfigService } from '@nestjs/config'
 import {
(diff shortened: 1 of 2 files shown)


ssciolla (approved):
LGTM

PR #368, opened 2022-03-25
Extract core setup to provide basic fix to end-to-end test (#237)

### LINKED ISSUE #237: Fix end-to-end test
We have one fairly basic end-to-end test right now (`ccm_web/server/test/app.e2e-spec.ts`), which is broken because of (I believe) a session middleware configuration issue. I have done some work to fix this, doing some refactoring in `main.ts` so some code can be shared. I'll open a separate PR for this.

ssciolla:
This PR moves the core setting up of middleware and options (needed by both the main process and testing processes) into a separate function, which is then called in both `main.ts` and `app.e2e-spec.ts`. The PR aims to resolve #237.

Reminder of testing instructions:
1) Start up application using `docker compose up`
2) Enter running container: `docker exec -it ccm-web /bin/bash`
3) Run end-to-end test: `npm run test:e2e`

### DIFF
--- a/ccm_web/server/src/main.ts
+++ b/ccm_web/server/src/main.ts
@@ -4,7 +4,7 @@ import { urlencoded, json } from 'express'
 import session from 'express-session'
 import morgan from 'morgan'
 import { Sequelize } from 'sequelize-typescript'
-import { ValidationPipe } from '@nestjs/common'
+import { INestApplication, ValidationPipe } from '@nestjs/common'
 import { ConfigService } from '@nestjs/config'
 import { NestFactory } from '@nestjs/core'
 import { NestExpressApplication } from '@nestjs/platform-express'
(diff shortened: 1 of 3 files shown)


ccm_web/server/test/app.e2e-spec.ts
```diff
@@ -1,30 +1,33 @@
...
       imports: [AppModule]
     }).compile()
 
     app = moduleFixture.createNestApplication()
-    app.use(cookieParser('SOME_COOKIE_SECRET'))
-    app.useGlobalPipes(new ValidationPipe())
-
+    const configService = app.get<ConfigService<Config, true>>(ConfigService)
+    const serverConfig = configService.get('server', { infer: true })
+    doAppCoreSetup(app, serverConfig)
```
ssciolla, line 21: I think this should be OK for the current test, but at some point we'd want to modify this so that we mock or modify `Config` so that we can use a test database and make other modifications.

ccm_web/server/src/main.ts
```diff
@@ -4,47 +4,40 @@ import { urlencoded, json } from 'express'
...
+  // Controls size limit of data in payload and URL
+  const payloadSizeLimit = '5mb'
+  app.use(json({ limit: payloadSizeLimit }))
+  app.use(urlencoded({ extended: true, limit: payloadSizeLimit }))
 
+  // Cookies and Sessions
   app.use(cookieParser(serverConfig.cookieSecret))
 
+  const sequelize = app.get(Sequelize)
   const SequelizeStore = ConnectSessionSequelize(session.Store)
```
pushyamig, line 37: Any reason for change of ordering?

ccm_web/server/test/app.e2e-spec.ts
pushyamig: Yeah... I agree

pushyamig (approved):
Thanks for being determined to fix this.

ccm_web/server/src/main.ts
ssciolla: Not too much, I guess I figured I'd just get `sequelize` when it was needed, kinda keeps the related code together. I could see either way (early set up, or set up just in time) being okay.

PR #369, opened 2022-03-25
Use more DTOs

ssciolla:
This PR adopts DTOS in two new places (queries) to ensure `ValidationPipe` settings are in place.

### DIFF
--- a/ccm_web/server/src/canvas/canvas.interfaces.ts
+++ b/ccm_web/server/src/canvas/canvas.interfaces.ts
@@ -1,3 +1,4 @@
+import { CanvasOAuthReturnQueryDto } from './dtos/canvas.oauth.query.dto'
 import { hasKeys } from '../typeUtils'
 
 // OAuth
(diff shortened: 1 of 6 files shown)


ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -1,17 +1,35 @@
...
 export interface OAuthGoodResponseQuery {
   code: string
   state: string
+  error: undefined
+  error_description: undefined
 }
+
 export interface OAuthErrorResponseQuery {
   error: string
+  code: undefined
```
pushyamig, line 15: why the type is undefined here?

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -1,17 +1,35 @@
...
   error_description?: string
   state?: string
 }
 
+export const isOAuthGoodResponseQuery = (value: CanvasOAuthReturnQueryDto): value is OAuthGoodResponseQuery => {
+  return (
+    value.code !== undefined &&
+    value.state !== undefined &&
+    value.error === undefined &&
+    value.error_description === undefined
```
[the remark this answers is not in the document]
ssciolla: I haven't tested all possibilities, but sometimes it's not provided. I don't think I've seen empty strings.

ccm_web/server/src/canvas/canvas.interfaces.ts
ssciolla: I guess I was just trying to make explicit that we expect `code` to not be defined when there is an error.

ssciolla: I don't think so in this case? Since if it's an error, we expect it to be `undefined`. Making it `any` would negate type checking on that property, and `unknown` I think would be similar to not including it in the interface at all.

ccm_web/server/src/canvas/canvas.interfaces.ts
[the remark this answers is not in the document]
ssciolla: Sure. I don't think this is very different functionally from what we had before, we're just now preventing people from specifying arbitrary other query parameters.

PR #370, opened 2022-03-25
I300 remove the session expiration check

### LINKED ISSUE #300: When session expires, user may not be able to initiate new session via LTI launch
Something like this was reported by Dan, but I have also experienced it on occasion as well. Currently, sessions expire after 24 hours. When this occurs, sometimes (or all the time, this needs more testing) an error is thrown in the console but subsequent attempts to relaunch the application via the external tool link in Canvas fail. Instead they see the standard unauthenticated view.

Users should be able initiate a new session by launching the LTI tool again when this occurs. A better error message than the generic authentication failure would also be helpful.

This needs more research.

pushyamig:
Fixes #300

### DIFF
--- a/ccm_web/client/src/utils/handleErrors.ts
+++ b/ccm_web/client/src/utils/handleErrors.ts
@@ -23,7 +23,7 @@ class UnauthorizedError extends Error {
   constructor () {
     super(
       'You are not authorized to perform that action. ' +
-      'You may need to authenticate differently.'
+      'If you believe this is an error, tru re-launching the application or contact support.'
     )
   }
 }
--- a/ccm_web/client/src/utils/handleErrors.ts
+++ b/ccm_web/client/src/utils/handleErrors.ts
@@ -18,20 +18,22 @@ class BadRequestError extends Error {
   }
 }
 
+const authzErrorMessage = (
+  'You are not authorized to perform that action. ' +
+  'Try re-launching the application, or contact support.'
+)
+
 class UnauthorizedError extends Error {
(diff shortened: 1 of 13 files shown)


ssciolla (approved):
I think this works! Thanks for taking the lead on this, I know it was frustrating. I should have thought through the `SessionGuard` implementation more when I wrote it originally; I guess I didn't have the best understanding of `express-session` yet.

Scenarios I tested:
1) Regular expiration of JWT and session (setting in `MAX_AGE_IN_SEC` to a low number). Re-launching fixes the problem, and the error message makes sense.
2) Deleting the session row in the database. Re-launching fixes the problem, and the error message makes sense for GET calls.
3) Modifying the session expiration so it expires before the JWT. Re-launching works, and error makes sense for GET calls. 

I opened a PR (https://github.com/pushyamig/canvas-course-manager-next/pull/2) with some languages changes, and also some additional handling so POST requests give better error messages in scenarios 2 and 3 (CSRF middleware precedes the Session Guard in execution, so I opted to fix messages).

ccm_web/client/src/utils/handleErrors.ts
```diff
@@ -23,7 +23,7 @@ class UnauthorizedError extends Error {
   constructor () {
     super(
       'You are not authorized to perform that action. ' +
-      'You may need to authenticate differently.'
+      'If you believe this is an error, try re-launching the application or contact support.'
```
ssciolla, line 26: Note that I removed the "If ..." part in my PR. Didn't seen necessary, seems better to be concise here.

Issue #372, opened 2022-03-29
Error message flashed in confusing way before redirect for OAuth


ssciolla:
Currently, if the user manually removes the tool's integration with Canvas from the Canvas side (in the user's Settings), then tries to reload CCM (or make an API call on a preloaded client), an error message will flash briefly before the user is redirected to the root (`App.tsx`) and then shown the authorize text (`AuthorizePrompt.tsx`). This is because an `UnauthorizedError` is ultimately thrown, but the redirect (`window.href = '/'`) takes a couple moments to take effect.

Unfortunately, because the `redirect` occurs in a plain JS/TS utility function, rather than a React component, it's not simple to change what the user sees without also modifying error handling, and making changes in many places.

One possible solution would be to delay a second or two after redirecting before resolving the function (`handleErrors` is `async`), which would likely hide the error, though the user may see loading content flicker.

https://user-images.githubusercontent.com/35741256/160708551-e8a7e636-4ed7-4885-a4dc-e6e27ed9f98f.mov

PR #374, opened 2022-03-30
Use an imperative validation message with inputs when the field is blank (#372)

### LINKED ISSUE #371: Change error messages with field input to use directive
Text below from @janelilr:

Update error message when a user has to type in an input. For example:

**Adding a non-UM user’s (one at a time) first and last name.**
Current error message: “Value for the last name may not be blank.”
New error message: “Please enter last name”
Current error message: “Value for the first name may not be blank.”
New error message: “Please enter first name”

**Editing course name**
Current error message: “Value for the course name may not be blank.”
New error message: “Please enter the course name”

**For Add U-M users, Add non-UM users, Format Third-Party Gradebook, and Create Sections features.**
The current CSV file error messages can stay.

ssciolla:
The PR aims to resolve #371.


pushyamig (approved):
Changes make sense to me.

PR #376, opened 2022-04-04
Make threads per container configurable (#375)

### LINKED ISSUE #375: Make number of `pm2` threads configurable
For flexibility in testing and deployment related to memory limits and performance, it makes sense to make the number of threads per pod or container. This is set in `ecosystem.config.js`, and we can use a similar pattern as we already do for `PM2_MEM_LIMIT`.

ssciolla:
This PR aims to resolve #375.

### DIFF
--- a/config/.env.sample
+++ b/config/.env.sample
@@ -77,8 +77,8 @@ HELP_URL=https://github_account_name.github.io/canvas-course-manager-next
 # These are only used with docker-compose-prod.yml and in OpenShift.
 # See ecosystem.config.js for pm2-related usage and doc links.
 # Memory limit for pm2 processes (optional)
-# Maps to "max_memory_restart"; don't use decimals (i.e. use 1800M over 1.8G)
+# This maps to "max_memory_restart"; don't use decimals (i.e. use 1800M over 1.8G).
 # PM2_MEM_LIMIT=1G
 # Number of thread instances per container/pod (optional)
-# Maps to "instances"; it can be "max" or an integer.
+# This maps to "instances"; it can be "max" or an integer.
 # PM2_INSTANCES=max


ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -113,6 +114,9 @@ export class AdminApiHandler {
     )
     const coursesWithSectionsResult = await Promise.all(coursesWithSectionsApiPromises)
     const finalResult = makeResponse<CourseWithSections>(coursesWithSectionsResult)
+    if (!isAPIErrorData(finalResult)) {
+      logger.debug(`Number of sections returned: ${finalResult.length}`)
+    }
```
ssciolla, line 119: The log message changes in this file are not directly related to the issue, but I think they are somewhat useful in diagnosing the issue we're working on that also prompted the configuration change. I can remove them if you'd like.

config/.env.sample
```diff
@@ -72,8 +72,13 @@ DB_NAME=ccm
...
 # Deployment variables
 # These are only used with docker-compose-prod.yml and in OpenShift.
+# See ecosystem.config.js for pm2-related usage and doc links.
 # Memory limit for pm2 processes (optional)
-# See ecosystem.config.js for usage and doc links; don't use decimals (i.e. use 1800M over 1.8G).
+# This maps to "max_memory_restart"; don't use decimals (i.e. use 1800M over 1.8G).
 # PM2_MEM_LIMIT=1G
+# Number of thread instances per container/pod (optional)
+# This maps to "instances"; it can be "max" or an integer.
+# PM2_INSTANCES=max
```
pushyamig, line 84: So the max is 5 threads correct?

ccm_web/ecosystem.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     name: 'CCM - NestJS app',
     script: 'server/src/main.js',
     exec_mode: 'cluster',
-    instances: 'max',
+    instances: process.env.PM2_INSTANCES || 'max',
```
pushyamig, line 11: I think it make sense to have `max` set locally since it will speed up the req/res locally. But shouldn't we default what were configuring on Prod instance?

ssciolla: That's an interesting question. I guess I like having to set it again in prod so that it is made more explicit. It's also more challenging to keep things up to date if we match with prod, since we may have to change prod (that's the whole reason we make it configurable).

ssciolla: A lot of our defaults in the project cater to local development, so I think this seems in line with those.

pushyamig: ok. I really don't have a strong opinion here, but I just felt if there is problem we could reproduce earlier locally. But that might not always be true with memory/CPU stuff since we cannot map exact Prod setup. 

I am ok to keep the configuration as is.

config/.env.sample
pushyamig: interesting!

pushyamig (approved):

PR #379, opened 2022-04-05
Handle too many matching course results (#378)

### LINKED ISSUE #378: Bail out on search queries with too many results, issue error
In the admin version of "Merge Sections", some short or imprecise course name queries can generate many many results. While the application may be able to handle these queries, the user may have to wait an excessive amount of time and ultimately have too many results to parse through. We could alternatively bail out of certain queries if too many courses match and issue an error to the user. The maximum number of courses could be configurable.

ssciolla:
The PR aims to resolve #378. It also fixes a logging miscalculation introduced in #375 and adds an `isInteger` value check to the consumption of a few configuration variables, including the new `CANVAS_MAX_SEARCH_COURSES`.

### DIFF
--- a/ccm_web/client/src/utils/handleErrors.ts
+++ b/ccm_web/client/src/utils/handleErrors.ts
@@ -18,15 +18,14 @@ class BadRequestError extends Error {
   }
 }
 
-const authzErrorMessage = (
-  'You are not authorized to perform that action. ' +
-  'Try re-launching the application, or contact support.'
-)
+const authzError = 'You are not authorized to perform that action.'
+const authzRecommendation = 'Try re-launching the application, or contact support.'
+const authzFullErrorMessage = `${authzError} ${authzRecommendation}`
 
 class UnauthorizedError extends Error {
   public name = 'UnauthorizedError'
   constructor () {
-    super(authzErrorMessage)
+    super(authzFullErrorMessage)
   }
 }
 
@@ -34,7 +33,7 @@ class ForbiddenError extends Error {
   public name = 'ForbiddenError'
 
   constructor (message?: string) {
-    super(message ?? authzErrorMessage)
+    super(`${message ?? authzError} ${authzRecommendation}`)
   }
 }
 
--- a/ccm_web/client/src/utils/handleErrors.ts
+++ b/ccm_web/client/src/utils/handleErrors.ts
@@ -14,7 +14,7 @@ Custom Error types
 class BadRequestError extends Error {
   public name = 'BadRequestError'
   constructor (message: string | undefined) {
-    super('Your request was not valid' + (message !== undefined ? `: ${message}` : '.'))
+    super('Your request was not valid.' + (message !== undefined ? ` ${message}` : ''))
   }
 }
 
--- a/ccm_web/client/src/utils/handleErrors.ts
+++ b/ccm_web/client/src/utils/handleErrors.ts
@@ -30,10 +30,13 @@ class UnauthorizedError extends Error {
   }
 }
 
-class ForbiddenError extends Error {
+export class ForbiddenError extends Error {
   public name = 'ForbiddenError'
-  constructor () {
-    super(authzErrorMessage)
+  warning: boolean
+
+  constructor (message?: string, warning?: boolean) {
+    super(message ?? authzErrorMessage)
+    this.warning = Boolean(warning)
   }
 }
 
@@ -143,7 +146,9 @@ const handleErrors = async (resp: Response): Promise<void> => {
       if (errorBody.redirect === true) redirect('/')
       throw new UnauthorizedError()
     case 403:
-      throw new ForbiddenError()
+      throw new ForbiddenError(
+        errorBody.warning === true ? apiErrorMessage : undefined, errorBody.warning
+      )
     case 404:
       throw new NotFoundError()
     default:
--- a/ccm_web/client/src/utils/handleErrors.ts
+++ b/ccm_web/client/src/utils/handleErrors.ts
@@ -18,14 +18,15 @@ class BadRequestError extends Error {
   }
 }
 
-const authzError = 'You are not authorized to perform that action.'
-const authzRecommendation = 'Try re-launching the application, or contact support.'
-const authzFullErrorMessage = `${authzError} ${authzRecommendation}`
+const defaultAuthzErrorMessage = (
+  'You are not authorized to perform that action. ' +
+  'Try re-launching the application, or contact support.'
+)
 
 class UnauthorizedError extends Error {
   public name = 'UnauthorizedError'
   constructor () {
-    super(authzFullErrorMessage)
+    super(defaultAuthzErrorMessage)
   }
 }
 
@@ -33,7 +34,7 @@ class ForbiddenError extends Error {
   public name = 'ForbiddenError'
 
   constructor (message?: string) {
-    super(`${message ?? authzError} ${authzRecommendation}`)
+    super(message ?? defaultAuthzErrorMessage)
   }
 }
 
--- a/ccm_web/client/src/utils/handleErrors.ts
+++ b/ccm_web/client/src/utils/handleErrors.ts
@@ -30,13 +30,11 @@ class UnauthorizedError extends Error {
   }
 }
 
-export class ForbiddenError extends Error {
+class ForbiddenError extends Error {
   public name = 'ForbiddenError'
-  warning: boolean
 
-  constructor (message?: string, warning?: boolean) {
+  constructor (message?: string) {
     super(message ?? authzErrorMessage)
-    this.warning = Boolean(warning)
   }
 }
 
@@ -146,9 +144,7 @@ const handleErrors = async (resp: Response): Promise<void> => {
       if (errorBody.redirect === true) redirect('/')
       throw new UnauthorizedError()
     case 403:
-      throw new ForbiddenError(
-        errorBody.warning === true ? apiErrorMessage : undefined, errorBody.warning
-      )
+      throw new ForbiddenError(apiErrorMessage)
     case 404:
       throw new NotFoundError()
     default:
--- a/ccm_web/server/src/api/api.admin.handler.ts
+++ b/ccm_web/server/src/api/api.admin.handler.ts
@@ -105,7 +105,7 @@ export class AdminApiHandler {
     const allCourses: CanvasCourse[] = []
     result.map(cs => allCourses.push(...cs))
     logger.debug(`Number of courses matching search term: ${allCourses.length}`)
-    if (allCourses.length > 50) throw new TooManyResultsError()
+    if (allCourses.length > 1000) throw new TooManyResultsError()
 
     // Get sections for those courses
     const coursesWithSectionsApiPromises = createLimitedPromises(
--- a/ccm_web/server/src/api/api.admin.handler.ts
+++ b/ccm_web/server/src/api/api.admin.handler.ts
@@ -72,6 +72,11 @@ export class AdminApiHandler {
     return parentAccounts
   }
 
+  logTooManyCourseResultsWarning (queryParams: AccountCoursesQueryParams): void {
+    const relevantParams = { by_teachers: queryParams.by_teachers, search_term: queryParams.search_term }
+    logger.warn('Query with the following search term(s) returned too many results: ' + JSON.stringify(relevantParams))
+  }
+
   async getAccountCourses (
     accountId: number, queryParams: AccountCoursesQueryParams
   ): Promise<CanvasCourse[] | APIErrorData> {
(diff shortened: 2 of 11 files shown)


ssciolla:
@pushyamig, I'm still doing testing and looking into this more, but I thought I'd get the review process started.

pushyamig:
Have you decide how many threads and mem limit we will be configuring in non-prod/prod? or we are still doing testing around that to decision?

ssciolla:
> Have you decide how many threads and mem limit we will be configuring in non-prod/prod? or we are still doing testing around that decision?

I think we're still doing testing, but I thought we decided a change like this was also probably necessary so people don't sit there waiting forever for large queries.

pushyamig:
I will test this in afternoon

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -75,10 +78,16 @@ export class AdminApiHandler {
     try {
       const endpoint = `accounts/${accountId}/courses`
       logger.debug(`Sending request to Canvas (get all pages) - Endpoint: ${endpoint}; Method: GET`)
-      const courses = await this.requestor.listItems<CanvasCourse>(endpoint, queryParams).toArray()
+      const courses = []
+      const pages = this.requestor.listPages<CanvasCourse[]>(endpoint, queryParams)
+      for await (const courseResponse of pages) {
+        courses.push(...courseResponse.body)
+        if (courses.length > this.maxSearchCourses) throw new TooManyResultsError()
```
pushyamig, line 85: This is nice, regardless if they searched by course name or uniqname the will be thrown and infact for both instructor and admin view.

pushyamig:
@ssciolla this seems to be working. I added this PR to ccm-dev. This look ok to me from what I could test.

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: How is this? https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/379/commits/db632ac5ebbad736a6fcf615d44e9a066f681a7a

I was uncertain how best to do this without duplicating code too much; this is what I arrived at. There may be something better still.

ssciolla: > This is nice, ~regardless if they searched by course name or uniqname the error will be thrown and infact for both instructor and admin view.~
> 
> Looks like the logic only applied to search by course name. I got the facts wrongs.

@pushyamig, I think you are partially correct. We use a different approach for instructor version of this feature, but if the admin user searches for a uniqname, this check still applies. I don't know if any individual teacher has >400 courses in a term, but that may be worth doing some research on.

ccm_web/client/src/utils/handleErrors.ts
```diff
@@ -14,26 +14,26 @@ Custom Error types
...
   }
 }
 
 class ForbiddenError extends Error {
   public name = 'ForbiddenError'
-  constructor () {
-    super(authzErrorMessage)
+
+  constructor (message?: string) {
+    super(`${message ?? authzError} ${authzRecommendation}`)
```
ssciolla, line 36: I thought this would be an improvement, but I'm still scratching my head on it. Basically, because `ForbiddenError` is also used by other parts of the application (e.g. if the CSRF token is missing), we need to make sure it still makes sense. Maybe it's fine to just always pass the message into the error object, and not worry about a recommendation always. Or I could add a recommendation to the CSRF error. Hm...

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -115,7 +136,8 @@ export class AdminApiHandler {
     const coursesWithSectionsResult = await Promise.all(coursesWithSectionsApiPromises)
     const finalResult = makeResponse<CourseWithSections>(coursesWithSectionsResult)
     if (!isAPIErrorData(finalResult)) {
-      logger.debug(`Number of sections returned: ${finalResult.length}`)
+      const sectionNum = finalResult.map(c => c.sections.length).reduce((runSum, a) => runSum + a, 0)
+      logger.debug(`Number of sections returned: ${sectionNum}`)
```
ssciolla, line 140: I discovered what I added before in PR #375 is actually a bug or miscalculation. `finalResult` is a list of courses with nested section data, so if we actually want the section count, we have to do something like the new line 139.

ccm_web/client/src/utils/handleErrors.ts
ssciolla: Okay, this may be the simplest way to fix this now. https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/379/commits/5a7d726994290fdc152c5e867fe7b8c2c478456b Let me know what you think. I can't think of any other `403` cases produced by our backend (or at least that wouldn't be picked up by the `CanvasError` handling first).

ccm_web/server/src/api/api.admin.handler.ts
pushyamig: >I don't know if any individual teacher has >400 courses in a term

I feel chances are very slim, if the logic is working then I don't think we need to do more research into this.

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -69,16 +72,30 @@ export class AdminApiHandler {
     return parentAccounts
   }
 
+  logTooManyCourseResultsWarning (queryParams: AccountCoursesQueryParams): void {
+    const relevantParams = { by_teachers: queryParams.by_teachers, search_term: queryParams.search_term }
+    logger.warn('Query with the following search term(s) returned too many results: ' + JSON.stringify(relevantParams))
```
pushyamig, line 77:
```suggestion
    logger.warn(`Query with the search term(s) ${JSON.stringify(relevantParams)} returned too many results`)
```

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -38,10 +39,12 @@ or account-scoped operations that make use of other handler instances for Canvas
 export class AdminApiHandler {
   requestor: CanvasRequestor
   userLoginId: string
+  maxSearchCourses: number
 
-  constructor (requestor: CanvasRequestor, userLoginId?: string) {
+  constructor (requestor: CanvasRequestor, userLoginId?: string, maxSearchCourses = 400) {
```
pushyamig, line 44: What do we need to set this number here instead of getting it from the config?

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -102,8 +119,12 @@ export class AdminApiHandler {
     const result = makeResponse<CanvasCourse[]>(coursesResponses)
     if (isAPIErrorData(result)) return result
     const allCourses: CanvasCourse[] = []
-    result.map(cs => allCourses.push(...cs))
+    result.forEach(cs => allCourses.push(...cs))
     logger.debug(`Number of courses matching search term: ${allCourses.length}`)
+    if (allCourses.length > this.maxSearchCourses) {
+      this.logTooManyCourseResultsWarning(queryParams)
+      throw new TooManyResultsError()
```
pushyamig, line 126: Will this code ever hit? by looking at the code from L90-L93 above if there are too many courses then you are throwing an exception and it will exit from this admin handler and go to the api service. 

What is the use case this code is necessary? if I am not mistaken the function `getCourseSectionsInTerm` is only called when searching by course name.

ccm_web/client/src/utils/handleErrors.ts
pushyamig: Yeah... 403 could be canvas error, but in our code we are only passing it with too many courses case. I think it is best to show the error message as received from BE. 

In case of CSRF, JWT or Session expiration, I see 401.

pushyamig: CSRF token missing case and Too many courses are 403, you are differentiating using this extra parameter in `error: Forbidden` in Response. I am not getting how are you using it, can you explain?

ssciolla: Basically, if the 403 error comes from our app (not Canvas), then now the error message will shown to the user (previously, we just showed them a default auth message). We still have a default in the client, but for it to be used, we'd have to send a 403 from the backend with no `message` in the body. So in order to make sure the error message was still helpful, I added the "Try re-launching the application..." text to the CSRF exception thrown in the backend. Does that help explain?

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: Was there something wrong with what I had before?

If I change it to what you have there will be an object in the middle of the sentence which could be confusing looking.

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: So we are setting it in the config, and passing it where it is needed. This is a concession to the fact that the `AdminApiHandler` is also used by the `createExternalUser` endpoint, which doesn't need `maxSearchCourses` (or `userLoginId` for that matter). It doesn't seem like we should require that endpoint to specify these if it's not going to use them, but also not setting them in the constructor has implications for `undefined`s in the course search process.

Ideally speaking, I'd probably lean towards pulling out the `createExternalUser` and `getUserInfo` calls into a separate class so can avoid this awkwardness.

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: So, it's possible that the user will be a sub-account admin in two or more accounts and that while each individual account doesn't have more than 400 courses matching, all together they have greater than 400 courses. This is checking whether that is the case for consistency. Make sense?

pushyamig: Ok, got it make sense now.

ccm_web/client/src/utils/handleErrors.ts
pushyamig: ok, I understand your logic. But I think having a 1 or 2 sentences as comments will help people who step into this scenario.

pushyamig:
Why is this still a draft? This is functional as far as I tested and reviewing the code.

ccm_web/server/src/api/api.admin.handler.ts
ssciolla: I documented by thinking here: https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/380
Icebox is appropriate.

ssciolla:
Okay, @pushyamig, I opened it for full review.

config/.env.sample
```diff
@@ -53,6 +53,8 @@ CANVAS_ADMIN_API_TOKEN=
 # CANVAS_NEW_USER_ACCOUNT_ID=1
 # The Canvas custom roles IDs to use when enrolling users to sections (optional)
 # CANVAS_CUSTOM_ROLES={ "Assistant": 34, "Librarian": 21 }
+# The number of matching courses allowed when searching for course sections
+# CANVAS_MAX_SEARCH_COURSES=1000
```
pushyamig, line 57: This could be changed to 400, reflecting the default in `config.ts`

ssciolla: Wups, I forgot to do that.

ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/379/commits/d59bd950214b3827e7e79a68a2e707e014561e99

ccm_web/server/src/api/api.admin.handler.ts
```diff
@@ -102,8 +119,12 @@ export class AdminApiHandler {
     const result = makeResponse<CanvasCourse[]>(coursesResponses)
     if (isAPIErrorData(result)) return result
     const allCourses: CanvasCourse[] = []
-    result.map(cs => allCourses.push(...cs))
+    result.forEach(cs => allCourses.push(...cs))
     logger.debug(`Number of courses matching search term: ${allCourses.length}`)
+    if (allCourses.length > this.maxSearchCourses) {
+      this.logTooManyCoursesErrorDetails(queryParams)
+      throw new TooManyResultsError()
```
pushyamig, line 126: I found a perfect user who belongs to multiple sub-account, but I couldn't trigger this case, mostly from the former call the TooManyResultError is triggered. But sometime you can't trigger all cases I will go by logic here.

ccm_web/server/src/api/api.controller.ts
```diff
@@ -157,7 +158,7 @@ export class APIController {
     return result
   }
 
-  @UseInterceptors(InvalidTokenInterceptor)
+  @UseInterceptors(InvalidTokenInterceptor, new TooManyResultsInterceptor('courses'))
```
pushyamig, line 161: what is the need to add the `new` keyword here? is there any difference?

config/.env.sample
```diff
@@ -53,6 +53,8 @@ CANVAS_ADMIN_API_TOKEN=
 # CANVAS_NEW_USER_ACCOUNT_ID=1
 # The Canvas custom roles IDs to use when enrolling users to sections (optional)
 # CANVAS_CUSTOM_ROLES={ "Assistant": 34, "Librarian": 21 }
+# The number of matching courses allowed when searching for course sections
+# CANVAS_MAX_SEARCH_COURSES=400
```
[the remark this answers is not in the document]
ssciolla: Let's discuss it elsewhere if it's not related to this PR.

ccm_web/server/src/api/api.controller.ts
ssciolla: The `new` just means we're passing an instance instead of a class, which I need to do because I want to make the plural entities string vary from method to method. Sounds like from the documentation that either will work; see https://docs.nestjs.com/interceptors#binding-interceptors

pushyamig (approved):

PR #389, opened 2022-05-06
#388 Upgrading to MySQL8

### LINKED ISSUE #388: Support ARM containers for launching mysql 5
Added the platform option in `docker-compose.yml`, since the current MySQL image does not support ARM containers (Apple M1 is an ARM-based system).

References:
https://docs.docker.com/desktop/mac/apple-silicon/#known-issues
https://devcoops.com/no-matching-manifest-for-linux-arm64-v8/

pushyamig:
fixes #388 

Testing process:
1. Able to launch the Application both in the Development and Prod mode
2. Able to connect to Dbeaver with MySQL driver.
Screen Shot 2022-07-01 at 11 51 57 AM

### DIFF
--- a/docker-compose-prod.yml
+++ b/docker-compose-prod.yml
@@ -32,5 +32,5 @@ services:
     ports:
       - 5400:3306
     volumes:
-      - ./.mysql_data_prod/mysql:/var/lib/mysql
+      - ./.mysql_data_prod:/var/lib/mysql
     container_name: ccm_db_prod
--- a/docker-compose-prod.yml
+++ b/docker-compose-prod.yml
@@ -33,5 +33,4 @@ services:
       - 5400:3306
     volumes:
       - ./.mysql_data_prod/mysql:/var/lib/mysql
-      - ./mysql:/docker-entrypoint-initdb.d:ro
     container_name: ccm_db_prod
--- a/docker-compose-prod.yml
+++ b/docker-compose-prod.yml
@@ -1,4 +1,4 @@
-version: '3.8'
+version: '3.9'
 
 services:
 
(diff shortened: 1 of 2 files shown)


pushyamig:
@zqian This brings a question that we are using older version of mysql v5 and I don't know in future if they will update these older versions with newer technology changes. 

Should be consider upgrading to latest version of mysql?

jonespm:
I recommended on MyLA to try mysql:8 to see if that works. It looks like the plan is we'll be upgrading our database to that this summer so should start testing on that locally.

pushyamig:
Yeah, I saw the discussion on slack. I am going to wait until how the migration goes and update the PR as needed.

docker-compose-prod.yml
```diff
@@ -20,18 +20,18 @@ services:
       - database
 
   database:
-    image: mysql:5.7
-    command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
+    image: mysql:8-oracle
```
pushyamig, line 23: I took this image as per CAE adoption to it. If prefer something else, please suggest.

docker-compose-prod.yml
```diff
@@ -20,18 +20,18 @@ services:
...
       - MYSQL_DATABASE=ccm
       - MYSQL_USER=admin
       - MYSQL_PASSWORD=admin
-    entrypoint: ['docker-entrypoint.sh', '--default-authentication-plugin=mysql_native_password']
     ports:
       - 5400:3306
     volumes:
-      - ./.mysql_data_prod:/var/lib/mysql:delegated
+      - ./.mysql_data_prod/mysql:/var/lib/mysql
+      - ./mysql:/docker-entrypoint-initdb.d:ro
```
ssciolla, line 36: This line shouldn't be necessary for CCM -- at least not yet. This is to modify the permissions for the user so they can create new databases, which is necessary for Django applications with tests (e.g. Placement Exams or Canvas App Explorer).

jonespm: Right, this is to run init scripts in MySQL which this project doesn't have yet, but might need in the future.

pushyamig: I saw a warning about change the password to sha2 something, without the line 36 things ran fine, but To remove that warning I kept it.

pushyamig: I removed it now and I don't see the warning that I saw. Strange!

docker-compose-prod.yml
```diff
@@ -20,18 +20,17 @@ services:
...
+      - MYSQL_TCP_PORT=3306
       - MYSQL_DATABASE=ccm
       - MYSQL_USER=admin
       - MYSQL_PASSWORD=admin
-    entrypoint: ['docker-entrypoint.sh', '--default-authentication-plugin=mysql_native_password']
     ports:
       - 5400:3306
     volumes:
-      - ./.mysql_data_prod:/var/lib/mysql:delegated
+      - ./.mysql_data_prod/mysql:/var/lib/mysql
```
jonespm, line 35: This isn't a huge deal but changing the mount point of `./.mysql_data_prod` to `./.mysql_data_prod/mysql` will remove the existing database. Though you might have had to remove it anyway to upgrade.

jonespm (approved):

docker-compose-prod.yml
pushyamig: Ok, I changed it. just to keep it as before.

pushyamig:
@ssciolla I request your approval, just to make sure that it run on non-Apple chip machine. But this criteria might not be accurate with upgrade to mysql8. But I want to make sure it runs for you locally without any problem.

ssciolla:
> @ssciolla I request your approval, just to make sure that it run on non-Apple chip machine. But this criteria might not be accurate with upgrade to mysql8. But I want to make sure it runs for you locally without any problem.

@pushyamig, testing now

ssciolla (approved):
LGTM

PR #390, opened 2022-05-06
fix an issue #387(Add Non-UM User: include found user's name on the screen)

### LINKED ISSUE #387: Add Non-UM User: include found user's name on the screen
When adding a non-UM user with the one at a time option and the user already exists, present that user's first and last name on the scree along with the "user found" message. This will help the tool user to make sure the external email address they entered was the right one and aligns with how the old CCM tool worked.

show981111:
Resolves #387 

Added the interface, CanvasUserCondensed, which contains name and loginId 
Returns this interface at userExists functions.

### DIFF
--- a/ccm_web/client/src/api.ts
+++ b/ccm_web/client/src/api.ts
@@ -1,6 +1,7 @@
 import Cookies from 'js-cookie'
 import {
   CanvasCourseBase, CanvasCourseSection, CanvasCourseSectionBase, CanvasEnrollment,
+  CanvasUserCondensed,
   CourseWithSections
 } from './models/canvas'
 import { ExternalUserSuccess } from './models/externalUser'
@@ -156,9 +157,10 @@ export const unmergeSections = async (sectionsToUnmerge: CanvasCourseSection[]):
   return await resp.json()
 }
 
-export const checkIfUserExists = async (loginId: string): Promise<boolean> => {
+export const checkIfUserExists = async (loginId: string): Promise<CanvasUserCondensed | false> => {
   const request = getGet()
   const resp = await fetch(`/api/admin/user/${loginId}`, request)
+  console.log(resp)
   try {
     await handleErrors(resp)
   } catch (error: unknown) {
@@ -172,7 +174,7 @@ export const checkIfUserExists = async (loginId: string): Promise<boolean> => {
       throw error
     }
   }
-  return true
+  return resp.json()
 }
 
 interface ExternalUser {
--- a/ccm_web/client/src/api.ts
+++ b/ccm_web/client/src/api.ts
@@ -157,10 +157,9 @@ export const unmergeSections = async (sectionsToUnmerge: CanvasCourseSection[]):
   return await resp.json()
 }
 
-export const checkIfUserExists = async (loginId: string): Promise<CanvasUserCondensed | false> => {
+export const getUserInfo = async (loginId: string): Promise<CanvasUserCondensed | null> => {
   const request = getGet()
   const resp = await fetch(`/api/admin/user/${loginId}`, request)
-  console.log(resp)
   try {
     await handleErrors(resp)
   } catch (error: unknown) {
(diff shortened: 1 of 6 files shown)


ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -1,5 +1,6 @@
 import { CanvasOAuthReturnQueryDto } from './dtos/canvas.oauth.query.dto'
 import { hasKeys } from '../typeUtils'
+import { string } from 'yup'
```
ssciolla, line 3: Remove?

ccm_web/client/src/api.ts
```diff
@@ -156,9 +157,10 @@ export const unmergeSections = async (sectionsToUnmerge: CanvasCourseSection[]):
   return await resp.json()
 }
 
-export const checkIfUserExists = async (loginId: string): Promise<boolean> => {
+export const checkIfUserExists = async (loginId: string): Promise<CanvasUserCondensed | false> => {
   const request = getGet()
   const resp = await fetch(`/api/admin/user/${loginId}`, request)
+  console.log(resp)
```
pushyamig, line 163: Remove?

pushyamig:
This change seems to be straight forward. @show981111 thanks for making this change. @ssciolla will be the approver.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -223,6 +223,11 @@ export interface CanvasUniqueIdErrorsBody extends CanvasErrorsBody {
   }
 }
 
+export interface CanvasUserCondensed {
+  loginId: string
+  name: string
+}
```
ssciolla, line 229: Couple things here:
1) Let's move this so it's next to the other `CanvasUser` interface.
2) This interface could be based on `CanvasUser`, or vice versa. I guess I see this as a limited subset of the existing object, so we could use [`Pick`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) here, and not redefine `login_id` as `loginId`.
```
type CondensedCanvasUser = Pick<CanvasUser, 'login_id' | 'name'>
```

ssciolla (commented):
Looking good, @show981111. I added a couple comments and suggestions that I'd like to see addressed. Let me know if you have any questions. I'm going to run it locally now and test.

ccm_web/client/src/components/UserEnrollmentForm.tsx
```diff
@@ -178,7 +178,7 @@ export default function UserEnrollmentForm (props: UserEnrollmentFormProps): JSX
         Enter the user&apos;s non-UM email address, and click &quot;Search&quot; to see if they are in Canvas. (Required)
       </Typography>
       {
-        showIncompleteAlerts && userExists === undefined &&
+        showIncompleteAlerts && (userInfo === undefined) &&
```
ssciolla, line 181: Just FYI, parentheses aren't really necessary here, but not a big deal, they can help readability.

ccm_web/client/src/models/canvas.ts
```diff
@@ -140,3 +140,8 @@ export const injectCourseName = (sections: CanvasCourseSection[], courseName: st
 export function sortSections<T extends CanvasCourseSection> (sections: T[]): T[] {
   return sections.sort((a, b) => { return a.name.localeCompare(b.name, undefined, { numeric: true }) })
 }
+
+export interface CanvasUserCondensed {
+  loginId: string
```
ssciolla, line 145: I think this still needs to be updated to be `login_id`? I guess it didn't come up as an issue cause you aren't using the property, but we should probably fix it.

ccm_web/client/src/components/UserEnrollmentForm.tsx
ssciolla: Fine to leave this as is.

ccm_web/server/src/canvas/canvas.interfaces.ts
```diff
@@ -223,6 +223,8 @@ export interface CanvasUniqueIdErrorsBody extends CanvasErrorsBody {
   }
 }
 
+export type CanvasUserCondensed = Pick<CanvasUser, 'login_id' | 'name'>
```
ssciolla, line 226: Please move this so it is next to the other `CanvasUser` type/interface.

Issue #395, opened 2022-07-05
Do Frontend dependency update


ssciolla:
At some point, we'll want to tackle a couple major version upgrades or other library maintenance tasks, besides Material UI #220. They include:

- React 18
-`react-router-dom` 6
-`eslint` version 8 and parsers (blocked, but probably not forever, by pending updates for `eslint-config-standard-with-typescript`)
~~ - Upgrade Node to version 18 ~~
~~ - NestJS, other backend dependencies ~~
~~ - Others? ~~

This can be a parent or "epic" issue, but I'm going to create separate issues to address the backend, frontend, and eslint changes.

### 2024 Frontend Dep Upgrades
@pushyamig and @jaydonkrooss update: we created a [separate branch](https://github.com/tl-its-umich-edu/canvas-course-manager-next/tree/2024-03-01-dep-update) to test concurrent dependency upgrades. Before testing and merging to main, here's the in-progress changes for FE side of things:

- [x] React 18, React-Dom 18, Material UI v5, notistack 3.0.1 
- [x] incremental updates to above packages
- [x] #418
  - [x] Pagination is broken
  - [x] Error window is broken (Try again window)
  - [x] Error icon is red (Review your CSV file) color is changed to blue
  - [x] Alignment issue b/w small and big window inside features workflow
  - [x] Progress bar is close to the section selection window
  - [x] Unmerging text/spinner is missing when unmerging feature 
  - [x] Gradebook formatter requirement list is broken
  - [x] In `CustomTable.tsx`, see this warning SelectProps' is deprecated.ts(6385)
TablePagination.d.ts(122, 6): The declaration was marked as deprecated here.
  - [x] Course Name component is misaligned and the text box is bigger than usual
  - [x] Create single Section textbox has alignment issue

- [x] other frontend packages: react-router-dom, Papaparse, winston, Yup, types/cookie-parser

Nice to have / tasks for after production release
- Upgrade testing packages: supertest, jest

jaydonkrooss:
Some frontend features worth observing to make sure that these upgrades are not breaking anything...

notistack v3 touches all UI snackbars (checking each one might be overkill but verify that both errors and confirmations are formatted and animated correctly):
- API errors (course name, merging/unmerging)
- "Changes were saved" confirmations (course name, create section)
- Other frontend errors ("only csv files accepted" on file upload)

[...]

PR #396, opened 2022-07-06
Perform minor dependency update (#394)

### LINKED ISSUE #394: Do minor dependency update
Some dependencies need to be updated.

ssciolla:
The PR aims to resolve #394.

Tasks:
- [x] Bump `npm` libraries to latest minor or patch versions
- [x] Bump `axios` and respond to breaking type changes
- [x] Remove unused frontend testing code and associated libraries (`@testing-library`)
- [x] Run `npm audit fix`
- [x] Document major versions available (#395)


ccm_web/server/src/canvas/canvas.service.ts
```diff
@@ -203,8 +203,8 @@ export class CanvasService {
       data = response.data
     } catch (error) {
       if (axios.isAxiosError(error) && error.response !== undefined) {
-        const { status, data } = error.response
-        if (status === HttpStatus.BAD_REQUEST && typeof data?.error === 'string' && data.error === 'invalid_request') {
+        const { status } = error.response
+        if (status === HttpStatus.BAD_REQUEST) {
```
ssciolla, line 207: I just found a bug related to this. I am wondering if Canvas changed the response data, since the error string here now seems to be `invalid_grant`. Maybe it was always broken, but I am pretty sure we tested this. I'll have to investigate more this afternoon. The change here seems to fix the problem.

ssciolla: See #397.

ccm_web/package.json
```diff
@@ -19,75 +19,74 @@
   "author": "",
```
pushyamig, line 19: I see on high severity and it suggest to fix `node_modules/moment` library. It is not CCM direct dependency but some libraries use it.

pushyamig:
I was able to run this changes locally and tested as per this [test plan](https://docs.google.com/spreadsheets/d/1kFDRcji_RySw_DtF-ZunvKhSZ9rYNMaPoVGJxVbje4c/edit#gid=0) . I don't see any show stoppers from these changes.

ccm_web/server/src/canvas/canvas.service.ts
ssciolla: Yeah, I was thinking we'd merge the other first, since it's a smaller change.

ccm_web/package.json
ssciolla: Should be resolved now.

ccm_web/server/src/canvas/canvas.service.ts
ssciolla: #398 was merged.

pushyamig:
I think we should close the security alert w.r.t to dependency update.

PR #409, opened 2024-01-23
402: properly cast string course_id to number for global variable

jaydonkrooss:
Consume course_id from canvas Id token as a string, then cast as a number for the globals API endpoint response

### DIFF
--- a/ccm_web/server/src/lti/lti.service.ts
+++ b/ccm_web/server/src/lti/lti.service.ts
@@ -61,7 +61,7 @@ export class LTIService implements BeforeApplicationShutdown {
       const customLTIVariables = token.platformContext.custom
       if (customLTIVariables.login_id === undefined || customLTIVariables.course_id === undefined ||
         customLTIVariables.roles === undefined || customLTIVariables.is_root_account_admin === undefined) {
-        return createLaunchErrorResponse(res, 'please check the LTI configuration in Canvas.')
+        return createLaunchErrorResponse(res, 'One or more required custom LTI variables were not defined. Please check the LTI configuration in Canvas.')
       }
       const loginId = customLTIVariables.login_id as string
       const courseId = customLTIVariables.course_id as string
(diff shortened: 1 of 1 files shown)


pushyamig:
@jaydonkrooss I will review this today or tomorrow.

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -94,16 +94,26 @@ export class LTIService implements BeforeApplicationShutdown {
...
+        req.session.data = sessionData
+      } catch (e) {
+        const logMessageEnding = e instanceof Error
+          ? `error ${String(e.name)} due to ${String(e.message)}`
+          : String(e)
+        logger.error(`Failed to build session data with course ID ${courseId} and roles ${roles}: ${logMessageEnding}`)
+        return createLaunchErrorResponse(res)
       }
-      req.session.data = sessionData
+
```
pushyamig, line 116: I see linting error saying `Trailing spaces not allowed.` in 2 place.

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -94,16 +94,26 @@ export class LTIService implements BeforeApplicationShutdown {
...
+          course: course,
+          isRootAdmin: isRootAdmin
+        }
+        req.session.data = sessionData
+      } catch (e) {
+        const logMessageEnding = e instanceof Error
+          ? `error ${String(e.name)} due to ${String(e.message)}`
+          : String(e)
+        logger.error(`Failed to build session data with course ID ${courseId} and roles ${roles}: ${logMessageEnding}`)
+        return createLaunchErrorResponse(res)
```
pushyamig, line 114: Why did you feel the need to adding a try-catch?

ccm_web/server/src/lti/lti.service.ts
```diff
@@ -94,16 +94,26 @@ export class LTIService implements BeforeApplicationShutdown {
...
-        roles: (roles.length > 0) ? roles.split(',') : [] // role won't be empty but adding a validation for safety
-      }
-      const sessionData = {
-        course: course,
-        isRootAdmin: isRootAdmin
+    
+      try {
+        // More data will be added to the session here later
+        const course = {
+          id: Number(courseId),
```
pushyamig, line 101: With this change I see the course ID is recognized as Number

pushyamig:
I don't see a need to add additional check if checking courseID has any letter since Number('AS122) could return NaN. So this change looks good. 

couple of linting errors, if they are fixed this PR is good to go

ccm_web/server/src/lti/lti.service.ts
jaydonkrooss: I was following a previous example of us handling errors from course ID changing to integer that I think you linked before: https://github.com/tl-its-umich-edu/canvas-app-explorer/commit/e6fc3ac1167f86e7574ded62304fdb406cd1aee3

Since we're reformatting roles and courseId for the sessionData , I felt it made sense to have a descriptive log if the formatting fails

pushyamig: ok

pushyamig:
Looks Good

pushyamig (approved):

PR #429, opened 2024-04-10
434: Fix loading spinner within enrollment section accordion

jaydonkrooss:
Nests the loading backdrop inside the section ids accordion, instead of the outer component

### DIFF
--- a/ccm_web/client/src/components/MultipleSectionEnrollmentWorkflow.tsx
+++ b/ccm_web/client/src/components/MultipleSectionEnrollmentWorkflow.tsx
@@ -270,14 +270,17 @@ export default function MultipleSectionEnrollmentWorkflow (props: MultipleSectio
             <Link href={sectionDataToDownload} download='course_section_ids.csv'>
               Download a CSV with the Canvas Course Section IDs data
             </Link>
-            <Accordion title='Course Section Canvas IDs' id='section-ids'>{sectionIdsTable}</Accordion>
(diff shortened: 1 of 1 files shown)


pushyamig:
I will review it today

pushyamig:
@jaydonkrooss the issue description is about **Add Non U-M Users Multiple section** feature. But the fix seems changing the Accordion control spinner background for **Add U-M user to multiple section**. 

your fix seems to be having the grey background to just accordion. I think looks better but check with @ktowneUM if she want you keep this way or as in prod right now. Based on that create a separate issue and tag this PR with that issue

jaydonkrooss:
> @jaydonkrooss the issue description is about **Add Non U-M Users Multiple section** option. But the fix seems changing the Accordion control spinner background for **Add U-M user to multiple section**. option
> 
> your fix seems to be having the grey background to just accordion. I think looks better but check with @ktowneUM if she want you keep this way or as in prod right now. Based on that create a separate issue and tag this PR with that issue

I see, I must have been mistaken about the original documentation of the UI bug. I've created a new issue and will try to address the original issue in a different PR

pushyamig:
@jaydonkrooss Try to link this PR to issue#434, otherwise the Commit history will pointing the PR to wrong issue. I thing you can use `git ammend` to change the commit message. or you could create a new PR. It's up to you.

jaydonkrooss:
@pushyamig I think I'll just create a new PR to rename the branch as well. would be simpler for me I think

PR #433, opened 2024-04-12
I330 alt cjs esm upgrade

pushyamig:


jaydonkrooss:
So just so I understand, you're saying we can't upgrade p-limit because it isn't interoperable with CJS and ts-node isn't properly compiling it, in development only. So we're just waiting for ts-node to support the "sub-path import" feature before p-limit upgrade?

pushyamig:
>So just so I understand, you're saying we can't upgrade p-limit because it isn't interoperable with CJS and ts-node isn't properly compiling it, in development only. So we're just waiting for ts-node to support the "sub-path import" feature before p-limit upgrade?

Yes, Right now that's my understanding. I feel we don't need to rush p-limit to upgrade v5 we will see how that PR with ts-node flows, otherwise we might have to come up with some hacky way to make it work which I felt can hold on for time being. Currently it is upgraded from V3 -> v4 which is still an ESM module.

I felt a bit waxed/fatigued with this particular issue so I we can come back to it later

pushyamig:
>p-limt v5 build

This is the Build/run that was successful when running Prod mode: https://github.com/pushyamig/canvas-course-manager-next/actions/runs/8651060404/job/23720922489

But Locally the Build/or Run will have errors with V5. SO rolled back https://github.com/pushyamig/canvas-course-manager-next/actions/runs/8652300727

jaydonkrooss:
Went through the test plan, locally testing features and migrations; all looks like it behaves as expected to me. Code looks good too.

jaydonkrooss (approved):

PR #439, opened 2024-04-18
427: fix section selection styling by refactoring ListItem Button

### LINKED ISSUE #427: Selecting a section doesn't have a grey background
Some where during the Upgrade of MUI v5 this is broken
**Features affected: Merge Section, Add-U-M user, Add Non U-M User, Format Third‑Party Gradebook.** (All features where section selection is allowed)

**Expected:** 
Screenshot 2024-04-03 at 9 18 36 AM
**Actual:**
Screenshot 2024-04-03 at 9 18 29 AM

jaydonkrooss:
Fixes #427 

Fix section selection styling by refactoring ListItem button. ~~In trying to fix styling bugs on Merge Section page, I moved unmerge button to an alternate position.~~ Merge section styling is updated to retain old layout

### DIFF
--- a/ccm_web/client/src/components/SectionSelectorWidget.tsx
+++ b/ccm_web/client/src/components/SectionSelectorWidget.tsx
@@ -54,8 +54,7 @@ const classes = {
   sectionSelectionContainer: `${PREFIX}-sectionSelectionContainer`,
   backdrop: `${PREFIX}-backdrop`,
   highlighted: `${PREFIX}-highlighted`,
-  button: `${PREFIX}-button`,
-  unmergeWrapper: `${PREFIX}-unmergeWrapper`
+  button: `${PREFIX}-button`
 }
 
 // TODO jss-to-styled codemod: The Fragment root was replaced by div. Change the tag if needed.
@@ -178,12 +177,6 @@ const Root = styled('div')((
   [`& .${classes.button}`]: {
     margin: theme.spacing(1),
     marginLeft: '24px',
-  },
-
-  [`& .${classes.unmergeWrapper}`]: {
-    position: 'absolute',
-    right: '0',
-    padding: 'inherit'
   }
 }))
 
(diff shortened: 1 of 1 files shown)


pushyamig:
@jaydonkrooss what was the need to move the unmerge button toward right? CCM views are designed based on the UX/UI principles. Personally I prefer to stay like, but I want to understand your thinking here

jaydonkrooss:
@pushyamig The old unmerge button was automatically placed below the course name because it was a part of the secondaryAction prop of ListItem. ListItemButton has no SecondaryAction prop, so I tried putting unmerge button simply as a child of ListItemButton, but the styling was not exactly the same (number of students text was causing problems with the spacing). This is a result of me messing around to make fit the button with position:absolute but now that I'm looking at it, I think I could maybe try to replicate the same result, if we're concerned about making style changes

jaydonkrooss:
@pushyamig ready for local testing & approval. The new version of the unmerge button now imitates the old layout,

pushyamig:
@jaydonkrooss for Merge section feature, some how  grey background when the spinner is rolling the out of view(looks like hidden). This apply to both left and right side of merge section some how this change effected it. Not happening from Feature branch latest

jaydonkrooss:
> @jaydonkrooss for Merge section feature, some how grey background when the spinner is rolling the out of view(looks like hidden). This apply to both left and right side of merge section some how this change effected it. Not happening from Feature branch latest

Good catch! Looks like I made a mistake nesting the backdrop inside the list element. Should be fixed in the latest commit

pushyamig:
In Merge Section after searching and moving the course sections from left to right, there is usually a think blue line indicating your section selection for crosslinting and that is missing

Screenshot 2024-04-24 at 2 12 44 PM

pushyamig:
>Good catch! Looks like I made a mistake nesting the backdrop inside the list element. Should be fixed in the latest commit

The issue seems to be fixed now.

jaydonkrooss:
> In Merge Section after searching and moving the course sections from left to right, there is usually a thick blue line indicating your section selection for crosslinting and that is missing!

I see, sorry I totally missed restoring this with the new component. Should be fixed now

pushyamig (approved):

PR #462, opened 2024-10-23
Backend dependency update

pushyamig:
Fixes some of the issue in the https://github.com/tl-its-umich-edu/canvas-course-manager-next/security/dependabot

`mysql2` is the one with critical severity, we have a direct dependency and `ltijs-sequelize` also uses this pks still using the older with critical severity. I did not want to force update since it might break thinks. So we have no choice to accept the risk until library owner updates it. Since the backend rewrite is coming I felt we could accept the risk

I have done some [mapping](https://docs.google.com/spreadsheets/d/1Z7XAShO_sEATae1dpZK8fVE3JoCUQ6Ld6dgPxCaQqsA/edit?usp=sharing) of the Dependabot alert we have to what to update our end

Not all package could be updated that were in either in high|moderate severity category since the corresponding pkg owners did not update it. Since we have some many packages it make the dependency process a bit tetious.

I have used [ncu package](https://www.npmjs.com/package/npm-check-updates) for checking the package update and vulnerability.

`npm audit report` states that either `no fix available` or `fix available via `npm audit fix --force` so i opted to keep as it is and accept risk at this point

The [Github Action ](https://github.com/pushyamig/canvas-course-manager-next/actions/runs/11485308605/job/31965067985 )build is successful from my local branch that I am opening this PR

Test Plan:
High level integration testing of various component of CCM


docker-compose-prod.yml
```diff
@@ -1,5 +1,3 @@
-version: '3.9'
```
pushyamig, line 1: We don't need to have docker version mentioned as this is removed from docker attributes.

jaydonkrooss:
Looks good to me. I tested a couple of ccm modules that use APIs.

jaydonkrooss (approved):

PR #490, opened 2024-11-08
161 use google analytics onetrust module

### LINKED ISSUE #161: Implement google analytics to track user metrics
A new Google analytics Id need to be created for test/qa/prod, the global API call should be sending the GA ID to UI. 

This library is used by MyLA and ROHO as well https://github.com/react-ga/react-ga. We could use this but shop around if there is something better available as well

jaydonkrooss:
Fixes #161


pushyamig:
this PR has conflict, it needs to be resolved for review.

pushyamig:
I will review this in the afternoon

ccm_web/server/src/main.ts
```diff
@@ -72,11 +73,16 @@ async function bootstrap (): Promise<void> {
...
   const serverConfig = configService.get('server', { infer: true })
+  const privacyURL = configService.get('privacyURL', { infer: true })
 
   app.set('trust proxy', 1)
 
   doAppCoreSetup(app, serverConfig)
 
+  app.use('/privacy', (req: Request, res: Response) => {
+    res.redirect(privacyURL);
+  });
```
pushyamig, line 84: Why is this URL added to ccm? isn't it should part of banner code to open in a new window?

.github/workflows/ccm.yml
```diff
@@ -51,7 +52,7 @@ jobs:
     
   release: 
      # Making sure that release only runs for tag pushes
-    if: startsWith(github.ref, 'refs/tags/') && github.repository_owner == 'tl-its-umich-edu'
+    if: startsWith(github.ref, 'refs/tags/') && (github.repository_owner == 'tl-its-umich-edu' || github.repository_owner == 'jaydonkrooss')
     needs: build # This ensures the build job finishes successfully before starting this job
```
pushyamig, line 56: This code should be removed now since it served it's purpose

pushyamig:
I tested this code in various scenarios and it working from fine from Chrome. I was looking at realtime data from analytics dashboard. 

@jaydonkrooss if I switch the browser to non-chrome like Edge, I don't see events are being tracked in the analytics Dashabord. But I do see Banner appearing. Is this your experience as well? is there any thing that saying,  realtime events can only be shown from Chrome?

pushyamig:
with the behavior where TagHound Browser extension showing duplicate events I don't feel the events in the GA dashboard is captured like that. But it is really head to judge to track single vs duplicate events from GA Dashboard. 

I tried to open MYLA from Prod Version and watched TagHound Extension behaviour, I did not see events being tracked at all from browser extension. So I don't fee taghound is very reliable. 

I think we should investigate further after the CCM release, the events tracking/testing for this GA React library. since I am not sure we really have a better way of testing this process. This is low priority and might need more understanding of GA tracking in Dashboard and when they are published.

ccm_web/server/src/main.ts
jaydonkrooss: The privacy path is a part of the documentation for the Umich Onetrust consent banner, it simply directs to <domain>/privacy/. I had to add this as environment variables to MyLA and ROHQ as well. https://vpcomm.umich.edu/resources/cookie-disclosure/

jaydonkrooss:
> I tested this code in various scenarios and it working from fine from Chrome. I was looking at realtime data from analytics dashboard.
> 
> @jaydonkrooss if I switch the browser to non-chrome like Edge, I don't see events(page_views) are being tracked in the analytics Dashabord. But I do see Banner appearing. Is this your experience as well? is there any thing that saying, realtime events can only be shown from Chrome?

Strange, my testing with Safari & Firefox show event live-tracking. It's possible that the browsers or VPN you were looking at had analytics tracking blocked by default, may need to change browser and VPN settings to make sure you're not blocking ad tracking

.github/workflows/ccm.yml
```diff
@@ -51,7 +51,7 @@ jobs:
...
```
pushyamig, line 55: Can you remove your GH repo?

pushyamig (approved):
Looks Good!

.github/workflows/ccm.yml
jaydonkrooss: changed back now

.github/workflows/ccm.yml
jaydonkrooss: removed

Issue #466, opened 2024-10-31
Separate FE build and wire Django Webpack Loader


pushyamig:
This issue is to separate the FE build using webpack, use Django Webpack Loader. Also support the Webpack bundle creation during development.

This is currently happening just making sure the migration work as well

zqian:
verified the use of webpack_loader

PR #492, opened 2025-01-14
I465 Adding django backend

### LINKED ISSUE #465: Integrate with Django Ecosystem
This issue will be primarily removal of Node backend and add Django support
1. Remove Node stuff
2. Adding Docker to support Python build
3. Using a Backend server - like Guincorn
4. support Typing. Django has support from this https://github.com/typeddjango/django-stubs

pushyamig:
Fixes #465 

This PR is 

1. Adding Django Backend
2. Removing all the NodeJS Backend packages from `package.json`.
3. Doing the Webpack build from only React code
4. Using docker Supervisor feature for running FE and BE
5. README instruction are there. Simply only doing `docker compose build` and `docker compose up`
6. `.env` file in not place in the project root directory instead in the `${HOME}/secrets/ccm/.env`. So create folder `/secrets/ccm`

The CCM homepage will launch, but for now I added dummy data for the API call  when the home page launches. This setup is going to be temporary until be have API integration happens. All the changes can be found in `api.ts`

This PR does not have
1. LTI integration, which is next
2. I haven't setup DockerBuild for running in Prod mode, which I will pause and will look when doing the Openshift setup
3. You might see a warning saying 
`ExperimentalWarning: --experimental-loader may be removed in the future; instead use register()`
That is because CCM migrated to ESM and not commonJS. So in future version of Node `commonJS` support might be removed.
 `node --loader ts-node/esm node_modules/.bin/webpack --config webpack/webpack.dev.ts --watch` 
    4. More On this can be found [here](https://github.com/webpack/webpack-cli/issues/2916). This is something know during  #https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/433
5. I did not remove the NodeJS code w.r.t to BE which is in `server`. I will remove it in future.


requirements.txt
```diff
@@ -0,0 +1,11 @@
+gunicorn==23.0.0
+mysqlclient==2.2.6
+
+#Django
+Django==5.1.4
```
jonespm, line 5: Django 5.1 isn't LTS, 5.2 LTS doesn't come out until April. You _may_ want to hold back to Django 4.2 (currently 4.2.18) until later this year unless this isn't going to be released until after April 2025.

requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+django-mysql==4.15.0
+django-webpack-loader==3.1.1  
+
+debugpy==1.8.8
```
jonespm, line 9: This should be updated to 1.8.11, also might want to check the versions of the other dependencies here?

README.md
```diff
@@ -13,45 +13,12 @@ the [Nest framework](https://docs.nestjs.com/) for the NodeJS server.
...
 
 This repository includes in the `config` directory a `.env.sample` template file for `.env`,
 with the expected keys provided. Comments above each key describe what the value is used for,
 and in some cases, indicate that the key-value pair is optional.
 
-The `config.ts` module in the `ccm_web/server` directory validates this file,
-using fallbacks when available and throwing an error if required values are not present.
-The application will exit if any of the required configuration values are not set properly.
-
-#### Canvas
```
jonespm, line 26: I think much of this section is still relevant. Might want to break it out the parts that won't change into separate .MD files to make it easier to maintain separately like we did for MyLA in `docs`.

docker-compose.yml
```diff
@@ -2,25 +2,21 @@ services:
...
+      dockerfile: Dockerfile
       args:
         - PORT=4000
-      target: base
-    working_dir: /base/
     environment:
       - PORT=4000
-      - NODE_ENV=development
-    env_file: .env
+      - DEBUG=True
```
jonespm, line 11: I found out with ROHQ that you can use this pattern to specify defaults in the docker-compose file

`      - DEBUG=${DEBUG:-True}`

I'd switch to this here for everything in the environment section. 

https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/e329460a143c18fdcaf3dbb24284879a30ebc9c2/docker-compose.yaml#L9

docker-compose.yml
```diff
@@ -2,25 +2,21 @@ services:
 
   web:
     build:
-      context: ./ccm_web/
-      dockerfile: ./Dockerfile
+      context: .
+      dockerfile: Dockerfile
       args:
```
jonespm, line 7: I don't think the args are needed here anymore.

requirements.txt
pushyamig: ok

jonespm (approved):
Approving but left a few comments which can be addressed now or in future issues if you just want to move this along. Thanks!

requirements.txt
pushyamig: I feel Releasing in Apr is very optimistic. I feel this is going to live in summer. Thats why I chose Django 5 version. I can change to 4 version if you like. LMK

requirements.txt
pushyamig: Done

docker-compose.yml
pushyamig: Thanks, Removed

requirements.txt
pushyamig: I will keep 5.1 and will upgrade to 5.2 before release. I will create an issue for that.

pushyamig:
This PR Finishes #466,

jaydonkrooss (approved):

pushyamig:
>Looks good to me, locally ran it with 404 errors on the API, which I assume is expected. I can help with APIs once we move on to that step, just let me know.

Yes, API endpoints are not wired up so those errors are expected. We are progressively building it. This PR is to separate the FE from the Node setup and build using on Webpack and make the Js Dump file `main-*.js` available via Django Setup. Along with that integrate with Django Ecosystem.  

When CCM launch it makes API call, so I have to set up some static JSON response and launching the home page so that It acceptable for this PR.

pushyamig:
@jonespm and @jaydonkrooss for the Quick Review.

PR #500, opened 2025-01-29
fixes #485: add Watchman monitoring url

### LINKED ISSUE #485: Monitoring with Django Watchman
for projects without Watchman, Zabbix only monitors the base URL, checking for a 200 OK response. For applications that have Watchman implemented, the Zabbix web scenario is configured to monitor the Watchman URL. This provides an enhanced monitoring experience, as Watchman performs additional checks, such as database, cache, and storage validation. Additionally, Watchman’s functionality is extensible, offering potential for further custom checks as needed.

https://github.com/mwarkentin/django-watchman

jaydonkrooss:
test plan: check to see if the monitoring endpoint is available using your local ngrok or loophole URL at <URL>/watchman/

The requested resource should be a report and response on the status of database, cache, etc.

### DIFF
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -45,7 +45,8 @@ INSTALLED_APPS = [
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'webpack_loader',
-    "lti_tool"
+    "lti_tool",
+    'watchman',
 ]
 
 MIDDLEWARE = [
@@ -193,6 +194,10 @@ LOGGING = {
         'canvas_oauth': {
             'handlers': ['console'],
             'level': 'WARN'
+        },
+        'watchman': {
+            'handlers': ['console'],
+            'level': 'DEBUG',
         }
     }
 }
@@ -229,3 +234,8 @@ HELP_URL = os.getenv('HELP_URL', 'https://ccm.tl-pages.tl.it.umich.edu')
 CANVAS_INSTANCE_URL = os.getenv('CANVAS_INSTANCE_URL', 'https://canvas.instructure.com')
 
 DEBUGPY_ENABLE = os.getenv('DEBUGPY_ENABLE', False)
+
+# Watchman settings (https://github.com/mwarkentin/django-watchman)
+WATCHMAN_TOKENS = os.getenv('DJANGO_WATCHMAN_TOKENS', None)
+WATCHMAN_TOKEN_NAME = os.getenv('DJANGO_WATCHMAN_TOKEN_NAME', 'ccm-watchman-token')
+WATCHMAN_CHECKS = ('watchman.checks.caches', 'watchman.checks.databases')
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -201,7 +201,7 @@ LOGGING = {
         },
         'watchman': {
             'handlers': ['console'],
-            'level': 'DEBUG',
+            'level': os.getenv('DJANGO_LOG_LEVEL', 'INFO'),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


backend/settings.py
```diff
@@ -193,6 +194,10 @@
         'canvas_oauth': {
             'handlers': ['console'],
             'level': 'WARN'
+        },
+        'watchman': {
+            'handlers': ['console'],
+            'level': 'DEBUG',
```
jonespm, line 200: This feels like it should maybe use `os.getenv('DJANGO_LOG_LEVEL', 'INFO')` like above rather than everything at DEBUG level? Does this cause a lot of log messages when accessing watchman?

pushyamig:
I see this working when called `https://fd8e-68-36-190-34.ngrok-free.app/watchman/` and it is responding with 

```
{
  "databases": [
    {
      "default": {
        "ok": true
      }
    }
  ],
  "caches": [
    {
      "default": {
        "ok": true
      }
    }
  ]
}
```

backend/settings.py
pushyamig: I see these debug logs when the `/watchman/` url is called

```
ccm_web    | [2025-02-13 21:48:46 +0000] [DEBUG] [decorators.py:26] Checking '_check_database' for 'default'
ccm_web    | [2025-02-13 21:48:46 +0000] [DEBUG] [decorators.py:26] Checking '_check_database' for 'default'
ccm_web    | [2025-02-13 21:48:46 +0000] [DEBUG] [decorators.py:26] Checking '_check_cache' for 'default'
ccm_web    | [2025-02-13 21:48:46 +0000] [DEBUG] [decorators.py:26] Checking '_check_cache' for 'default'
```

pushyamig: Not sure why duplicate logs of same check, since these are debug might now be noise in prod

pushyamig:
@jonespm Do you want the `/bare_status/` url, since with [Myla](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/urls.py#L42) it is configured?

backend/settings.py
jonespm: Yeah we'd want it to be able to run at `INFO `or `WARNING `probably in prod. Just seemed fine to match the DJANGO level.

jonespm:
Zabbix is currently using /bare_status for MyLA, so I guess it's worth adding. It's not included in the default urls.

backend/settings.py
pushyamig: I think after the adjusting the Log level to Django log variable I don't see the duplication log messages

config/.env.sample
```diff
@@ -39,6 +39,11 @@ PRIVACY_URL=https://umich.edu/about/privacy/
 # Redis settings
 REDIS_LOCATION=redis://ccm_redis:6379
 
+# (optional) The token for accessing the /status URL; leaving it undefined means the route is unprotected
+# DJANGO_WATCHMAN_TOKENS=
+# (optional) The name of the header or parameter used for passing the above token
+# DJANGO_WATCHMAN_TOKEN_NAME=ccm-watchman-token
+
```
pushyamig, line 46: @jonespm Do you think we will be using these variables in CCM Non Prod deployment? I am ok these as optional parameters. 

I ask since none of our app's seems to use this watchman tokens.

pushyamig (approved):
Nice Work! I will approve it working as expected. 

@jonespm if you don't have additional comment, I request you approve this PR as well.

config/.env.sample
jonespm: I think it's fine to leave these in there but we haven't found a need to protect them in other apps. Since it checks the DB it might be possible for someone to have a denial of service?

jaydonkrooss: Let me know if we ought to set up this variable for production env

Issue #475, opened 2024-10-31
PUT: Course Rename API


pushyamig:
This is to support the Course renaming feature. Making sure the CSRF wiring works

ktowneUM:
Passed

Issue #474, opened 2024-10-31
GET Course Call API


pushyamig:
Needs a structure that maps to current CCM design

ktowneUM:
course name matches

PR #503, opened 2025-02-17
I471 drf integration

### LINKED ISSUE #471: Add Django Rest framework(drf) support
https://www.django-rest-framework.org/

pushyamig:
Fixes #471 #475 #474 

- Complete integration of DRF and wiring the Course #475 and #474.
- Swagger Support - Try launching it from the CCM home page
- Complete integration support for Canvas OAuth Revoked token. So if canvas token is expired, revoked the workflow should now support all use cases
- DRF API tracking via [drf-api-tracking](https://github.com/lingster/drf-api-tracking)
- API integration with UCF canavsapi library
- Cleanup of FE with csrf workflow with Node backend. All the PUT/POST/DELETE call wired up for Django CSRF support. Look at `api.ts`.
- Not fixing this warning `Your models in app(s): 'canvas_oauth', 'rest_framework_tracking' have changes that are not yet reflected in a migration, and so won't be applied.` with this PR
- Unit Test  written for PUT/GET API calls both happy/unhappy path and Revoked token access.
- ------------------------------------
`Test cases to test`: 
1. Home page Course GET/PUT calls are working. Make sure Canvas API key is enabled with then. Check the `canvas_scopes.py` for details
2. Revoke the Canvas OAuth token by going into Canvas User settings, launch CCM it will throw an error. Once user relaunches the CCM the Canavs Oauth Workflow should kick in. 
   1.  We catch the error during a Canvas API call, delete the Canvas token in DB and then throw an error to the user, so user might relaunch the CCM to trigger the Canvas Oauth workflow once they see the error.
   2. So while catching the InvaildTokenError we identify the revoked token state, delete the canvas token of user stored in DB , throw error and user launches the CCM to trigger oAuth process

4. Make sure API GET/PUT are being tracked in the `rest_framework_tracking_apirequestlog` table in DB
5. Another use case to test is delete the Canvas API key and create new one, make sure you have canvas token from before in local DB when you test this use case. This use is mainly triggered during Canvas Test/Beta instance reset with Prod data.

---------------------------------------

Tip: 
1. <React.StrictMode> </React.StrictMode>in the `index.tsx`  duplicate Course GET API call. This side effect is only noticed in Develop mode. This was added as part of React 18 (and duplication won't happen in Prod mode when launching). This commenting will help see the Revoked token error message with GET Call. Alternatively you can also test with PUT call
2. Do a docker build since both FE and BE has dependencies changes
3. Config change to support CSP for DRF and tracking.


pushyamig:
Resolving conflicts from #500

jonespm (commented):
I'm not going to have time to fully review and run this until next week but looking over the code it look good.

backend/ccm/canvas_api/exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+class CanvasHTTPErrorData(TypedDict):
+    status_code: int
+    message: str
+    
+class CanvasHTTPError(Exception):
+    """
+    Custom exception for HTTP errors originating from Canvas API interactions
+    """
+
+    canvas_error_prefix = 'An error occurred while communciating with Canvas. '
```
jonespm, line 18:
```suggestion
    canvas_error_prefix = 'An error occurred while communicating with Canvas. '
```

backend/settings.py
```diff
@@ -251,3 +254,19 @@
...
+
+REST_FRAMEWORK = {
+    'DEFAULT_SCHEMA_CLASS': 'drf_spectacular.openapi.AutoSchema',
+    'DEFAULT_AUTHENTICATION_CLASSES': [
+        'rest_framework.authentication.SessionAuthentication',
+    ],
+}
+
+SPECTACULAR_SETTINGS = {
+    'TITLE': 'Canavs Course Manager API',
```
jonespm, line 266:
```suggestion
    'TITLE': 'Canvas Course Manager API',
```

jonespm:
Did we have an issue opened for "Not fixing this warning Your models in app(s): 'canvas_oauth', 'rest_framework_tracking' have changes that are not yet reflected in a migration, and so won't be applied. with this PR"? 

You could assign to me and I could apply a similar as used in Instructor Tools

backend/ccm/canvas_api/canvas_credential_manager.py
```diff
@@ -0,0 +1,52 @@
+import logging
+from http import HTTPStatus
```
jonespm, line 2: Perhaps it's minor but we might want to look into having some code quality checks installed in this app as it gets developed. Running [isort ](https://pycqa.github.io/isort/) on this would clean these imports up a bit and make it more consistent. 

I've liked the [requirements of  Canvas API](https://github.com/ucfopen/canvasapi/blob/develop/scripts/run_tests.sh) and with you adding tests we could run coverage, black, isort, flake8 and _maybe_ some other custom scripts. Easier to apply this now than later. We could discuss this and create this as a new issue.

pushyamig: Sure, I like the idea, we can implement it. Please create the issue and we will implement it.

backend/ccm/canvas_api/canvas_credential_manager.py
```diff
@@ -0,0 +1,52 @@
...
+    BadRequest, CanvasException, Conflict, Forbidden, InvalidAccessToken, RateLimitExceeded,
+    ResourceDoesNotExist, Unauthorized, UnprocessableEntity
+)
+from .exceptions import CanvasHTTPError 
+
+logger = logging.getLogger(__name__)
+
+class CanvasCredentialManager:
+
+  EXCEPTION_STATUS_MAP = {
```
jonespm, line 21: I think this would be more organized for this and `canvas_api_exception` method to be in `exceptions.py`.

I'm kind of even confused about what this does. Maybe you can describe why we need this custom MAP and exception handling or add some comments around this?

backend/ccm/canvas_api/course_api_handler.py
```diff
@@ -0,0 +1,93 @@
...
+            
+            # Format the course object to return specific course info
+            formatted_course = {
+                "id": course.id,
+                "name": course.name,
+                "enrollment_term_id": course.enrollment_term_id
+            }
+            
+            return Response(formatted_course, status=HTTPStatus.OK)
+        except CanvasException as e:
```
jonespm: I feel like these 2 errors are redundant between these methods and probably will be repeated in multiple places. We probably should have a centralized method like `handle_canvas_error`? It seems like you have one like that in canvas_credential_manager, just maybe need to move that to some other file? I'm not sure but I feel like there will just be a lot of code duplication here as this keeps being developed.

backend/ccm/canvas_api/course_api_handler.py
```diff
@@ -0,0 +1,93 @@
...
+        serializer = CourseSerializer(data=request.data)
+        if serializer.is_valid():
+            update_data = serializer.validated_data
+            try:
+                canvas_api: Canvas = CANVAS_CREDENTIALS.get_canvasapi_instance(request)
+                # Get the course instance
+                course: Course = canvas_api.get_course(course_id)
+                # Call the update method on the course instance
+                put_course_res: str= course.update(course={'name': update_data.get("newName"), 'course_code': update_data.get("newName")})
+                formatted_course = {'id': course.id, 'name': put_course_res, 'enrollment_term_id': course.enrollment_term_id }
```
jonespm: Is this just returning a value from the PUT call? I don't see this value even used in the frontend but maybe I'm just missing where it's used.

PUT doesn't actually need to return values, it can also just return an `HTTPStatus.NO_CONTENT` to say that the request was successful.

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: The canvasapi return only success/error object, in CCM response to API calls from FE we return a status code and based on the status code different action are taken in FE. 
So we are keeping the internal map, I don't think the response Object from canvasapi return status code. Hence the map

jaydonkrooss:
sorry for the delay, got a little stuck trying to switch over to the python version locally.
Course name API is normally working, and all the additional functionality runs well, (checked swagger, revoking canvas Oauth, & DRF tracking). 

The code itself looks good to me as well, aside from Matt's suggestions to better organize error handling.

jonespm (approved):
Added a few comments that could be left out, fixed or taken up in a future PR. Looks good otherwise and happy to move forward with this in.

jaydonkrooss (approved):

backend/ccm/canvas_api/course_api_handler.py
pushyamig: Yes, what your observation is correct. But I am just following a pattern returning the FE with a response object that needed. More on this here https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/main/ccm_web/server/src/api/api.course.handler.ts#L59

I am trying to minimize the changes needed for FE. So Mapping the Response Object that BE creates based on what FE is expecting. 

This way we don't have to do a huge refactor on FE. There are some cascading effect don't know where these response used in FE beyond the PUT call for display

pushyamig:
I want to look into the Error handling aspect for this PR, i feel it important to spend some time here.

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: I added comment on this.

backend/ccm/canvas_api/course_api_handler.py
pushyamig: I have refactored the code with a simplified error approach 
https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/503/commits/3baba4701542b2a02a769721c76ba3766d19ef89

pushyamig:
I have changed the code that was raised due to the duplication for error handling code, I have adopted the test case with this new approach. So if there is more feedback, I will implement in future issue.

pushyamig:
I see an issue when running `docker compose build` . This issue is also happening in `develop` branch. I think `setuptools` is enforcing `license-file` in 'metadata' (setup.cfg), please use the underscore name `license_file` instead. 
Checking out [django-canvas-oauth ](https://github.com/Harvard-University-iCommons/django-canvas-oauth/blob/master/setup.cfg#L2) might needs an update. But I am trying other ways to hack a fix

### If you clean your docker images it will trigger this error. I feel this issue is a blocker and needs a fix. This issue is nothing to do with my changes in the PR`
```
2.639 [notice] To update, run: pip install --upgrade pip
2.908 Collecting django-canvas-oauth (from -r requirements.txt (line 21))
2.908   Cloning https://github.com/Harvard-University-iCommons/django-canvas-oauth.git (to revision v1.1.1) to /tmp/pip-install-zzuee6it/django-canvas-oauth_aa33978249d947a8845b23314ad3d7d6
2.910   Running command git clone --filter=blob:none --quiet https://github.com/Harvard-University-iCommons/django-canvas-oauth.git /tmp/pip-install-zzuee6it/django-canvas-oauth_aa33978249d947a8845b23314ad3d7d6
3.628   Resolved https://github.com/Harvard-University-iCommons/django-canvas-oauth.git to commit 299997426b3f9ca830e7d3085f70aba6fc92504c
3.631   Installing build dependencies: started
5.750   Installing build dependencies: finished with status 'done'
5.750   Getting requirements to build wheel: started
5.858   Getting requirements to build wheel: finished with status 'error'
5.861   error: subprocess-exited-with-error
setuptools.errors.InvalidConfigError: Invalid dash-separated key 'license-file' in 'metadata' (setup.cfg), please use the underscore name 'license_file' instead.
5.861       [end of output]
```

pushyamig:
This is the issue where this error case is triggering: https://github.com/pypa/setuptools/pull/4837/files

pushyamig:
I am enforcing older version of `setuptools` and that seems to get the build working

```
# Ensure pip, setuptools, and wheel are installed before installing dependencies
RUN python -m ensurepip && \
    pip install --no-cache-dir --upgrade pip wheel && \
    pip install --no-cache-dir setuptools==67.0.0
```

pushyamig:
The error with `setuptools` is not happening any more, I removed the fix as mentioned [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/503#issuecomment-2749354149). I don't know why I was seeing this issue. 

@jaydonkrooss also ran this branch locally and he did not report the error. 

Either way this PR is not the reason why temporarily i was seeing the issue. So I am merging this PR now. Leave comment if anybody have any thoughts

PR #506, opened 2025-02-28
505 upgrade react ga module 1.2.1

jaydonkrooss:


config/.env.sample
```diff
@@ -25,8 +25,6 @@ FRAME_DOMAIN=canvas-test.it.umich.edu
 # MAX_AGE_IN_SEC=86400
 # Google Analytics environment Tracking ID
 # GOOGLE_ANALYTICS_ID=GA-XXXXXXXXX
-# OneTrust domain for cookie consent
-# ONE_TRUST_DOMAIN=DOMAIN-EXAMPLE
 # Domains allowed for CSP
 # ALLOWED_SCRIPT_DOMAINS=[]
```
pushyamig, line 29: Under this config, there is item that say `*.onetrust.com`does this mean this need to be removed as well?

pushyamig:
I see this error `App.tsx:44 Refused to load the script 'https://umich.edu/apis/umconsentmanager/consentmanager.js' because it violates the following Content Security Policy directive: "script-src 'self' *.google-analytics.com *.analytics.google.com *.googletagmanager.com cdn.cookielaw.org *.onetrust.com portal-infoassure.it.umich.edu". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.`

pushyamig:
I don't see the Banner pop up, may be due to CSP errors. I even deleted the previous accepted google analytics cookies.

config/.env.sample
pushyamig: I added `*.umich.edu`, but I still see the error i mentioned earlier

pushyamig: Ok That fixed the issue and banner is showing up.

ccm_web/client/src/App.tsx
```diff
@@ -25,10 +25,24 @@ function App (): JSX.Element {
...
+      gaHandlers.onConsentReject
+      ) {
+        const consentParams: InitializeConsentManagerParams = {
+            developmentMode: false,
+            alwaysShow: false,
+            onConsentApprove: gaHandlers.onConsentApprove,
+            onConsentReject: gaHandlers.onConsentReject,
+        }
+        umConsentInitialize(consentParams);
+    }
```
pushyamig, line 45: The banner pop's up when in EU region. When I accepted the Cookies when I refreshed the page or launched the CCM again, I notice the banner came up again, I felt Once the cookies are accepted, the banner shouldn't pop-up correct?

config/.env.sample
pushyamig: @jaydonkrooss I feel the `ALLOWED_SCRIPT_DOMAINS` need to update with accepted domains, looks like this is what needed for banner to work `["*.google-analytics.com", "*.analytics.google.com", "*.googletagmanager.com", "cdn.cookielaw.org", "*.umich.edu", "umich.edu"]`

don't know if `cdn.cookielaw.org` is needed.

I think you can add of this for documentation purposes

pushyamig:
I think the changes made here should be ported to `develop` branch. I don't think cherry pick option justify the pull, might need to  due manual edit.

@jaydonkrooss  Create an issue in CCM Python Rewrite board, for integrating the python way. This issue need not be ported immediately to develop but sometime soon after release of this done.

config/.env.sample
pushyamig: I think  it's ok just add to it in `.env.sample`

ccm_web/client/src/App.tsx
pushyamig: If you feel it is helpful, you can add the build from your branch to CCMDev and see if the problems goes away. I think the github action workflow `github.event_name == 'workflow_dispatch'` you can do a on demand build and output of it can be pushed to Openshift.

jonespm: I believe this is related to this issue I just filed https://github.com/tl-its-umich-edu/react-ga-onetrust-consent/issues/16.

config/.env.sample
jaydonkrooss: Added the links to env sample

ccm_web/client/src/App.tsx
pushyamig: @jaydonkrooss We were having a discussing in slack, Loophole might act different from ngrok. have you noticed any difference?

pushyamig (approved):
We still have the issue with cookie not being set for local run, I am ok merging it if this is solved on non-prod servers

ccm_web/client/src/App.tsx
jaydonkrooss: Its a little hard to understand, I think Loophole at least has the `um_privacy_consent...` cookie (which saves selected option on refresh) only when you open the app in a separate tab. As an I-frame in LTI, though, I can't get the cookie to track in neither loophole or ngrok.

PR #514, opened 2025-03-13
513 cookie consent banner hotfix

jaydonkrooss:


ccm_web/client/src/App.tsx
```diff
@@ -114,6 +94,24 @@ function App (): JSX.Element {
...
+                  }
+              },
+              externalLinkBlank: true
+            }
+        `}  
+        </script>
+        <script async src="https://umich.edu/apis/umconsentmanager/consentmanager.js"/>
+      </Helmet>
+    </HelmetProvider>
+    </>
```
pushyamig, line 114: In My Local CCM instance doesn't run from Loophole, I see LTI launch error and 403 errors

If I switch to Ngrok, I see LTI launches happen but Cookie banner won't launch from chrome, but it does from Edge. 

 the cookie behavior seems to be
 1.  `um_privacy_consent` usually appears before banner appears, 
 2. once User accept the cookie I see `um_privacy_conent_1222` being 
 
 So typically you would atleast see 2 cookies if user accepts
 
 From Edge I Cookies are not set, not sure this is because I am testing from Ngrok and not Loophole. I am having some difficulty with Proxy services.

pushyamig: I will debug more on this tomorrow.

ccm_web/client/src/App.tsx
```diff
@@ -114,6 +94,24 @@ function App (): JSX.Element {
...
+      <Helmet>
+        <script type="text/javascript">
+          {`window.umConsentManager = {
+              googleAnalyticsID: ${globals?.googleAnalyticsId ?`'${globals?.googleAnalyticsId}'`: false},
+              googleAnalyticsCustom: {
+                  streamConfig: {
+                      cookie_flags: "SameSite=None; Secure"
+                  }
+              },
+              externalLinkBlank: true
```
pushyamig, line 107: privacyStatement: false, might need to be added

ccm_web/client/src/App.tsx
```diff
@@ -114,6 +94,24 @@ function App (): JSX.Element {
         <Route path="/*" element={<NotFound />}/>
       </Routes>
     </Layout>
+    <HelmetProvider>
```
pushyamig, line 97: This can be a separate component

pushyamig: Some thing like this will also work 

```
import React, { useEffect } from 'react'

interface ConsentManagerProps {
  googleAnalyticsId: string | false;
}

const ConsentManagerBanner: React.FC<ConsentManagerProps> = ({ googleAnalyticsId }) => {
  useEffect(() => {
    if (googleAnalyticsId) {
      window.umConsentManager = {
        mode: 'prod', // values: 'prod', 'dev'
        customManager: {
          enabled: true,
          alwaysShow: false,
          preferencePanel: {
            beforeCategories: false, // HTML
            afterCategories: false   // HTML
          },
          googleAnalyticsCustom: {
            streamConfig: {
              cookie_flags: 'SameSite=None; Secure'
            }
          },
        },
        privacyUrl: '/privacy/',
        googleAnalyticsID: googleAnalyticsId ? googleAnalyticsId : false, // e.g. G-XXXXXXXXXX
        cookies: {
          necessary: [], // {name: '', domain: '', regex: ''}
          analytics: []
        }
      }

      const script = document.createElement('script')
      script.src = 'https://umich.edu/apis/umconsentmanager/consentmanager.js'
      script.async = true
      document.head.appendChild(script)
    }
  }, [googleAnalyticsId])

  return null
}

export default ConsentManagerBanner

```
`types/global.d.ts` -- These types could be inlined as well 
```
interface UmConsentManager {
  mode: string;
  customManager: {
    enabled: boolean;
    alwaysShow: boolean;
    preferencePanel: {
      beforeCategories: boolean;
      afterCategories: boolean;
    };
    googleAnalyticsCustom: object
  };
  privacyUrl: string;
  googleAnalyticsID: string | false;
  cookies: {
    necessary: Array<{ name: string; domain: string; regex: string }>;
    analytics: Array<{ name: string; domain: string; regex: string }>;
  };
}

interface Window {
  umConsentManager: UmConsentManager;
}

```

App.tsx


```
<HelmetProvider>
      <ConsentManager googleAnalyticsId={globals.googleAnalyticsId} />
</HelmetProvider>

This should work.

Comment (pushyamig) [reply ccm_web/client/src/App.tsx:97]: The above code can remove certain stuff from ConsentManager and fix with 

```
googleAnalyticsID: ${globals?.googleAnalyticsId ?`'${globals?.googleAnalyticsId}'`: false},
              googleAnalyticsCustom: {
                  streamConfig: {
                      cookie_flags: "SameSite=None; Secure"
                  }
              },
              externalLinkBlank: true
              ```
              This formart

pushyamig:
I will to a another round of testing in the afternoon.

ccm_web/client/src/App.tsx
pushyamig: @jaydonkrooss , I won't push for this change, if things are working without the component approach. I was just suggesting since I figured it is possible to write all this in React as well.

pushyamig:
I think this is working as expected, we will keep the Helmet approach since it seems to be working, did not see any side effects. With the Python Rewrite it won't be needed and could use the Myla style configuration since context Processor also used with Python rewrite https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1647/files

I tested from Edge browser, somehow i still has problem with ConsentBanner testing with chrome, I feel something with VPN and Chrome is conflicting. So I did all my testing from Edge browser.

pushyamig (approved):
This looking good, lets get this running on CCM test. I

pushyamig:
@jaydonkrooss I have `pu-ccm` image stream running, I that should not be used with QA official testing, but the build should be from ccm `main` branch which is configured as of now

PR #515, opened 2025-03-27
477: Create Section API handler to get sections

### LINKED ISSUE #477: GET: fetch lists of section in a course
Most of the CCM features like Add UM User, Add Non-UM User list section. so this all purpose API should be supported at the beginning of the API integration

jaydonkrooss:
add section endpoints: GET. Updating API scopes and refreshing the API token is required

Be sure to test:
* Swagger support
* GET UI workflow: Page 2 of Add U-M Users (both single and multiple) 
* GET UI workflow: Page 2 of Add Non-UM Users (both single and multiple)


backend/ccm/canvas_api/section_api_handler.py
```diff
@@ -0,0 +1,110 @@
...
+            # Call the Canvas API package to get section details.
+            logger.info(f"Getting section data for course")
+            # Get list of sections, including total_students info
+            sections = course.get_sections(include=['total_students'])
+            logger.info(f"Section data retrieved: {sections}")
+            
+            # Format the section data to return specific section info
+            formatted_sections = []
+            for section in sections:
+                formatted_sections.append({
```
pushyamig: I would like this to be data class

```
from dataclasses import dataclass


@dataclass
class Section:
    id: int
    name: str
    course_id: int
    nonxlist_course_id: int
    total_students: int = 0
```
```
    
    this way this object can be reused. I feel GET/PUT Course can be adopt a similar approach but missed in that PR

File: backend/ccm/canvas_api/section_api_handler.py [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+        try:
+            canvas_api: Canvas = CANVAS_CREDENTIALS.get_canvasapi_instance(request)
+            logger.info(f"Retrieving course for section data with course_id: {course_id}")
+            course: Course = canvas_api.get_course(course_id)
+            logger.info(f"Course data retrieved: {course}")
+            
+            # Call the Canvas API package to get section details.
+            logger.info(f"Getting section data for course")
+            # Get list of sections, including total_students info
+            sections = course.get_sections(include=['total_students'])
```
pushyamig: Do this handles Pagination?

backend/ccm/canvas_api/section_api_handler.py
```diff
@@ -0,0 +1,110 @@
...
+        Get section data from Canvas.
+        """
+        try:
+            canvas_api: Canvas = CANVAS_CREDENTIALS.get_canvasapi_instance(request)
+            logger.info(f"Retrieving course for section data with course_id: {course_id}")
+            course: Course = canvas_api.get_course(course_id)
+            logger.info(f"Course data retrieved: {course}")
+            
+            # Call the Canvas API package to get section details.
+            logger.info(f"Getting section data for course")
```
pushyamig: Both Log statement can combined to `Getting section data for course {course_id}`

backend/ccm/canvas_api/section_api_handler.py
```diff
@@ -0,0 +1,110 @@
...
+            # Get list of sections, including total_students info
```
pushyamig: `course.get_sections(include=['total_students'])` This could also be 
`canvas_api.get_course(course_id).course.get_sections(include=['total_students'])`

For some reason `canvasapi` pkg is have it as separate step.

backend/ccm/canvas_api/section_api_handler.py
```diff
@@ -0,0 +1,110 @@
+import logging
```
pushyamig: we can debate on this the name here can be course_section_api_handler, since it start with /course/course_id/ and a pure Section API like /section/section_id/ with out course reference, we have like that

```
'url:POST|/api/v1/sections/:id/crosslist/:new_course_id',
  'url:DELETE|/api/v1/sections/:id/crosslist',
  // Enrollments
  'url:GET|/api/v1/sections/:section_id/enrollments',
  'url:POST|/api/v1/sections/:section_id/enrollments',
```

backend/ccm/models.py
```diff
@@ -1,3 +1,10 @@
 from django.db import models
 
 # Create your models here.
```
pushyamig: I feel this should go in `canvas_api` folder. Since outside of it I feel they don't have any relevance. I feel the defaults `models.py` is used for django Model/migration. Maybe `canvas_models.py`

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: I didn't initially think about this, but with my brief research it seems that `get_sections` from canvasapi returns a [PaginatedList](https://github.com/ucfopen/canvasapi/blob/develop/canvasapi/paginated_list.py). 

So the way this code is written it should properly paginate, because of the iteration needed for formatting on line 51:
>`for section in sections:`

backend/ccm/canvas_api/canvasapi_serializer.py
```diff
@@ -1,5 +1,11 @@
...
 class CourseSerializer(serializers.Serializer):
     # Define the fields you want to update. Adjust fields according to the Canvas API.
-    newName = serializers.CharField(max_length=255, required=True)
\ No newline at end of file
+    newName = serializers.CharField(max_length=255, required=True)
+
+class SectionSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Section
+        fields = ['id', 'name', 'course_id', 'total_students', 'nonxlist_course_id']
```
pushyamig: As far my understanding goes,  Here I don't think Section is Model in our Database. 
```
ModelSerializer in Django REST Framework (DRF) is a shortcut for creating serializers that map directly to Django models. 
```
https://www.django-rest-framework.org/api-guide/serializers/#modelserializer

What you have done here is what we did created a User serializer for `auth_user` table. HEre is the referance on this https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/develop/backend/ccm/serializer.py

pushyamig: I don't think we need a serializer is needed for GET in this scenario. Since I only used serializer if request has some data to send like POST/PUT

backend/ccm/canvas_api/section_api_handler.py
pushyamig: I think log statement L39 should be enough, I don't see the need for log statements on L41 and L44.

backend/ccm/canvas_api/section_api_handler.py
pushyamig: So Canvas by defaults send only 10 responses of any type and Max 100.  If we want to get more response we need to mention per_page=100

So I couldn't tell from the code if the Pagination in `canvasapi` is returning 10 or `canvasapi` has a default as 100

backend/ccm/models.py
```diff
@@ -1,3 +1,10 @@
...
+
+class Section(models.Model):
+    id = models.AutoField(primary_key=True)
+    name = models.CharField(max_length=255)
+    course_id = models.IntegerField()
+    total_students = models.IntegerField()
+    nonxlist_course_id = models.IntegerField()
```
pushyamig: I feel this is redundant, since Section is not a Table in our Database.

backend/ccm/models.py
pushyamig: I think YOu can ignore this comment. but `models.py` is not needed

backend/ccm/models.py
```diff
@@ -1,3 +1,3 @@
 from django.db import models
```
pushyamig: Delete this file

pushyamig: I did not mean to delete the file, I think since there are no changes to the file, this file shouldn't appear in the PR. Sorry for the confusion

backend/ccm/canvas_api/section_api_handler.py
```diff
@@ -0,0 +1,74 @@
...
+CANVAS_CREDENTIALS = CanvasCredentialManager()
+
+@dataclass
+class Section:
+    id: int
+    name: str
+    course_id: int
+    total_students: int
+    nonxlist_course_id: int
+class CanvasSectionAPIHandler(LoggingMixin, APIView):
```
pushyamig: I feel this class should be renamed for `CourseSectionAPIHandler`, which tell you that this is handling course specific things `course/:course_id/section`, you can drop `Canvas` since it will make it too long

Also rename the file course_section_api_handler.py

backend/ccm/canvas_api/section_api_handler.py
```diff
@@ -0,0 +1,74 @@
...
+                section_data = Section(
+                    id=section.id,
+                    name=section.name,
+                    course_id=section.course_id,
+                    total_students=section.total_students,
+                    nonxlist_course_id=section.nonxlist_course_id
+                )
+                #serialize the section data
+                serializer = SectionSerializer(section_data)
+                formatted_sections.append(serializer.data)
```
pushyamig: I don't think `Section` dataclass and  `SectionSerializer` **both** are needed here, Let me think about this little more and get back to you.

pushyamig:
I haven't run this locally yet, I will do that in the afternoon. Meanwhile I have some comment left that can be addressed.

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: I'm not sure the line numbers you're seeing are the same on my IDE, are you referring to `Retrieving course for section data with course_id: {course_id}` being enough for `get_sections`?

pushyamig: Yes, I was referring to Line number from Github here.

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: Sure, I guess it may be a little overkill on formatting to use both the serializer and dataclass. Do you think we wouldn't use dataclass for the GET call, and we'd use it somewhere else?

pushyamig: Here is my thinking on it with `SectionSerializer`
1. Usually we add Serializer if we are doing something afterward like store in DB. Here is some of the text from DRF docs `Serializers allow complex data such as querysets and model instances to be converted to native Python datatypes that can then be easily rendered into JSON, XML or other content types. Serializers also provide deserialization, allowing parsed data to be converted back into complex types, after first validating the incoming data.`
2. Here we are simply taking the response object that is needed for us and send it to FE, if 200 response from Canvas Means the JSON response is a good response. so don't need to serialize step further.
3. If we are adding `SectionSerializer` for Response obj then we need to add `SectionSerializer.is_valid()` check again. It's rare to anticipate Canvas 200 response  had bad response, if it is then that will be caught as Exception.
5. Only benefit i see with `SectionSerializer` in response object is if it return a list and `SectionSerializer` don't need iteration to get the data needed, but I don't know if that is true since I did not pull the code locally
6. I suggested Dataclasses Section can be reused with url:POST|/api/v1/courses/:course_id/sections'. This will keep thing bit organized
7. Serializing the input request data is good since browser request can be malicious we need to do server side validation, also if FE And BE development is done at separate time then having a Server validation with Serialization is a quick way to tell what is BE expecting with Request Object

I am learning DRF as part of this project which I did not used it before. So I might have my Blind spots as well.

jaydonkrooss: I see. So the main focus of serializer is more useful to us in validating request data safely, for example with POST. I haven't tried using SectionSerializer as a way to paginate the data retrieved as a list. With your advice, I'm fine with leaving it as a for-loop for safe processing.

backend/ccm/models.py
jaydonkrooss: edited to remove the models.py changes

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: I agree with this; API renamed  to course_section

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: My recent changes combined the API calls in a one-liner. Also went ahead with one log statement showing the course_id to replace the others.

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: created the dataclass at the top of the file.

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: combined into one log statement

backend/ccm/canvas_api/canvasapi_serializer.py
jaydonkrooss: I've removed both serializer and model for the GET call

backend/ccm/canvas_api/section_api_handler.py
jaydonkrooss: good tip, renamed to "Course Section"

backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -0,0 +1,65 @@
...
+            # Format the section data to return specific section info
+            formatted_sections = []
+            for section in sections: 
+                formatted_sections.append({
+                    "id": section.id,
+                    "name": section.name,
+                    "course_id": section.course_id,
+                    "total_students": section.total_students,
+                    "nonxlist_course_id": section.nonxlist_course_id
+                })
```
pushyamig: @jaydonkrooss I did some research on the `SectionSerializer` you created before, the advantage 
1. You don't need iterate over this list of section, which I feel is cool. But Canvas Object that is returned and serializer attributes name must match, if that is not the case, I think they will be a way to map that . But looks the the FE object expecting is same as what Canvas Object variable. so that shouldn't be a problem

```
serializer = SectionSerializer(sections, many=True)
            return Response(serializer.data, status=HTTPStatus.OK)

```

but I do feel that since we are passing response object through Serializer we need to have a error check, some thing like below
```
if not serializer.is_valid():
            err_response: CanvasHTTPError = CanvasCredentialManager.handle_serializer_errors(serializer.errors, request.data)
            return Response(err_response.to_dict(), status=err_response.status_code)
```
https://www.django-rest-framework.org/api-guide/serializers/#validation

backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -0,0 +1,65 @@
...
+
+CANVAS_CREDENTIALS = CanvasCredentialManager()
+
+@dataclass
+class Section:
+    id: int
+    name: str
+    course_id: int
+    total_students: int
+    nonxlist_course_id: int
```
pushyamig: This is not being used in the class here

pushyamig:
Locally this PR is working fine, and I see this API call is hooked to multiple features which is pulling this API response.

pushyamig:
I have some feedback around serializer which could be considered

backend/ccm/canvas_api/course_section_api_handler.py
jaydonkrooss: removed

backend/ccm/canvas_api/course_section_api_handler.py
jaydonkrooss: after some discussion and research I found a way to make a generic CanvasObject Serializer which accomplishes the step of serializing the Section type into the proper native python datatypes. Because the serializer is used for responses and is readonly, I don't think the usual data validation is needed, according to the DRF documentation.

backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -0,0 +1,49 @@
...
+from .exceptions import CanvasHTTPError
+from canvas_oauth.exceptions import InvalidOAuthReturnError
+
+from backend.ccm.canvas_api.canvas_credential_manager import CanvasCredentialManager
+
+from rest_framework_tracking.mixins import LoggingMixin
+
+logger = logging.getLogger(__name__)
+
+CANVAS_CREDENTIALS = CanvasCredentialManager()
```
jonespm: I didn't test this but I think this works but it might be better to have this as a local variable and created in an `__init__ ` probably something like this?

```
    def __init__(self, credential_manager=None):
        self.credential_manager = credential_manager or CanvasCredentialManager()
```

Then the test could probably just inject a mock credential manager instead of patching. (Though you still could just directly patch `backend.ccm.canvas_api.canvas_credential_manager.CanvasCredentialManager.get_canvasapi_instance`)

Was there some reason you went with a global here? I think it's _fine_ either was just want to use the cleanest pattern.

backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get(self, request: Request, course_id: int, per_page: int = 100) -> Response:
+        """
+        Get section data from Canvas.
+        """
+        try:
+            canvas_api: Canvas = CANVAS_CREDENTIALS.get_canvasapi_instance(request)
+            
+            # Call the Canvas API package to get section details.
+            logger.info(f"Retrieving course for section data with course_id: {course_id}")
```
jonespm, line 43: I think it makes more sense for these 2 log messages to be debug, but we might have to create an issue to review other log messages once we get closer to prod.

backend/ccm/canvas_api/urls.py
```diff
@@ -1,6 +1,9 @@
 from backend.ccm.canvas_api.course_api_handler import CanvasCourseAPIHandler
 from django.urls import path
 
+from backend.ccm.canvas_api.course_section_api_handler import CourseSectionAPIHandler
+
 urlpatterns = [
   path('course/<int:course_id>/', CanvasCourseAPIHandler.as_view() , name='course'),
+  path('course/<int:course_id>/sections/', CourseSectionAPIHandler.as_view() , name='courseSection')
```
jonespm, line 8: We have `APPEND_SLASH=False` in the settings. We may want to just not add on slashes at the end of these API calls. 

That would avoid the change to the frontend.

I see the other pattern for course id had a slash. What do you think @pushyamig?

In [Canvas API](https://canvas.instructure.com/doc/api/courses.html#method.courses.show) there's no slashes at the end of the URL's for their API calls so it would be consistent with Canvas to leave them off and better to just make a decision on this now early.

backend/ccm/canvas_api/course_section_api_handler.py
jaydonkrooss: I didn't consider it as a local variable, I was mainly following the pattern that @pushyamig set up with the course API & unit tests. I agree that it could be a cleaner improvement, will try this.

backend/ccm/canvas_api/urls.py
pushyamig: I think I am ok, with this suggestion, But I think we need to check with LTI Login URL `/init/1223` works the same way without it. I think currently it is configured as ``/init/1223/`

backend/ccm/canvas_api/course_section_api_handler.py
pushyamig: I felt Prod is configured to WARN status. so I feel the info might not be logged at all. 

I think the way the Django Logging configured, changing to logger.debug it won't log logger.debug with below configuration

ROOT_LOG_LEVEL=INFO
DJANGO_LOG_LEVEL=INFO
DEBUG=True

unless you change DJANGO_LOG_LEVEL or ROOT_LOG_LEVEL to debug then there will be clutter of logs from underneath libraries which will be noise during developement

backend/ccm/canvas_api/course_section_api_handler.py
pushyamig: I think if we are changing it then we need to change it `course_api_handler`

backend/ccm/canvas_api/urls.py
pushyamig: We can deal with this as a separate issue not here

backend/ccm/canvas_api/course_section_api_handler.py
pushyamig: >I think if we are changing it then we need to change it course_api_handler

We can deal with this as a separate issue not here

backend/ccm/canvas_api/course_section_api_handler.py
jaydonkrooss: Considering we'll keep Warn level on prod, I'm going to leave the log level to info for now. If its too noisy in development we could change it.

backend/ccm/canvas_api/urls.py
jaydonkrooss: created a new issue: https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/516

pushyamig:
I will finish the review this morning

backend/ccm/canvas_api/canvasapi_serializer.py
```diff
@@ -1,5 +1,37 @@
...
+from canvasapi.section import Section
 
 class CourseSerializer(serializers.Serializer):
     # Define the fields you want to update. Adjust fields according to the Canvas API.
-    newName = serializers.CharField(max_length=255, required=True)
\ No newline at end of file
+    newName = serializers.CharField(max_length=255, required=True)
+
+class CanvasObjectReadonlySerializer(serializers.BaseSerializer):
+    """
```
pushyamig, line 9: Nitpick, I feel the name could be shorted `CanvasObjROSerializer` or `CanvasObjectROSerializer`

backend/ccm/canvas_api/canvasapi_serializer.py
```diff
@@ -1,5 +1,37 @@
...
+                data[attr] = value 
+            elif isinstance(value, list):
+                # For lists, check if they contain CanvasObjects or primitives
+                if all(isinstance(item, (str, int, bool, float, type(None))) for item in value):
+                    data[attr] = value
+                else:
+                    data[attr] = [self.to_representation(item) for item in value]
+            else:
+                # Fallback for other types
+                data[attr] = str(value)
```
pushyamig, line 40: I feel this approach is good start, But I really did not feel the need to send the entire Section Object response from Canvas to Frontend end when making an API. 

Sorry if there was any confusion that led to this approach.

Currently this is what I See with the code change

```
 {
        "course_id": 403334,
        "created_at": "2025-03-20T16:54:22Z",
        "created_at_date": "2025-03-20 16:54:22+00:00",
        "end_at": null,
        "id": 846977,
        "integration_id": null,
        "name": "sec 16",
        "nonxlist_course_id": null,
        "restrict_enrollments_to_section_dates": null,
        "sis_course_id": "ccmC1626457811556-24",
        "sis_import_id": null,
        "sis_section_id": null,
        "start_at": null,
        "total_students": 0
    },
 ```

 I Still feel only send what is needed and FE is expecting and keeping the response slim
 
 ```
  {
        "course_id": 403334,
        "id": 846977,
        "name": "sec 16",
        "nonxlist_course_id": null,
        "total_students": 0
    },
    ```

Comment (pushyamig) [reply backend/ccm/canvas_api/canvasapi_serializer.py:40]: a quick ChatGPT approach suggested this ( Please change as you see fit)
```

allowed_fields = {"course_id", "id", "name", "nonxlist_course_id", "total_students"}
            serializer = CanvasObjectReadonlySerializer(sections, many=True, allowed_fields=allowed_fields)
```
            
  ```
class CanvasObjectReadonlySerializer(serializers.BaseSerializer):
    """
    Serializer for generic Canvas objects from the Canvas API
    Adapted from the Django REST Framework documentation:
    https://www.django-rest-framework.org/api-guide/serializers/#creating-new-base-classes
    """
    def __init__(self, *args, allowed_fields=None, **kwargs):
        super().__init__(*args, **kwargs)
        self.allowed_fields = allowed_fields

    def to_representation(self, instance):
        data = {}
        for attr in dir(instance):
            if self.allowed_fields and attr not in self.allowed_fields:
                continue  # Skip attributes not in allowed_fields
            value = getattr(instance, attr)
            if attr.startswith('_'):
                continue  # Skip private/internal attrs
            if callable(value):
                continue  # Skip methods
            if hasattr(value, '__dict__') and isinstance(value, object):
                data[attr] = self.to_representation(value)
            elif isinstance(value, (str, int, bool, float, type(None))):
                data[attr] = value
            elif isinstance(value, list):
                if all(isinstance(item, (str, int, bool, float, type(None))) for item in value):
                    data[attr] = value
                else:
                    data[attr] = [self.to_representation(item) for item in value]
            else:
                data[attr] = str(value)
        return data
```

pushyamig: I think I am fine this take an different Issue and solve it there.

pushyamig: My Personal thoughts give FE just what is needed, this way we are not exposing too much info to FE. @jonespm please share you thought on this approach I am suggesting

pushyamig:
This PR is functional and handling Happy and unhappy case. Unit tests are written supporting the code changes. 

I have one comment regarding entire response Object sent to FE, I would like to hear thoughts from @jaydonkrooss and @jonespm, but we can merge it and could deal that as separate issue

backend/ccm/canvas_api/canvasapi_serializer.py
jaydonkrooss: This is a good approach, I added this in with small differences to cover any edge cases. Just making sure that we convert to primitives before filtering, in case the response comes back with lists or nested objects.

backend/ccm/canvas_api/canvasapi_serializer.py
jaydonkrooss: Shortened to CanvasObjectROSerializer

backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -0,0 +1,53 @@
...
+        try:
+            canvas_api: Canvas = self.credential_manager.get_canvasapi_instance(request)
+            
+            # Call the Canvas API package to get section details.
+            logger.info(f"Retrieving course for section data with course_id: {course_id}")
+            # Get list of sections, including total_students info
+            sections = canvas_api.get_course(course_id).get_sections(include=['total_students'], per_page=per_page)
+            allowed_fields = {"course_id", "id", "name", "nonxlist_course_id", "total_students"}
+            serializer = CanvasObjectROSerializer(sections, allowed_fields=allowed_fields, many=True)
+            logger.info(f"Section data retrieved: {serializer.data}")
```
pushyamig: can this log statement be logger.debug? this is cluttering the logs since I have a more than 100 section the course i am testing.

pushyamig (approved):
The changes Looks Good, The Serializer pattern for response object is nice add and support generalized Canvas Response object that can be reused by all the Canvas API calls.

backend/ccm/canvas_api/course_section_api_handler.py
jaydonkrooss: I went ahead and changed this to debug, alongside a short log statement to show what allowed_fields are being returned

Issue #511, opened 2025-03-10
Create a backend Endpoint to create single and multiple section


pushyamig:
I think create section is part of multiple workflow, so it best to create the backend endpoint and separate the Frontend integration. 

SwaggerUI can be used for the testing the backend endpoint

zqian:
Verified on SwaggerUI for creating sections via POST /api/course/{course_id}/sections

Issue #510, opened 2025-03-10
Integrate single/multiple section creation frontend with backend


pushyamig:
Feature like enroll users (umich and non-umich) allows for single section creation, so integration between FE and BE is needed

pushyamig:
This issue has a dependency on #477 since the FE validating the same name section are not created so a GET call is made to change if the new name is not same as existing name

pushyamig:
This issue might be automatically solved or have minor change once the #511 is resolved. With out #477 we fully cannot resolve the creating worked or not.

ktowneUM:
tested
add um users (one section)
add non-um user
create sections

PR #517, opened 2025-04-14
Issue 476 create section

### LINKED ISSUE #476: POST: Create section/s
Add support for creating single section and Bulk add support.

Need to add support for making concurrent API call, needs some research what is good for pkg available

pushyamig:
Fixes #476 #510 #511 

Step need while running locally, enable following scopes

     # Courses
    'url:GET|/api/v1/courses/:id',
    'url:PUT|/api/v1/courses/:id',
    # Sections
    'url:GET|/api/v1/courses/:course_id/sections',
    'url:POST|/api/v1/courses/:course_id/sections'

**Code Changes**
1. This PR addressing the complete integration of Create section single and bulk. Creating single section is used in the enrolling users features. 
    1. Bulk creating section using Concurrency approach and creating 60 is taking about 2 secs. Asyncio is used for doing this action, which is native Python lib feature.
3. In case of Canvas Token related error like revoked token, now the BE raises the error and DRF Custom Exception handler create a error response and FE does a browser redirect. FE Redirect is something already present
4. Refactoring the Canvas API Error response handler, this was done since the Bulk create section can return a mixed success/error response, only error response and all success. So `canvas_credential_manager.py`, `exception.py` handles all. There is a slight change in how `any_api_handler.py` call these classes when returning the error response.
5. Unit test are written to support all the use cases that is relevant to write test. 
6. API tracker should be tracking the Create section API calls

**Canvas Token anamolies testing steps**
1. Go to Canvas accounts --> Setting --> Delete the CCM API authorization ---> comeback to CCM ---> Make and API call from FE like get section, update course name --> Redirects to User Pre-OAuth page
2. When you add new features/API for CCM, then new scopes are added to Canvas Developer API registration, and scopes are added to CCM code as well, but the Canvas Access Token in the DB doesn't include permission with news scoped add, Canvas Throws unauthorized, in this case as well we are handling it using Browser redirect as described in `Code Changes` Step 2  
3. If Canvas API Registration is reset due to Prod sync, and we create a new API Key then the Canvas Access Token in DB are not valid with new API key. So this cases are handling it using Browser redirect as described in `Code Changes` Step 2 .Expired token in DB will trigger the use case.

### DIFF
--- a/backend/ccm/canvas_api/canvas_credential_manager.py
+++ b/backend/ccm/canvas_api/canvas_credential_manager.py
@@ -28,8 +28,9 @@ class CanvasCredentialManager:
   
   def handle_serializer_errors(self, serializer_errors: dict, input: str) -> CanvasHTTPError:
       logger.error(f"Serializer error: {serializer_errors} occured during the API call.")
-      # err_response: CanvasHTTPError = CanvasHTTPError(SerializerError(failed_input=str(input), serializer_error=serializer_errors))
-      err_response: CanvasHTTPError = CanvasHTTPError({'failed_input':str(input), 'serializer_error':serializer_errors})
+      # Create a SerializerError instance and pass it to CanvasHTTPError
+      serializer_error_instance = SerializerError(failed_input=str(input), serializer_error=serializer_errors)
+      err_response: CanvasHTTPError = CanvasHTTPError(serializer_error_instance)
       return err_response
   
   def handle_canvas_api_exceptions(self, exceptions: Union[HTTPAPIError, List[HTTPAPIError]]) -> CanvasHTTPError:
--- a/backend/ccm/canvas_api/canvas_credential_manager.py
+++ b/backend/ccm/canvas_api/canvas_credential_manager.py
@@ -1,63 +1,44 @@
 import logging
-from http import HTTPStatus
+from typing import List, Union
 from django.conf import settings
 from canvas_oauth.oauth import get_oauth_token
 from rest_framework.request import Request
-from canvas_oauth.models import CanvasOAuth2Token
 from canvas_oauth.exceptions import InvalidOAuthReturnError
 
 from canvasapi import Canvas
-from canvasapi.exceptions import (
-    BadRequest, Conflict, Forbidden, InvalidAccessToken, RateLimitExceeded,
-    ResourceDoesNotExist, Unauthorized, UnprocessableEntity
-)
-from .exceptions import CanvasHTTPError 
+from canvasapi.exceptions import (InvalidAccessToken, Unauthorized)
+from .exceptions import CanvasAccessTokenException, CanvasHTTPError, HTTPAPIError, SerializerError 
 
 logger = logging.getLogger(__name__)
 
 class CanvasCredentialManager:
-  
-  # Map of Canvas API exceptions to HTTP status codes
-  EXCEPTION_STATUS_MAP = {
-    BadRequest: HTTPStatus.BAD_REQUEST.value,
-    InvalidAccessToken: HTTPStatus.UNAUTHORIZED.value,
-    Unauthorized: HTTPStatus.UNAUTHORIZED.value,
-    Forbidden: HTTPStatus.FORBIDDEN.value,
-    RateLimitExceeded: HTTPStatus.FORBIDDEN.value,
-    ResourceDoesNotExist: HTTPStatus.NOT_FOUND.value,
-    UnprocessableEntity: HTTPStatus.UNPROCESSABLE_ENTITY.value,
-    Conflict: HTTPStatus.CONFLICT.value,
-    InvalidOAuthReturnError: HTTPStatus.FORBIDDEN.value
-}
 
   def __init__(self):
     super().__init__()
     self.canvasURL = f"https://{settings.CANVAS_OAUTH_CANVAS_DOMAIN}"
   
-  def get_canvasapi_instance(self, request: HTTPStatus) -> Canvas:
+  def get_canvasapi_instance(self, request: Request) -> Canvas:
     try:
       access_token = get_oauth_token(request)
     except InvalidOAuthReturnError as e:
       # This issue occurred during non-prod Canvas sync when the API key was deleted, but the token remained in CCM databases. Expired token will trigger the usecase.
       logger.error(f"InvalidOAuthReturnError for user: {request.user}. Remove invalid refresh_token and prompt for reauthentication.")
-      raise InvalidOAuthReturnError(str(e))
+      raise CanvasAccessTokenException()
     return Canvas(self.canvasURL, access_token)
   
-  def handle_canvas_api_exception(self, exception: Exception, request: Request, input: str = None) -> CanvasHTTPError:
-    logger.error(f"API error occcured : {exception}")
-    self.handle_revoked_token(exception, request)
-    
-    for class_key in self.EXCEPTION_STATUS_MAP:
-        if isinstance(exception, class_key):
-            return CanvasHTTPError(str(exception), self.EXCEPTION_STATUS_MAP[class_key], input)
-    return CanvasHTTPError(str(exception), HTTPStatus.INTERNAL_SERVER_ERROR.value, input)
-
-  def handle_revoked_token(self, exception, request):
-      if isinstance(exception, (InvalidAccessToken, InvalidOAuthReturnError)):
-          CanvasOAuth2Token.objects.filter(user=request.user).delete()
-          logger.error(f"Deleted the Canvas OAuth2 token for user: {request.user} since they might have revoked access.")
-  
-  def handle_serializer_errors(self, serializer_errors: dict, input: str = None) -> CanvasHTTPError:
+  def handle_serializer_errors(self, serializer_errors: dict, input: str) -> CanvasHTTPError:
       logger.error(f"Serializer error: {serializer_errors} occured during the API call.")
-      err_response: CanvasHTTPError = CanvasHTTPError(serializer_errors, HTTPStatus.INTERNAL_SERVER_ERROR.value, str(input))
+      err_response: CanvasHTTPError = CanvasHTTPError(SerializerError(failed_input=str(input), serializer_error=serializer_errors))
       return err_response
+  
+  def handle_canvas_api_exceptions(self, exceptions: Union[HTTPAPIError, List[HTTPAPIError]]) -> CanvasHTTPError:
+    logger.error(f"API error occurred: {exceptions}")
+    exceptions = exceptions if isinstance(exceptions, list) else [exceptions]
+    error_exceptions = [entry["error"] for entry in exceptions]
+    
+    if any(isinstance(exc, (InvalidAccessToken, Unauthorized)) for exc in error_exceptions):
+      # Invalid access token occures when a user revokes Canvas Authorization from Canvas Profile settings.
+      # Unauthorized happens when you add more API scopes but the User Authorization is still limited to eariler API scopes.
+      raise CanvasAccessTokenException()
+    
+    return CanvasHTTPError(exceptions)
--- a/backend/ccm/canvas_api/canvas_credential_manager.py
+++ b/backend/ccm/canvas_api/canvas_credential_manager.py
@@ -1,13 +1,11 @@
 import logging
-from typing import List, Union
 from django.conf import settings
 from canvas_oauth.oauth import get_oauth_token
 from rest_framework.request import Request
 from canvas_oauth.exceptions import InvalidOAuthReturnError
 
 from canvasapi import Canvas
-from canvasapi.exceptions import (InvalidAccessToken, Unauthorized)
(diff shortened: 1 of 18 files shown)


jaydonkrooss:
I'm starting my review of this, and I'll have more time to finish the test cases tomorrow.

backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -36,19 +36,81 @@ def get(self, request: Request, course_id: int, per_page: int = 100) -> Response
...
+    def create_section_sync(self, canvas_api: Canvas, course_id: int, section_name: str):
+        """Creates a section synchronously with automatic retry handling."""
+        try:
+            logger.info(f"Creating section: {section_name} for course_id: {course_id} at {time.strftime('%H:%M:%S')}")
+            section = canvas_api.get_course(course_id).create_course_section(course_section={"name": section_name})
+            
+            # Serialize the section and add total_students manually
+            serializer = CanvasObjectROSerializer(section, allowed_fields=self.course_section_allowed_fields)
+            serialized_data = serializer.data
+            serialized_data["total_students"] = 0  # Default value for total_students
```
jaydonkrooss: I previously mentioned an approach where we could have a descendant serializer class of `CanvasObjectROSerializer` create the `total_students` field automatically. I'm not sure if that pattern is necessary yet though, since we're not doing complex calculations for added data.

I think a simpler approach could be to pass in new data as `additional_data` similar to how we have the `allowed_fields` variable.
```
course_section_additional_data = { "total_students" : 0 } 
serializer = CanvasObjectROSerializer(
  section, 
  allowed_fields=self.course_section_allowed_fields, 
  additional_data=course_section_additional_data
)
```
and a change to the serializing function in `CanvasObjectROSerializer` like so:
```
def to_representation(self, instance):
        data = self.convert_to_primitives(instance)
        # Filter out fields not in allowed_fields
        if self.allowed_fields:
            data = {key: value for key, value in data.items() if key in self.allowed_fields}
        data.update(self.additional_data) 
        return data
```
Where `additional_data` could be empty `{}` by default. This is just a small tweak suggestion, not a blocking necessity.

backend/ccm/canvas_api/canvas_credential_manager.py
```diff
@@ -1,63 +1,46 @@
...
+      serializer_error_instance = SerializerError(failed_input=str(input), serializer_error=serializer_errors)
+      err_response: CanvasHTTPError = CanvasHTTPError(serializer_error_instance)
       return err_response
+  
+  def handle_canvas_api_exceptions(self, exceptions: Union[HTTPAPIError, List[HTTPAPIError]]) -> CanvasHTTPError:
+    logger.error(f"API error occurred: {exceptions}")
+    exceptions = exceptions if isinstance(exceptions, list) else [exceptions]
+    
+    # Check for token-related errors
+    if any(isinstance(exc.original_exception, (InvalidAccessToken, Unauthorized)) for exc in exceptions):
```
jaydonkrooss: Does it always make sense for the delete token + redirect flow to happen on an Unauthorized response? I could see it being convenient for our current API development, but otherwise, I'm not super familiar with all the cases in which this could happen for users. 

I don't know if this is possible, but what if the redirect flow happens and the user doesn't actually have the required API access? I feel that is more confusing and harder to diagnose than a regular error message. 

If you think such cases are covered elsewhere or the risk is minimal in practice we can ignore this and move on.

backend/ccm/canvas_api/canvas_credential_manager.py
```diff
@@ -1,63 +1,46 @@
...
-            return CanvasHTTPError(str(exception), self.EXCEPTION_STATUS_MAP[class_key], input)
-    return CanvasHTTPError(str(exception), HTTPStatus.INTERNAL_SERVER_ERROR.value, input)
-
-  def handle_revoked_token(self, exception, request):
-      if isinstance(exception, (InvalidAccessToken, InvalidOAuthReturnError)):
-          CanvasOAuth2Token.objects.filter(user=request.user).delete()
-          logger.error(f"Deleted the Canvas OAuth2 token for user: {request.user} since they might have revoked access.")
-  
-  def handle_serializer_errors(self, serializer_errors: dict, input: str = None) -> CanvasHTTPError:
+  def handle_serializer_errors(self, serializer_errors: dict, input: str) -> CanvasHTTPError:
```
jaydonkrooss: Sorry that this is litigating an older change, but I'm wondering whether serializer errors should be under the responsibility of the `canvas_credential_manager`. Maybe this could be done elsewhere or the way serializer errors are handled could be reworked in general, since I don't know if the return value of CanvasHTTPError is needed here.

backend/ccm/canvas_api/exceptions.py
jaydonkrooss, line 1: The new exception logic is a little hard to follow and might benefit from being refactored.

From my understanding of how you wrote this,`HTTPAPIError` is only used to deliver the exception information as an input to `CanvasHTTPException`. Wouldn't it make more sense for just one class to be used for exception information, which also contains the logic to store invalid_input, map the statuses, and return to_dict? I understand you rewrote`CanvasHTTPException` to additionally help with processing a list of exceptions as well, but I'm wondering if that is a separate responsibility, possibly for a new Exception subclass like `AggregatedAPIError` or something.

@jonespm I know that you suggested about organizing exception-handling in the last PR, if you have any suggestions feel free to share or correct me if I'm wrong.

backend/ccm/canvas_api/canvas_credential_manager.py
```diff
@@ -1,63 +1,46 @@
...
       logger.error(f"Serializer error: {serializer_errors} occured during the API call.")
-      err_response: CanvasHTTPError = CanvasHTTPError(serializer_errors, HTTPStatus.INTERNAL_SERVER_ERROR.value, str(input))
+      # Create a SerializerError instance and pass it to CanvasHTTPError
+      serializer_error_instance = SerializerError(failed_input=str(input), serializer_error=serializer_errors)
+      err_response: CanvasHTTPError = CanvasHTTPError(serializer_error_instance)
       return err_response
+  
+  def handle_canvas_api_exceptions(self, exceptions: Union[HTTPAPIError, List[HTTPAPIError]]) -> CanvasHTTPError:
```
jaydonkrooss: I was thinking about whether this fits in under "Canvas credential manager", and I thought of a suggestion: 

Related to my comment on [refactoring exception logic](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/517/files#r2047703965), this function may not need to use process one exception type and return another. It could instead take just one CanvasHTTPError type and check for Token issue to be raised, and you wouldn't need to return anything (just let the exception be caught and trigger DRF custom exception flow). In this case it ought to  be renamed, something like `check_invalid_auth_token`

backend/ccm/canvas_api/course_api_handler.py
```diff
@@ -64,21 +60,25 @@ def get(self, request: Request, course_id: int) -> Response:
...
-            put_course_res: str= course.update(course={'name': update_data.get("newName"), 'course_code': update_data.get("newName")})
-            formatted_course = {'id': course.id, 'name': put_course_res, 'enrollment_term_id': course.enrollment_term_id }
-            return Response(formatted_course, status=HTTPStatus.OK)
-        except (CanvasException, InvalidOAuthReturnError, Exception) as e:
-            err_response: CanvasHTTPError = CANVAS_CREDENTIALS.handle_canvas_api_exception(e, request, str(request.data))
-            return Response(err_response.to_dict(), status=err_response.status_code)
+            course_name_update_res: str= course.update(course={'name': update_data.get("newName"), 'course_code': update_data.get("newName")})
+            serializer = CanvasObjectROSerializer(course, allowed_fields=self.course_allowed_fields)
+            serialized_data = serializer.data
+            serialized_data["name"] = course_name_update_res
```
jaydonkrooss: based on [my other comment](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/517/files#r2047272930), you can use an `additional_data` field for the serializer here as well.

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: We send FE `ccm_globals` that has what role the user has in the Canvas. [We don't give any permission to features in CCM that Canvas did not give them](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/main/ccm_web/client/src/models/feature.ts). With that said there is one exception with Canvas Non-UM user, but that usecase is handled differently. (With Okta every thing might change). 

So somebody  having access to Canvas/CCM due to the course admin role should see the error, unless they are doing something fishy. With that said, I see your point 

`Possible Solution`
but only way we separate fishy 401 to users experience with invalid scopes, is digging into the error message that canvas is [sending 'insufficient scopes on access token'](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/main/ccm_web/server/src/api/invalid.token.interceptor.ts#L15) there is risk that canvas might change it without informing. we had an issue with it when Canvas changed the [Revoked access token case](https://github.com/instructure/canvas-lms/commit/cdf0cecf03feff7ff9f372e8433148ec6f47c7a0)

I did bring this up during our weekly meeting and I did not hear any resistance to this approach. But I am ok to change it to `Possible Solution` but I don't see any other way to differentiate Regular 401 to insufficient. 

I think we need to add the error message to .env or load as secrets in case we see a prod issue with error message change from Canvas

Thoughts @jonespm

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: FE does Validate client side error like if `name` is > 255 or if user upload more then 60 section, I am just doing it as xtra cautionary step. I looked the serializer error as when making an Canvas API call, So I  bundled it with CanvasHTTPError.

backend/ccm/canvas_api/exceptions.py
pushyamig: `HTTPAPIError` was added since with Concurrent API Call, I am just collecting all the response success/error. I wanted a Object hold the error message caused by an input since FE report to the user ` your section A is some bad naming structure that canvas invalidate`. so I came up with this Class. This is generic one. 

Once I have all the Canvas API error collected as `HTTPAPIError` I give it to `CanvasHTTPException` so that the response can be formatted in the way FE expected.

```
{
    "statusCode": 400,
    "errors": [
        {
            "canvasStatusCode": 400,
            "message": "Canvas response body had unhandled shape: {\"errors\":{\"name\":[{\"attribute\":\"name\",\"type\":\"too_short\",\"message\":\"too_short\"}]}}",
            "failedInput": ""
        },
        {
            "canvasStatusCode": 400,
            "message": "Canvas response body had unhandled shape: {\"errors\":{\"name\":[{\"attribute\":\"name\",\"type\":\"too_short\",\"message\":\"too_short\"}]}}",
            "failedInput": ""
        }
    ]
}```

pushyamig: I change the logic so that I moved the `handle_canvas_api_exceptions` and  `handle_serializer_errors` to `CanvasErrorHandler` ( I renamed it)

I couldn't remove `HTTPAPIError` and all make the logic as part of single `CanvasErrorHandler`, this is the best I came up with. Some improvements are made as per request

backend/ccm/canvas_api/course_section_api_handler.py
pushyamig: Done

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: As discussed I am now checking for error message with Insufficient scope to narrow down the 401 status.

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: from keep the end user perspective, I did not differentiate separate error type, I looked at this as API error occurred when making canvas API call. This can be handled as separate, but I feel this is not a high priority

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: I removed then and now part of `CanvasErrorHandler` class

backend/ccm/canvas_api/course_api_handler.py
pushyamig: DOne

jaydonkrooss:
Local testing & unit tests perform well. The code changes look good to me as well, thanks for helping rework the exception organization again, I think it'll be a good pattern to work with.

jaydonkrooss (approved):

backend/ccm/canvas_api/canvas_credential_manager.py
pushyamig: Created an issue https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/520

PR #518, opened 2025-04-15
#482: Create Enrollments API handler for Third-Party Gradebook

### LINKED ISSUE #482: Third-party Gradebook Feature
This feature parse the CSV gradebook  based on section selected. So might need to wire this according

jaydonkrooss:
Fixes #482
For local testing, enable the following new scopes to the local canvas API configuration:

~~# Under "Sections"
'url:GET|/api/v1/sections/:id', 
'url:GET|/api/v1/courses/:course_id/sections/:id',~~
```

# Under "Enrollments"
'url:GET|/api/v1/sections/:section_id/enrollments',
```
~~NOTE: the use of UCF Canvas API for getting section via section ID adds TWO NEW scopes to the existing list of scopes (both section scopes `sections/:id'` and `courses/:course_id/sections/:id`). Upon replacing the Node backend, these two scopes will need to be added to the Dev, Test, & Production API configs.~~

**Code Changes**
1. Uses DRF & UCF Canvas API:  In the Third-Party Gradebook feature, enrollments are gathered after the checklist selection of course sections (before the CSV upload step). Retrieves a list of unique student ids based on the user's selected sections' IDs.
2. Changes the retrieval logic for performance: Instead of the frontend requesting enrollments for one section at a time and compiling login Ids client-side (`sections/<section_id>/students`), the frontend now makes **one request** with the list of section ids (`sections/students/?section_ids=<comma-separated list of section_id>`). This simplifies responsibility, slightly improves performance, and sets up further optimizations (such as Asyncio for concurrent requests) if needed in the future.
3. Tweaked the CanvasObjectROSerializer to properly serialize lists of CanvasObjects (in this case Enrollments from a Section). 

**Testing steps**
* Check that UI is properly making the single request for both single and multiple section cases
* The list of user login_ids retrieved should not contain any duplicates. If a student exists in two or more sections requested, their login_id appears only once in the successful response
* Check Exception handling, such as revoking Canvas API token successfully causes the redirect flow introduced in #517 
* test in Swagger UI 
* Unit tests pass


backend/ccm/canvas_api/section_enrollments_api_handler.py
```diff
@@ -0,0 +1,98 @@
...
+                section = canvas_api.get_section(section_id)
+            except (CanvasException, InvalidOAuthReturnError, Exception) as e:
+                self.canvas_error.handle_canvas_api_exceptions(HTTPAPIError(str(section_ids), e))
+                return Response(self.canvas_error.to_dict(), status=self.canvas_error.to_dict().get('statusCode'))
+            
+            try:
+                enrollments = section.get_enrollments(include=['user'], per_page=100)
+                allowed_fields = {"user"}
+                serializer = CanvasObjectROSerializer(enrollments,allowed_fields=allowed_fields, many=True)
+            except (CanvasException, InvalidOAuthReturnError, Exception) as e:
```
pushyamig: InvalidOAuthReturnError exception is now throws an error, which is caught by DRF custom exception handler and not needed to be caught here.

backend/ccm/canvas_api/section_enrollments_api_handler.py
```diff
@@ -0,0 +1,98 @@
...
+                api_errors.append(HTTPAPIError(str(section_id), e))
+                logger.error(f"Error retrieving enrollments for section_id {section_id}: {e}")
```
pushyamig, line 82: Can we club the statements for the 2 api call try/except together since they are throwing same exceptions?
 
 If an exception is thrown at let say `get_section` there is no point in making `get_enrollments`

pushyamig:
>(both section scopes sections/:id' and courses/:course_id/sections/:id)

Jaydon, for clarity I only see the 3rd party Feature with this PR only using scope `sections/:id` and `/api/v1/courses/:course_id/sections/:id` ? 

I ask since that is what the code reflects with following call
```
 section = canvas_api.get_section(section_id)
enrollments = section.get_enrollments(include=['user'], per_page=100)
```

I see don't see this scope `courses/:course_id/sections/:id` being used with this PR, but it is needs elsewhere.

pushyamig:
I have left few comment otherwise the feature is functional and tests are working as well.

backend/ccm/canvas_api/section_enrollments_api_handler.py
pushyamig: Please see `canvas_credential_manager.py` `get_canvasapi_instance` for additional details

backend/ccm/canvas_api/section_enrollments_api_handler.py
jaydonkrooss: I've changed the logic so that there is no longer a call to `get_section`. now there is only the logic to create a list of api_errors from `get_enrollments`

pushyamig: Nice Find @jaydonkrooss

jaydonkrooss:
> > (both section scopes sections/:id' and courses/:course_id/sections/:id)
> 
> Jaydon, for clarity I only see the 3rd party Feature with this PR only using scope `sections/:id` and `/api/v1/courses/:course_id/sections/:id` ?

@pushyamig  you were right, I think I was mistaken on the scopes. That being said, I just now put out an update to call `get_enrollments` with just one API call, so now there's only the enrollments scope needed for this endpoint.

backend/ccm/canvas_api/section_enrollments_api_handler.py
jaydonkrooss: removed the `InvalidOAuthReturnError ` catch for this case

backend/ccm/canvas_api/section_enrollments_api_handler.py
```diff
@@ -0,0 +1,93 @@
...
+                continue
+        
+        time_end = time.perf_counter()
+        logger.info(f"Time taken to get enrollments: {time_end - time_start:.2f} seconds")
+        if api_errors:
+            self.canvas_error.handle_canvas_api_exceptions(api_errors)
+            return Response(self.canvas_error.to_dict(), status=self.canvas_error.to_dict().get('statusCode'))
+        else:
+            return Response(list(unique_login_ids), status=HTTPStatus.OK)
+
```
pushyamig, line 94: Optional: Rather than if/else for readability is pattern will be better

```
if api_errors:
            error_response = self.canvas_error.to_dict(api_errors)
            return Response(error_response, status=error_response.get('statusCode', HTTPStatus.INTERNAL_SERVER_ERROR))
        
        return Response({"unique_login_ids": list(unique_login_ids)}, status=HTTPStatus.OK)
```

pushyamig (approved):
The PR improved the workflow the No. of calls FE making to BE when multiple section was chosen. 

Mocking the sections call to get to the section enrollments is improving the efficiency instead of 2 only 1 api is needed. We can apply this  pattern to other cases as well. @jaydonkrooss if you create an issue with number of places this change is need that will be great.

I think we are not taking async approach to multiple section more than 5 since the scope of the use of this feature limited. But can be improved in future if it is needed

backend/ccm/canvas_api/section_enrollments_api_handler.py
jaydonkrooss: good suggestion. added a quick edit to change the pattern

jaydonkrooss:
> Mocking the sections call to get to the section enrollments is improving the efficiency instead of 2 only 1 api is needed. We can apply this pattern to other cases as well. @jaydonkrooss if you create an issue with number of places this change is need that will be great.

@pushyamig  I created Issue #524 to evaluate. I think the main benefit would be to simplify/clean the code by reducing the number of API calls, such as not retrieving course for course sections for example. It might not be necessary for performance in that case, more of a 'nice-to-have' improvement.

> I think we are not taking async approach to multiple section more than 5 since the scope of the use of this feature limited. But can be improved in future if it is needed

We'll discuss this soon in the CCM meeting, but I checked with Dan and he agreed the "many sections" case is outside of the requested feature scope.

pushyamig:
>@pushyamig I created Issue https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/524 to evaluate. I think the main benefit would be to simplify/clean the code by reducing the number of API calls, such as not retrieving course for course sections for example. It might not be necessary for performance in that case, more of a 'nice-to-have' improvement.

Thanks for creating the Issue

pushyamig:
> but I checked with Dan and he agreed the "many sections" case is outside of the requested feature scope.

Thanks for the follow up. I feel we need to update this info to the issue description so that when testing we will keep the testing close to feature issue

pushyamig:
I resolved all my comments, the PR can be merged

Issue #523, opened 2025-05-05
Enrolling user with Canvas api


pushyamig:
This Issue handling wiring FE and BE by taking the FE request of enrolling user, and making it as backend task. The FE is returned with some message "You task in progress and will send an email once done". FE code change is possible. 

The scope is that we need to handle 1000+ enrollment, so  we might need to adopt Async Operation 

```
 # Create tasks for enrolling
  tasks = [enroll_user_canvas(enrollment) for enrollment in enrollments]
  # Gather results concurrently
  results = await asyncio.gather(*tasks)
```
I might think this might run for >5min for 1000+, probably we will know more the time taken as we go

pushyamig:
This issue depend on #521

jaydonkrooss:
Note: look into canvas API call to submit to canvas notification inbox ( possibly giving user the option to submit to email or canvas inbox )

jaydonkrooss:
Note: I'll look into whether we can optionally have the user submit to the Canvas Inbox as well as (or instead of) email.

Could open as a separate issue, more of a nice-to-have feature, not essential to the next release.

ktowneUM:
Text when adding bulk enrollments: 
Adding new users is currently in progress. Please check your email for confirmation of success or notification of any issues.

ktowneUM:
done

Issue #481, opened 2024-10-31
Add non U-M User Feature


pushyamig:
Adding Non-UM users is going to be different from current CCM prod. These are the following changes that going to happen

1. IAM Friend web Service is going to be retired and will be replaced with Okta Guest Service
2. We no longer  need Cirrus integration through CCM

The workflow is going to be

1. New users will be created in Canvas same way as existing process. 
     2. creating Canvas Login ID by replacing `test@example.edu` with `test+example.edu`. 
     3. **We will be using the CCM Admin token enabled for create user API call**. This is a special case since feature allowed by Teachers, TA. None of those roles have privileges for creating user at root which is only reserved for Canvas Admins. 
       
                      const endpoint = `accounts/${accountID}/users`
                      const method = HttpMethod.Post
                       const body = {
                      user: {
                        // API doesn't provide separate given- and surname fields
                        name: fullName,
                        sortable_name: sortableName,
                        skip_registration: true
                      },
                      pseudonym: {
                        unique_id: loginId,
                        send_confirmation: false
                      },
                      communication_channel: {
                        type: 'email',
                        address: email,
                        skip_confirmation: true
                      },
                      force_validations: false
                      }
   
5. CCM will send email to  create a Guest account in Okta
6. Enrolling user to course section once added to Canvas Root. The above process is same  for **1. Adding single User 2. Bulk enrolling users** to section
Details of the new and old workflow is show [here](https://docs.google.com/presentation/d/1ockQg84i-Zh5sp29wP_dyPf9oE5yQCIUBgDayTWolFk/edit?slide=id.g36b86fe2394_3_53#slide=id.g36b86fe2394_3_53)

pushyamig:
Before we start this we need confirmation from IAM team direction what service they suggest for adding Non-UM user

pushyamig:
I am assigning this task to myself, @jaydonkrooss when somebody times opens up first they can work on it first. LMK your thoughts

ktowneUM:
add non-um user was thoroughly tested

PR #526, opened 2025-05-29
I521 django q2 setup

### LINKED ISSUE #521: Create the pipeline for the running the process using Django-q2
This issue setup the workflow that is needed for running a tasks as background tasks, this setup could be used when enrolling users to canvas. 

The goal is to run this backend task process via supervisor rather running it as docker service as DB or web.
There might be a need to setup a different supervisor file needed for prod/local.

#519 issue help understand that Django-q2 is a better process for enrolling 1000+ users and we will use same process any kind of enrollment number. 

Email setup process is needed informing users about success or failure

pushyamig:
Fixes #521 

This is establishing the pipeline for the DjangoQ with `django-q2` package. The goal is to run the enrolling user ( UM and Non-UM)  task as background instead of immediate API action. 

Django’s database backend as a message broker instead of Redis since the amount of request we would expect with Enrolling users in CCM
https://django-q2.readthedocs.io/en/master/configure.html#orm 

Configuration: Kept simple If you don't set any config defaults will apply
https://django-q2.readthedocs.io/en/master/configure.html#

Django Admin Console show state of Background Tasks

**Note**:  You will see the below warning, looking at the Package developer, this only effect when you have [Million of Failed ](https://github.com/django-q2/django-q2/issues/245#issuecomment-2448347017)of tasks. Based on this response we can [ignore](https://github.com/django-q2/django-q2/issues/245) 
```
WARNINGS:
django_q.Task: (models.W037) MySQL does not support indexes with conditions.
	HINT: Conditions will be ignored. Silence this warning if you don't care about it.

```

### DIFF
--- a/backend/ccm/background_tasks/math.py
+++ b/backend/ccm/background_tasks/math.py
@@ -1,0 +1,14 @@
+import logging
+logger = logging.getLogger(__name__)
+
+def add(task):
+    """
+    A simple background task that adds two numbers.
+    Call as below from Django Views
+      from django_q.tasks import async_task
+      task_data = {'a': 5, 'b': 8}
+      async_task('backend.ccm.background_tasks.math.add', task=task_data)
+    """
+    a = task.get('a', 0)
+    b = task.get('b', 0)
+    logger.info(f"Adding 2 numbers {a} and {b}: {a + b}")
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -280,12 +280,12 @@ SPECTACULAR_SETTINGS = {
 
 DRF_TRACKING_ADMIN_LOG_READONLY = True
 
+# https://django-q2.readthedocs.io/en/master/configure.html
 Q_CLUSTER = {
-    'name': 'DjangORM',
-    'workers': 2,
-    'timeout': 90,
-    'retry': 120,
-    'queue_limit': 50,
-    'bulk': 10,
+    'name': 'CCM_Cluster',
+    'workers': int(os.getenv('Q_CLUSTER_WORKERS', 4)),
+    'timeout': int(os.getenv('Q_CLUSTER_TIMEOUT', 1800)), # 30 minutes
+    'retry': int(os.getenv('Q_CLUSTER_RETRY', 3600)), # 1hr
+    'bulk': int(os.getenv('Q_CLUSTER_BULK', 5)),
     'orm': 'default'
 }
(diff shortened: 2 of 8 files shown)


pushyamig:
Test: 
1. Docker Build and start up is going on. 
2.  Test by calling the `math.py` from create section workflow. This is for testing only and as shown [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/526/commits/e7eccd6f29def9763cdbd485f73d9922b83a5ae3#diff-56c8db03544d7e47d457a9fa8b231a397a672298ae038170ec229a4d3872ea97)

backend/ccm/background_tasks/math.py
```diff
@@ -0,0 +1,14 @@
+import logging
```
pushyamig, line 1: I added this so that when called from Backend when doing an action of on User request like `Bulk create a sections` you could see if the task is executed as background task or not as demo'ed [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/526/commits/e7eccd6f29def9763cdbd485f73d9922b83a5ae3#diff-56c8db03544d7e47d457a9fa8b231a397a672298ae038170ec229a4d3872ea97)

pushyamig:
This PR is functional as a Prototype with example `math.py` task and will act as base for Enrolling and U-M #523  and Non-U-M users #481  issues

backend/settings.py
```diff
@@ -277,4 +278,14 @@
...
-DRF_TRACKING_ADMIN_LOG_READONLY = True
\ No newline at end of file
+DRF_TRACKING_ADMIN_LOG_READONLY = True
+
+# https://django-q2.readthedocs.io/en/master/configure.html
+Q_CLUSTER = {
+    'name': 'CCM_Cluster',
+    'workers': int(os.getenv('Q_CLUSTER_WORKERS', 4)),
+    'timeout': int(os.getenv('Q_CLUSTER_TIMEOUT', 1800)), # 30 minutes
+    'retry': int(os.getenv('Q_CLUSTER_RETRY', 3600)), # 1hr
```
pushyamig, line 288: We could adjust the `timeout` and `retry` after when we get some number on how long the task for enrolling user takes, probably based on the Enrolling U-M user we be used as benchmark since we will executing  1000+ enrollment per request

jaydonkrooss:
I'm starting on my review today

jaydonkrooss:
Local testing runs well after adding it to create section handler for my local instance. I see the logged result in the terminal logs as well as django admin page.

Unit tests pass as well. Code looks good to me.

jaydonkrooss (approved):

Issue #480, opened 2024-10-31
Merge/Unmerge Sections


pushyamig:
Add the unmerge Section API, using crosslist from canvasAPI

ktowneUM:
tested
teacher cannot unmerge
admins can unmerge

PR #527, opened 2025-06-03
525 Merge Sections GET - admin-level sections

### LINKED ISSUE #525: Get mergeable sections - Account Admin search
Add the two new methods of retrieving sections available to merge on the "Merge Sections" feature

### GET Sections for account admin (by search parameters)
* Frontend calls: `/api/admin/sections?sections?term_id=${termId}&instructor_name=${searchText}` OR `/api/admin/sections?sections?term_id=${termId}&course_name=${searchText}`
* Goal: as an account admin, get all sections in any other courses, for which:
     * The course is accessible to the user via account admin access 
     * The course's term matches the original course term
     * The course's name matches the given search request OR the course contains an instructor who's uniqname matches the given search request (toggleable in the search bar)
* UI Test plan: 
     * As a root account admin, go to the "Merge Sections" feature and type a search value of uniqname (then try course name); Results appear for other courses matching the same term and are filtered by instructor uniqname OR by course name.
     * Use "Act as user" to view the feature as a subaccount user and repeat both types of searches. The section results should ONLY originate inside the admin's accessible courses.

jaydonkrooss:
- [x] ~~fix empty accounts response from canvasAPI~~
- [x] ~~refactor section retrieval to be inside course search logic~~
- [x] ~~use asyncio TaskGroup to speed up admin sections retrieval~~
- [x] ~~write unit tests~~

Add new scope to canvas API: (under accounts) 
`'url:GET|/api/v1/accounts'`
`'url:GET|/api/v1/accounts/:account_id/courses'`

Gathers all sections for courses that match the term ID and search value (either instructor name or course name). Before getting courses, the code gets all accounts that are accessible to the user (defaults to the highest level parent accounts that don't overlap). Then the course search is set to run concurrently using asyncio for each account, and the sections which are attached to the course info are also searched concurrently for each course (similar to the [code for instructor sections GET](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/553/files#diff-4ee41444fead56f4951e834a6ec701515606977369966a210a14c040a9f76c43R70-R82)). So in total 3 stages of canvasAPI calls.

Test plan: 
for an admin-level user in Canvas test/beta, run the following searches on the "Merge Sections" page:
* Use "instructor name" search term to retrieve courses with sections
* Use "course name" search term to retrieve courses with sections 
* For both results, check the accuracy of the response data. As a sanity check, you can compare with Node version in test to compare the number of sections returned.

Lastly, inspect and run unit tests


jaydonkrooss:
I'm using Asyncio to reduce the runtime of the nested for loops logic, which uses three different API queries to get sections (accounts, courses, sections). @jonespm @pushyamig I'll work on drafting unit tests soon, but I'd appreciate any comments on the structure of the code  and the use of asyncio, in case I need to revise something in the logic beforehand. You'll notice some differences in async vs sync logic compared to code in [course_section_api_handler.py](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/develop/backend/ccm/canvas_api/course_section_api_handler.py)

From learning a little bit about the library, it seems that `asyncio.gather` is an older async approach that doesn't have the same control over concurrent tasks that the recently added TaskGroups may have. [Here's the article](https://www.dataleadsfuture.com/why-taskgroup-and-timeout-are-so-crucial-in-python-3-11-asyncio/) I referenced in our meeting today on this topic. It appears to me that TaskGroups may open up the opportunity for new enhancements (Semaphores, better cancellation & exception handling, etc), but I'm open to consider instead just sticking with the `gather` approach. 

If it looks good to you guys, we can create  a new issue to get the structure of course_section_api_handler to be written similarly with the use of TaskGroup instead.

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,189 @@
...
+        time_get_canvas_api_start: float = time.perf_counter()
+        canvas_api: Canvas = await asyncio.to_thread(self.credential_manager.get_canvasapi_instance, request)
+        time_get_canvas_api_end: float = time.perf_counter()
+        logger.info(f"TIME taken to get Canvas API instance: {time_get_canvas_api_end - time_get_canvas_api_start:.2f} seconds")
+        
+        logger.info(f"Retrieving admin sections data for term_id: {term_id}")
+        time_filter_accounts_start: float = time.perf_counter()
+        try:
+            # Get accounts accessible to the admin user
+            accounts_by_account_id = await self.retrieve_accounts_map(canvas_api)
```
pushyamig: The `retrieve_accounts_map` function is not part of the `CanvasAdminSectionsAPIHandler` , but calling it as self.retrieve_accounts_map() your intent seems to be using it as class function, this is unreachable code now.

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,189 @@
...
+        course_name = request.query_params.get('course_name', None)
+
+        if not term_id:
+            return Response({"error": "Term ID is required as a parameter"}, status=HTTPStatus.BAD_REQUEST)
+        
+        if not ((instructor_name != None) ^ (course_name != None)): # XOR condition ensures one, but not both, is provided
+            return Response({"error": "Provide either 'instructor_name' or 'course_name' as a parameter. Both cannot be provided together."},
+                            status=HTTPStatus.BAD_REQUEST)
+        time_get_canvas_api_start: float = time.perf_counter()
+        canvas_api: Canvas = await asyncio.to_thread(self.credential_manager.get_canvasapi_instance, request)
```
pushyamig: I wasn't sure why pulling Canvas() object that hold Canvas URL/Token needs to be called via `asyncio.to_thread`? was there a need for it? I feel it should work either way

backend/ccm/canvas_api/urls.py
```diff
@@ -1,11 +1,15 @@
...
 from backend.ccm.canvas_api.course_section_api_handler import CanvasCourseSectionAPIHandler
+from backend.ccm.canvas_api.instructor_sections_api_handler import CanvasInstructorSectionsAPIHandler
 from backend.ccm.canvas_api.section_enrollments_api_handler import CanvasSectionEnrollmentsAPIHandler
 
 urlpatterns = [
   path('course/<int:course_id>/', CanvasCourseAPIHandler.as_view() , name='course'),
   path('course/<int:course_id>/sections/', CanvasCourseSectionAPIHandler.as_view() , name='courseSection'),
   path('sections/students/', CanvasSectionEnrollmentsAPIHandler.as_view() , name='sectionEnrollments'),
+  path('admin/sections/', CanvasAdminSectionsAPIHandler.as_view(), name='adminSections'),
+  path('instructor/sections/', CanvasInstructorSectionsAPIHandler.as_view(), name='instructorSections'),
```
pushyamig: My observation is the admin/instructor for pulling the data to show the course section in different use cases, should have the Merge/cross-list API call added as part this issue?

backend/ccm/canvas_api/admin_sections_api_handler.py
pushyamig: I see the same pattern for rest of the function, there are not part of class but a standalone function, i am not saying this is good or bad, understand the purpose if you want them to be class functions or standalone.

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,189 @@
...
+        errors: list
+    ) -> list:
+    # First retrieve cour+ses for the account
+    courses = []
+    time_get_courses_start: float = time.perf_counter()
+    try:
+        def get_courses(courseQueryParams):
+            return account.get_courses(**courseQueryParams)
+        logger.info(f"Retrieving courses for account_id: {account.id}")
+        courses = await asyncio.to_thread(get_courses, courseQueryParams)
```
pushyamig: What is the need for the nested functional call?

 will it work if you extract the below to a separate function?  felt that could simply thing for readability, again i am understanding the purpose without my opinion 
 
```
 try:
        def get_courses(courseQueryParams):
            return account.get_courses(**courseQueryParams)
    except: 
        pass
```

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,189 @@
...
+            'state': ['created', 'claimed', 'available'],
+            'enrollment_term_id': term_id,
+            'per_page': 100,
+        }
+        if instructor_name:
+            courseQueryParams['by_teachers'] = ['sis_login_id:' + instructor_name]
+            logger.info(f"Searching for courses with instructor name: {instructor_name}")
+        if course_name:
+            courseQueryParams['search_term'] = course_name
+            logger.info(f"Searching for courses with name: {course_name}")
```
pushyamig: I feel L85-90 could be written before the accounts API call if the request by uniqname or course_name. That will be a better code flow IMHO.

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,189 @@
...
+            for account_id in filtered_account_ids:
+                account = accounts_by_account_id[account_id]
+                tg.create_task(
+                    fetch_sections_for_account_task(
+                        account,
+                        courseQueryParams,
+                        self.courses_allowed_fields,
+                        self.sections_allowed_fields,
+                        courses_with_sections_results, 
+                        courses_with_sections_errors)
```
pushyamig: I suppose the flow is such that if any point of time in pool of api calls if an exception happen, that it will fetch an incomplete info and will return error response to user?

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,189 @@
...
+            logger.info(f"Searching for courses with instructor name: {instructor_name}")
+        if course_name:
+            courseQueryParams['search_term'] = course_name
+            logger.info(f"Searching for courses with name: {course_name}")
+        
+        # Get sections from all courses for filtered accounts
+        courses_with_sections_results = []
+        courses_with_sections_errors = []
+        time_fetch_all_sections_start: float = time.perf_counter()
+        async with asyncio.TaskGroup() as tg:
```
pushyamig: can you say how TaskGroups is behaving in multiple API flow? 
1. First we have to get `User/accounts` 
2. accounts/courses - this is where the concurrency will help. If users is part of 2 account then 2 concurrent calls fetching all courses including Pagination. I guess UCF library going all that?
3. course/sections - similar workflow to step 2

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __init__(self, credential_manager=None):
+        self.credential_manager = credential_manager or CanvasCredentialManager()
+        self.canvas_error = CanvasErrorHandler()
+        super().__init__()
+    
+    def get(self, request: Request) -> Response:
+        return asyncio.run(self.get_async(request))
+
+    async def get_async(self, request: Request) -> Response:
```
pushyamig: I know this is still WIP, felt this function has big chunk of code, could be broken down for readability. I would say focus on this at the very last of this code review process

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: I rewrote this file. instead the internal functions are class functions only

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: This is rewritten and is no longer put on an asyncio thread

backend/ccm/canvas_api/urls.py
jaydonkrooss: I'll be working on cross-list and unmerge together, but as a [separate issue](https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/480) from this PR. As mentioned in our last meeting, we can test using the instructor sections flow.

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: I agree with this suggestion, moved courseQueryParams modifications to earlier, before any canvas api calls

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: After rewriting this, if any step (get accounts, get courses, get sections) fails, we return HTTPAPIError before moving on to the next step. Let me know if you suggest any changes with the way it's written

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: I'm no longer using TaskGroups with this code, I think the `asyncio.gather` pattern with semaphores and `return_exceptions` is sufficient

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: I've rewritten to try to organize the functions a little better. Instead of doing a nested call for sections, I have three different functions for each stage to build the response (`_get_accessible_accounts`,`_get_courses`, and `_attach_sections_to_courses`). The second and third functions use concurrency - courses for multiple accounts, and sections for multiple courses.

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: I rewrote this and removed it. the nested call I originally wrote is not really necessary

pushyamig:
I will review it today

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,204 @@
...
+from backend.ccm.canvas_api.canvas_credential_manager import CanvasCredentialManager
+from backend.ccm.canvas_api.constants import MAX_CONCURRENCY
+from backend.ccm.canvas_api.exceptions import CanvasErrorHandler, HTTPAPIError
+from backend.ccm.canvas_api.canvasapi_serializer import CanvasObjectROSerializer
+from backend.ccm.canvas_api.instructor_sections_api_handler import CanvasInstructorSectionsAPIHandler
+from backend.ccm.utils import timeit
+
+logger = logging.getLogger(__name__)
+
+class CanvasAdminSectionsAPIHandler(LoggingMixin, APIView):
```
pushyamig, line 25: Node currently handle if a use case when there are more than 400 courses then we deal with a error response. as it is noted [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/develop/ccm_web/server/src/api/api.admin.handler.ts#L38)

I think we are telling the user to refine the search with better search term I feel this need to be included, since this act as guardrail to the backend process, where we don't want to overwork. This call will take a while even with async imagine fetching all the sections for those 400 courses. If we just minic what is written in Node will be good.

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        term_id = request.query_params.get('term_id', None)
+        instructor_name = request.query_params.get('instructor_name', None)
+        course_name = request.query_params.get('course_name', None)
+        if not term_id:
+            return Response({"error": "Term ID is required as a parameter"}, status=HTTPStatus.BAD_REQUEST)
+        if not ((instructor_name != None) ^ (course_name != None)): # XOR condition ensures one, but not both, is provided
+            return Response({"error": "Provide either 'instructor_name' or 'course_name' as a parameter. Both cannot be provided together."},
+                            status=HTTPStatus.BAD_REQUEST)
+
```
pushyamig: This can be handled with Serializer. I know Merge section is what uses query param,,,, 

This is what Copilot suggested, did not test it since I reached Copilot Chat messages quota error :)

from rest_framework import serializers
```
#canvasapi_serializer
class AdminSectionsQuerySerializer(serializers.Serializer):
    terms_id = serializers.CharField(required=True)
    instructor_name = serializers.CharField(required=False, allow_null=True)
    course_name = serializers.CharField(required=False, allow_null=True)

    def validate(self, data):
        instructor = data.get('instructor_name')
        course = data.get('course_name')
        # XOR: Only one of instructor_name or course_name must be provided, not both or neither
        if not ((instructor is not None) ^ (course is not None)):
            raise serializers.ValidationError(
                "Provide either 'instructor_name' or 'course_name' as a parameter. Both cannot be provided together."
            )
        return data
```

     #CanvasAdminSectionsAPIHandler  
```
        def get(self, request: Request) -> Response:
    serializer = AdminSectionsQuerySerializer(data=request.query_params)
    if not serializer.is_valid():
        return Response(serializer.errors, status=HTTPStatus.BAD_REQUEST)
    validated_data = serializer.validated_data
    term_id = validated_data['terms_id']
    instructor_name = validated_data.get('instructor_name')
```
    course_name = validated_data.get('course_name')

pushyamig: The errors w.r.t params `term_id`, `instructor_name` and `course_name` is not presented in usual JSON format, then the FE will take as some kind of backend processing error will catch it handles it with generic response to user 

with this `return Response({"error": "Term ID is required as a parameter"}, status=HTTPStatus.BAD_REQUEST)` 
UI will respond error as 
Screenshot 2025-09-12 at 3 02 16 PM

so something like below with 
```
self.canvas_error.handle_serializer_errors('Term ID is required as a parameter', str(term_id))
 return Response(self.canvas_error.to_dict(), status=self.canvas_error.to_dict().get('statusCode'))
```
```
{
    "statusCode": 500,
    "errors": [
        {
            "canvasStatusCode": 500,
            "message": "Term ID is required as a parameter",
            "failedInput": "None"
        }
    ]
}

```
presented to user like this

Screenshot 2025-09-12 at 3 02 50 PM

These query_params validation errors are highly unlikely, but needs to have some structured support

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,204 @@
...
+                key for key, account in account_instance_map.items() if
+                    # account is the root account OR the subaccount of an unlisted account
+                    account.parent_account_id is None 
+                    or not(account.parent_account_id in account_instance_map)
+            ]
+            logger.info(f"Found {len(accessible_account_ids)} of {len(accounts)} accounts accessible to admin user")
+            return accessible_account_ids, account_instance_map
+        except (CanvasException, Exception) as e:
+            failed_input = f"user id {user_id}"
+            raise HTTPAPIError(failed_input, e)
```
pushyamig, line 138: As I read this the `user_id` is neither used in the `get_accounts()` api call, and seems to be for logging purpose. I don't see the purpose of sending this id to FE. since they won't know. This is id is `auth_user` primary key ID.  IMHO `failed_input` still in this case is `term_id` and ( course_name or instructor_uniqname)

with that said, I do see Node code uses it and it is ported here as well, which is ok. instead of `user. id` get the `user.username = uniqname` since that is something get populated during the LTI launch.

pushyamig:
The code with `get_Accounts | get_Courses | get_Section` search by Course_name and Section_name is working as expected. 

I have some 3 comments on error/validation if addressed this should be good to go

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: Agreed, I forgot to add better serializer and swagger support, so I edited to include for both instructors and admin sections. Unit tests are also impacted & fixed.

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: Replaced with username. I felt that the course search terms and `term_id` isn't actually impacting for this exception, so I decided any user info would be better to log & respond to FE as "failed_input" with. I agree that username is better in this case than id; to me it's just a hint to the user that search by account failed, not the actual course search.

jaydonkrooss:
@pushyamig I've made edits to address comments. let me know if anything else is needed!

backend/ccm/canvas_api/admin_sections_api_handler.py
pushyamig: Ok, Can you also add course_name or intructor_name to the failed input along with username? I agree that term_id won't make sense to user since it is behind the scene API params we prep

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -0,0 +1,234 @@
...
+        ],
+    )
+    def get(self, request: Request) -> Response:
+        """
+        Get sections data from Canvas for admins.
+        """
+        serializer = AdminSectionsQuerySerializer(data=request.query_params)
+        if not serializer.is_valid():
+            self.canvas_error.handle_serializer_errors(serializer.errors, f"{request.query_params}")
+            return Response(serializer.errors, status=self.canvas_error.to_dict().get('statusCode'))
```
pushyamig:
```suggestion
            return Response(self.canvas_error.to_dict(), status=self.canvas_error.to_dict().get('statusCode'))
```

pushyamig: Minor change, `self.canvas_error.to_dict()` this is what will capture the error in Error response format FE expected

backend/ccm/canvas_api/admin_sections_api_handler.py
pushyamig: i fell this still need to be handled near L105 `courses_success, courses_response = self._get_courses(coursesQueryParams, course_instance_map, accessible_account_ids, account_instance_map)`.  This is how Node is doing it as it is noted [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/develop/ccm_web/server/src/api/api.admin.handler.ts#L38)

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: done.

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: I forgot to address this earlier, thanks for the reminder. I added the maximum as a constant and created a new unit test for the Max error logic

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: done. I also added this to the instructor_sections serializer error as well

pushyamig: Currently 
```

            self.canvas_error.handle_serializer_errors(self.canvas_error.to_dict(), f"{request.query_params}")
            return Response(serializer.errors, status=self.canvas_error.to_dict().get('statusCode'))
```
Needs to be changed to
```
self.canvas_error.handle_serializer_errors(serializer.errors, f"{request.query_params}")
return Response(self.canvas_error.to_dict(), status=self.canvas_error.to_dict().get('statusCode'))
```

I think the order of `serializer.errors` and `self.canvas_error.to_dict()` is bit mixed up

backend/ccm/canvas_api/instructor_sections_api_handler.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, credential_manager=None):
     def get(self, request: Request) -> Response:
         serializer = InstructorSectionsQuerySerializer(data=request.query_params)
         if not serializer.is_valid():
-            self.canvas_error.handle_serializer_errors(serializer.errors, f"{request.query_params}")
```
pushyamig: Same comment applies here https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/527/commits/1fb50545190423b55cc0fa54b125a19bce36d6a0#r2360336425

backend/ccm/canvas_api/admin_sections_api_handler.py
```diff
@@ -174,6 +174,12 @@ def _get_courses_by_account_sync(
         """ Synchronous helper to fetch and append courses from a given account. """
         try:
             account_courses = list(account.get_courses(**coursesQueryParams))
+            # Number of courses cannot exceed maxiumum
+            if len(account_courses) > MAX_SEARCH_COURSES:
+                relevantParams = {'by_teachers': coursesQueryParams.get('by_teachers'), 'search_term': coursesQueryParams.get('search_term')}
+                logger.error(f"Query with the following search term(s) returned too many course results: {relevantParams}")
+                raise Exception(f"Course search exceeded maximum of {MAX_SEARCH_COURSES} courses in account. Please refine your search.")
+
```
pushyamig, line 182: Nice this is working beautifully!

backend/ccm/canvas_api/admin_sections_api_handler.py
jaydonkrooss: Got it, that's a mistake on my part. I fixed it and adjusted tests for both admin and instructor.

backend/ccm/canvas_api/instructor_sections_api_handler.py
jaydonkrooss: resolved, and tests are adjusted.

pushyamig (approved):

Issue #488, opened 2024-11-01
Create Test openshift project


pushyamig:
CCM test ---> Canvas Test

zqian:
Tested the CCM Dev tool on Canvas Test

PR #528, opened 2025-06-19
setting up the CI/CD pipeline for Build and running test

### LINKED ISSUE #487: Add the Github Build Process for Python Build
Since We be doing all the Development of Python Rewrite in the Develop branch of the main repo, we need to have a github build process to support but Python and Node. 

This would mean adding the Develop branch to the github workflow.

pushyamig:
Fixes #487 #488 

1. The Build is running in Openshift and the process is document [here],(https://docs.google.com/document/d/17M3qrZlu0bS4hHdzWv6GXQwkXFk_54E5ry8I77IJBvg/edit?tab=t.0). check out the ccm-dev project for looking at the setup
2. The Django test should run by opening the PR here
Screenshot 2025-06-19 at 10 12 26 AM
4. I have change the CCM port to **4001**, this will be a breaking change so need to pay attention

Do a local build that things startup, run fines


pushyamig:
I removed the Coverage Comment step to see what is covered, since that is not possible if this branch that is opening the PR is coming from Fork vs branch from the PR. ChatGPT suggest PAT but not recommended for public repo

More [here](https://chatgpt.com/share/68541937-2bc0-8013-bd5c-c507ffefe27d) 

I have coverage report in the GH action run, but can't be shown as a comment as it did in [my CCM repo](https://github.com/pushyamig/canvas-course-manager-next/pull/3#issuecomment-2957052112)

supervisor_docker.conf
```diff
@@ -24,8 +40,9 @@ startretries=2
 stopsignal=QUIT
 
 [program:qworker]
-command=sh -c "while ! nc -z ccm_db 3306; do echo 'Waiting for MySQL...'; sleep 1; done; watchfiles --filter python 'python manage.py qcluster' /code/backend"
+command=sh -c "while [ ! -f /tmp/backend_ready ]; do echo 'Waiting for backend to be ready...'; sleep 2; done; if [ \"$RUN_QWORKER_DEV_MODE\" = \"true\" ]; then echo 'Running in qworker DEV mode'; watchfiles --filter python 'python manage.py qcluster' /code/backend; else echo 'Running in qworker PROD mode'; python manage.py qcluster; fi"
 directory=/code
```
pushyamig, line 44: I am starting up this workers once Backend is fulling stared and has a Local and prod mode where the watchfile is not triggered.

start.sh
```diff
@@ -62,6 +65,9 @@ else
     GUNICORN_RELOAD="--reload"
     GUNICORN_TIMEOUT=0
 fi
+# Signal backend is ready for qworker
+( echo "Backend finished starting up." && touch /tmp/backend_ready ) &
```
pushyamig, line 69: I’m ensuring that the file /tmp/backend_ready is created to control when the qworker process starts.

I previously encountered an error during the initial database migration where qworker attempted to start before the migration completed. This caused a failure because the required qworker_orm table did not yet exist.

This issue only occurs on the first run when the database tables haven’t been created yet. Creating this file acts as a signal that the backend is ready, preventing qworker from starting prematurely. This is the solution I came up with to address the problem.

start.sh
```diff
@@ -10,7 +13,7 @@ if [ -z "${GUNICORN_WORKERS}" ]; then
 fi
 
 if [ -z "${GUNICORN_PORT}" ]; then
-    GUNICORN_PORT=4000
+    GUNICORN_PORT=4001
```
pushyamig, line 16: changed the port since node and python version is served up from same Openshift project, felt changing it might be easier

backend/tests/test_views.py
```diff
@@ -39,8 +41,9 @@ def test_redirect_oauth_view_raises_missing_token_error(self):
         scopes_encoded = '+'.join([scope.replace('|', '%7C').replace('/', '%2F').replace(':', '%3A') for scope in DEFAUlT_CANVAS_SCOPES])
         self.assertIn(f'scope={scopes_encoded}', url_redirect)
 
+    @mock.patch('webpack_loader.loader.WebpackLoader.get_bundle', return_value={})
     @patch('canvas_oauth.oauth.refresh_oauth_token')
-    def test_canvas_token_exists_refreshed_if_called(self, mock_refresh_oauth_token):
+    def test_canvas_token_exists_refreshed_if_called(self, mock_refresh_oauth_token, mock_get_bundle):
         expires = timezone.make_aware(timezone.datetime(2023, 12, 31, 23, 59, 59))
         CanvasOAuth2Token.objects.create(
             user=self.user, access_token='access-token', expires=expires, refresh_token='refresh-token'
```
pushyamig, line 49: This tests are changed in order to fit the running the test with GH action.

supervisor_docker.conf
```diff
@@ -1,5 +1,20 @@
+[unix_http_server]
+file=/tmp/supervisor.sock
+chmod=0700
+
+[supervisorctl]
+serverurl=unix:///tmp/supervisor.sock
+
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
```
pushyamig, line 10: I added this so that i can see what process supervisor running. Open for changing it if it is redundant

`supervisorctl -c /code/supervisor_docker.conf status`

.github/workflows/ccm.yml
```diff
@@ -23,7 +23,7 @@ jobs:
   build:
     # to test a feature, change the repo name to your github id
     if: github.repository_owner == 'tl-its-umich-edu' || github.event_name == 'workflow_dispatch'
-    runs-on: ubuntu-latest
+    runs-on:  'ubuntu-latest'
     steps:
```
pushyamig, line 27: I did not add the ITS runners after the slack comment, I am open to adding it back if we have process what we agree on

jonespm: That's fine, we can't use them yet on public repos anyway.

.github/workflows/test.yml
```diff
@@ -0,0 +1,82 @@
...
+    branches:
+      - main
+      - develop
+  workflow_dispatch:
+
+
+jobs:
+  test:
+    runs-on:  'ubuntu-latest' 
+    services:
```
jonespm, line 14: I think my only comment on this is we may want to just look into starting up docker-compose and docker directly to keep these in sync and not redefine them here. I'm fine going with this as-is though and not making that change now. 

Then everything would be run by steps in docker-compose such as build, applying migrations, running the tests. This was a AI suggestion when I asked about this. This would need some work to do but be much easier to keep in sync as we make updates and changes 

```
# .github/workflows/ci-docker-compose.yml
name: CI with Docker Compose

on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main

jobs:
  build-test-compose:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout code
        uses: actions/checkout@v4

      - name: Build and start services with Docker Compose
        # Use --build to ensure images are built from Dockerfiles if defined in compose
        # Use -d for detached mode
        run: docker-compose -f docker-compose.yml up -d --build

      - name: Wait for services to be healthy
        # Implement a robust wait strategy for all critical services
        # For MySQL:
        run: |
          echo "Waiting for MySQL to be ready..."
          # Example: Using a shell loop with docker-compose exec
          for i in $(seq 1 60); do # Wait up to 5 minutes (60 * 5s)
            docker-compose exec mysql mysqladmin ping -h 127.0.0.1 -u root -p"${{ secrets.MYSQL_ROOT_PASSWORD }}" --silent 2>/dev/null
            if [ $? -eq 0 ]; then
              echo "MySQL is up and running!"
              break
            fi
            echo "Waiting for MySQL... ($i/60)"
            sleep 5
            if [ $i -eq 60 ]; then
              echo "MySQL did not become ready in time."
              exit 1
            fi
          done
          # Add similar waits for other services if needed

      - name: Apply database migrations
        # Execute migrations, potentially using a service from docker-compose
        run: docker-compose exec my-app-service npm run migrate # Adjust 'my-app-service' to your app's service name in docker-compose.yml

      - name: Run integration tests
        # Execute tests, connecting to services via their names (e.g., 'mysql', 'my-app-service')
        # You might run tests from your application container, or a dedicated test runner container
        run: docker-compose exec my-app-service npm run integration-test # Adjust service name and test command

      # Optional: Steps to push your application's Docker image if built here
      # ... (same as previous example)

      - name: Stop services and clean up
        if: always() # Ensure this runs even if previous steps fail
        run: docker-compose -f docker-compose.yml down --rmi all --volumes
```

supervisor_docker.conf
jonespm: It may be cleaner to move this command to another start script like start_worker.sh so we don't have this complex logic in here. But that's more of a nice-to-have.

jonespm (approved):
I left a few comments, those are fine if new issues were just created for some of these rather than fixing this now. I think there will likely be some further improvements here in the future.

jaydonkrooss:
As written, these changes look good to me (local code runs, coverage test logs look good). 

I agree with Matt's suggestions for organization, which could easily be separate issues to address later. I think starting the test instance for testing would be good to prioritize.

jaydonkrooss (approved):

.github/workflows/test.yml
pushyamig: I will work this as separate issue

supervisor_docker.conf
pushyamig: create Separate issue

PR #544, opened 2025-07-11
#542 switching to asgi worker

### LINKED ISSUE #542: Support async (ASGI) in app

-- | -- | -- | --

pushyamig:
Fixes #542


jonespm (approved):
Tested this out locally and switched to servestatic, looks like it's all working well.

pushyamig:
Thanks @jonespm for looking into it!

PR #546, opened 2025-07-14
feat: optimize course section api handlers to reduce canvas api calls

### LINKED ISSUE #524: Evaluate where mocking canvasAPI classes could improve performance
In the PR for implementing the third-party gradebook feature #482 (PR #518), I modified how we called the UCF Canvas API so that only one true endpoint was called. This was done by creating a dummy Section object with the given section ID's instead of the documented approach of retrieving the sections via API. This approach simplifies the code and improves performance slightly.

The same approach could also be applied to any other endpoints where we are retrieving "middleman" Canvas objects for the explicit purpose of calling the class's methods to retrieve the actual requested resource. For example, GET course section has to retrieve the course object **before** calling `course.get_sections`, but this step could be eliminated if needed. 

As of now, this is mostly a "nice-to-have" improvement: evaluate where we may want to apply this method to simplify & save on performance, and properly mock the respective unit tests.

PraveshKunwar:
Summary: resolves #524.

### DIFF
--- a/backend/ccm/canvas_api/course_section_api_handler.py
+++ b/backend/ccm/canvas_api/course_section_api_handler.py
@@ -43,6 +43,8 @@ class CanvasCourseSectionAPIHandler(LoggingMixin, APIView):
             # Call the Canvas API package to get section details.
         try:
             logger.info(f"Retrieving sections for course_id: {course_id}")
+            # Create a course object with just the ID to avoid unnecessary API call
+            # Skips the get_course() call and directly uses get_sections()
             course = Course(canvas_api._Canvas__requester, {'id': course_id})
             # Get list of sections, including total_students info
             sections = course.get_sections(include=['total_students'], per_page=per_page)
@@ -72,6 +74,8 @@ class CanvasCourseSectionAPIHandler(LoggingMixin, APIView):
         logger.info(f"Creating {sections} sections for course_id: {course_id}")
         canvas_api: Canvas = self.credential_manager.get_canvasapi_instance(request)
         
+        # Create a Course object with just the ID to avoid unnecessary API call
+        # If the course doesn't exist, the section creation will fail with proper error
         course = Course(canvas_api._Canvas__requester, {'id': course_id})
            
         start_time: float = time.perf_counter()
--- a/backend/ccm/canvas_api/course_section_api_handler.py
+++ b/backend/ccm/canvas_api/course_section_api_handler.py
@@ -42,9 +42,10 @@ class CanvasCourseSectionAPIHandler(LoggingMixin, APIView):
             
             # Call the Canvas API package to get section details.
         try:
-            logger.info(f"Retrieving course for section data with course_id: {course_id}")
+            logger.info(f"Retrieving sections for course_id: {course_id}")
+            mock_course = Course(canvas_api._Canvas__requester, {'id': course_id})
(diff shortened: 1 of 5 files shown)


backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -42,9 +42,10 @@ def get(self, request: Request, course_id: int, per_page: int = 100) -> Response
             
             # Call the Canvas API package to get section details.
         try:
-            logger.info(f"Retrieving course for section data with course_id: {course_id}")
+            logger.info(f"Retrieving sections for course_id: {course_id}")
+            mock_course = Course(canvas_api._Canvas__requester, {'id': course_id})
```
pushyamig: I don't feel `mock_course` a good variable name, I still feel `course` is ok to name, add a Comment here why we are adopted this approach

pushyamig:
In `course_api_handler.py` for the PUT call, Below call needs to be mocked, since the PUT call is for updating the Course Name and for that we can mock the below call as you have done in the PR. This is another place a change is needed
`course: Course = canvas_api.get_course(course_id)`

backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -42,9 +42,12 @@ def get(self, request: Request, course_id: int, per_page: int = 100) -> Response
             
             # Call the Canvas API package to get section details.
         try:
-            logger.info(f"Retrieving course for section data with course_id: {course_id}")
+            logger.info(f"Retrieving sections for course_id: {course_id}")
+            # Create a course object with just the ID to avoid unnecessary API call
+            # Skips the get_course() call and directly uses get_sections()
+            course = Course(canvas_api._Canvas__requester, {'id': course_id})
```
pushyamig, line 48: Can you hook up the debugger here are see if it will step out smoothly? Once I noticed it was failing

PraveshKunwar: I'm not sure why, my debugger isn't working when I go through the configuration

pushyamig: There are 2 options for debugging you should pick the top one. DM me if you can't figure it out
Screenshot 2025-07-15 at 8 10 04 AM

PraveshKunwar: It is stepping out for me I believe

pushyamig:
Tests are failing, need to look into passing the tests with updated code

PraveshKunwar:
> Tests are failing, need to look into passing the tests with updated code

I believe the tests are passing now

pushyamig:
>I believe the tests are passing now
Can you tell what was the reason for failure of the test earlier?

PraveshKunwar:
>  I believe the tests are passing now
>  Can you tell what was the reason for failure of the test earlier?

it was failing because after we optimized the API calls, the test mocks were still using the old API pattern, so we had to update the mocks to work with the new course constructor approach instead of the `get_course()` method.

pushyamig (approved):
This work needed understanding of canvasapi class structure and how to bypass API call and only going instant object and fixing the failing the tests with this change.  
I hope this gave a perspective on what we are trying to do since this is not the way the pkg developer asked to do but our needs demanded a change. Good Work!

Issue #536, opened 2025-06-30
Move the worker starting up to a script


pushyamig:
More details https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/528/files#r2170004064

pushyamig:
This is done but not closed. Currently the worker script is https://github.com/tl-its-umich-edu/canvas-course-manager-next/tree/develop/shell_scripts

PR #551, opened 2025-07-18
#542 Development support with asgi

### LINKED ISSUE #542: Support async (ASGI) in app

pushyamig:
Fixes #542 #536

There is a bug when shifting to Asgi from wsgi that the `--reload` was not happening. We are now using new Uvicorn worker, and the author [here](https://github.com/Kludex/uvicorn-worker?tab=readme-ov-file) suggested the to use Pure ASGI using UVICORN for development and a mix of GUNICON with Uvicorn pulling the best of both worlds of ASGI and WSGI. 

I looked up ROHQ after witnessing this bug, they followed same approach.

### DIFF
--- a/shell_scripts/worker.sh
+++ b/shell_scripts/worker.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+echo "qworker is starting..."
+
+# Wait for backend to be ready since supervisor will spawn all the 3 processes backend, qworker, and frontend same time
+while [ ! -f /tmp/backend_ready ]; do
+    echo 'qworker: Waiting for backend to be ready...'
+    sleep 2
+done
+
+# this is to ensure that the backend/DB is fully ready before starting the qworker
+echo "qworker: Backend is ready, starting qworker..."
+if [ "$RUN_QWORKER_DEV_MODE" = "true" ]; then
+    echo 'qworker: Running in DEV mode'
+    rm /tmp/backend_ready
+    watchfiles --filter python 'python manage.py qcluster' /code/backend/ccm/background_tasks
+else
+    echo 'qworker: Running in PROD mode'
+    python manage.py qcluster
+fi
+
--- a/start.sh
+++ b/start.sh
@@ -48,8 +48,6 @@ done
 echo Running python migrations
 python manage.py migrate
 
-echo "Setting domain of default site record"
-echo "DEBUGPY_ENABLE: $DEBUGPY_ENABLE"
 
 if [ "${DEBUGPY_ENABLE:-"false"}" == "false" ]; then
     echo "Starting Gunicorn with uvicorn worker for production"
--- a/start.sh
+++ b/start.sh
@@ -34,12 +34,6 @@ if [ -z "${DJANGO_SECRET_KEY}" ]; then
     echo "DJANGO_SECRET_KEY not set, using random value"
 fi
 
-if [ "${GUNICORN_RELOAD}" ]; then
-    GUNICORN_RELOAD="--reload"
-else
-    GUNICORN_RELOAD=
-fi
-
 # To have a more static default secret key, this should still be defined
 if [ -z "${DJANGO_SECRET_KEY}" ]; then
     export DJANGO_SECRET_KEY=`python -c 'from django.core.management.utils import get_random_secret_key; print(get_random_secret_key())'`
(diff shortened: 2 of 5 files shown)


start.sh
```diff
@@ -54,23 +48,15 @@ done
...
+    echo "Starting Gunicorn with uvicorn worker for production"
+    CMD="gunicorn backend.asgi:application --bind 0.0.0.0:${GUNICORN_PORT} --workers=\"${GUNICORN_WORKERS}\" -k uvicorn_worker.UvicornWorker --timeout=\"${GUNICORN_TIMEOUT}\" "
 else
-    echo "Starting Gunicorn for DEBUGPY debugging"
-    # Workers need to be set to 1 for DEBUGPY
-    GUNICORN_WORKERS=1
-    GUNICORN_RELOAD="--reload"
-    GUNICORN_TIMEOUT=0
+    echo "Starting uvicorn for Development"
+    CMD="uvicorn backend.asgi:application --host=0.0.0.0 --port=${GUNICORN_PORT} --reload"
```
pushyamig: Here is the author recommendation https://github.com/Kludex/uvicorn-worker?tab=readme-ov-file#development

supervisor_docker.conf
```diff
@@ -1,14 +1,3 @@
...
-file=/tmp/supervisor.sock
-chmod=0700
-
-[supervisorctl]
-serverurl=unix:///tmp/supervisor.sock
-
-[rpcinterface:supervisor]
-supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
-
-
```
pushyamig, line 11: This is just for debugging purpose to see the processes in the terminal, felt not needed.

supervisor_docker.conf
```diff
@@ -40,7 +29,7 @@ startretries=2
 stopsignal=QUIT
 
 [program:qworker]
-command=sh -c "while [ ! -f /tmp/backend_ready ]; do echo 'Waiting for backend to be ready...'; sleep 2; done; if [ \"$RUN_QWORKER_DEV_MODE\" = \"true\" ]; then echo 'Running in qworker DEV mode'; watchfiles --filter python 'python manage.py qcluster' /code/backend; else echo 'Running in qworker PROD mode'; python manage.py qcluster; fi"
```
pushyamig, line 43: The background task process setup all file are being located in the `background_tasks` folder. so felt if the worker just reload/watches that folder would be sufficient rather than both Unicorn and QWorkers doing it. This is just for development, since `watchfiles` package only used for reloading the development

jaydonkrooss:
Code looks good to me, startup + background tasks run fine locally. separating the shell code into their own script files makes sense as well.

jaydonkrooss (approved):

shell_scripts/worker.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+echo "qworker is starting..."
+
+# Wait for backend to be ready since supervisor will spawn all the 3 processes backend, qworker, and frontend same time
+while [ ! -f /tmp/backend_ready ]; do
```
jonespm, line 6: I'd think if this needs to wait, it could use the same pattern to check for the database and just look for localhost and the port, then it wouldn't need a temp file?

We could leave this and think about it more as we go toward production, whether this worker would run in the same pod, a separate pod or as a scheduled task. 

```
if [ -z "${GUNICORN_PORT}" ]; then
    GUNICORN_PORT=4001
fi 
```

jonespm (approved):
I tested and it also works for me too. I left a comment about the worker startup that we may want to review but I'm fine with it going in as-is with this.

shell_scripts/worker.sh
pushyamig: The Reason why this is there is because during initial setting up...
1. qworker need access to its DB tables so migrations needs to happen
2. If Backend is starting up and so is qworkers and migration are not done then you will see `django_q_ormq` table is not available. 
3. so it's not just DB availability but running through all migration. Initially that's what I was doing checking DB host/ports are available or not, but that check was not enough.

pushyamig: >We could leave this and think about it more as we go toward production, whether this worker would run in the same pod, a separate pod or as a scheduled task.

I am open to this idea during prod

PR #552, opened 2025-07-22
Making it DJango async style

### LINKED ISSUE #542: Support async (ASGI) in app

pushyamig:
Fixes #542 #152 

Alternative to asyncio.run() for starting the async flow in Django. I feel with this approach we don't have worry about performance what will happen when concurrent users start hitting us and behavior of running the async workflow.  

- https://youtu.be/B5uQPwX4VLo?si=qcmm3mZiD-B8hWP6. (This provide a good overview of Django Async changes needed)
- https://docs.djangoproject.com/en/5.2/topics/async/  (Django Async support)
- https://docs.djangoproject.com/en/5.2/topics/async/#async-adapter-functions 

I think world in Python/Django is shifting towards supporting async development. But the ecosystem is still not fully caught up. `Django ORM` is one use case of stepping towards async support.`requests` supports only sync, but libraries like `aiohttp` and `httpx` has support for async and support the `asyncio`. `asyncio` module is default async support in Python is well supported by most of the major packages and is a much mature state. 

`FastAPI` and `Sanic` are emerging asynchronous web framework. 

I guess async is evolving topic and something we started to adopt  (like adding Uvicorn addition ), learn and grow. I guess my learning so far is that there is no playbook async web development since the echo system is moving towards it and we just need to pick and chose what best work for our project.

### DIFF
--- a/backend/ccm/canvas_api/course_section_api_handler.py
+++ b/backend/ccm/canvas_api/course_section_api_handler.py
@@ -5,12 +5,12 @@ from rest_framework.views import APIView
 from rest_framework import authentication, permissions
 from rest_framework.response import Response
 from rest_framework.request import Request
-from asgiref.sync import sync_to_async
 
 from canvasapi.exceptions import CanvasException
 from canvasapi import Canvas
 from canvasapi.course import Course
 from drf_spectacular.utils import extend_schema
+from asgiref.sync import async_to_sync
 
 from .exceptions import CanvasErrorHandler, HTTPAPIError
 
@@ -80,7 +80,6 @@ class CanvasCourseSectionAPIHandler(LoggingMixin, APIView):
         course = Course(canvas_api._Canvas__requester, {'id': course_id})
            
         start_time: float = time.perf_counter()
-        from asgiref.sync import async_to_sync
         results = async_to_sync(self.create_sections)(course, sections)
         end_time: float = time.perf_counter()
         logger.info(f"Time taken to create {len(sections)} sections: {end_time - start_time:.2f} seconds")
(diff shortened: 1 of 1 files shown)


backend/ccm/canvas_api/course_section_api_handler.py
```diff
@@ -96,10 +97,17 @@ def post(self, request: Request, course_id: int) -> Response:
...
+    async def sem_task(self, semaphore, course, name):
+        async with semaphore:
+            return await self.create_section(course, name)
+
+    @async_to_sync
     async def create_sections(self, course: Course, section_names: list):
-        """Creates multiple sections concurrently."""
-        tasks = [self.create_section(course, name) for name in section_names]
+        """Creates multiple sections concurrently, guarded by a semaphore."""
+        max_concurrent = 10  # Set your desired concurrency limit here
```
jonespm, line 108: Should this be a configurable setting or is 10 a likely constant value? Maybe it could go into the new constants,py in #560 if it is not meant to be configurable.

jonespm (approved):
Aside from my question about the configurability this looks good to me.

backend/ccm/canvas_api/course_section_api_handler.py
pushyamig: OK, I can make this change, but `constants.py` file is not added here. Maybe Once this merged I will make the change it #560?

I think so far 10 working fine for create sections and enroll users, so i will keep it that way.  In in future if that changes, I will make a change to the particular usecase

jonespm: Yes that's fine. I'd merge #560 first, then update the branch here and update that file.

pushyamig:
I am merging it now. I will resolve the comment by @jonespm in Enroll user PR.

PR #556, opened 2025-07-28
feat: fixed migration warnings for canvas_oauth and rest_framework_tracking

### LINKED ISSUE #504: Fix the warning from model migration
For some reason even after models from `canvas_oauth` and `rest_framework_tracking` we are seeing this warning message, although it doesn't effect the functionality but annoyance in log message
 
`Your models in app(s): 'canvas_oauth', 'rest_framework_tracking' have changes that are not yet reflected in a migration, and so won't be applied. with this PR`

This is was something that was fixed in the instructor Tools Project and similar approach can be taken here as well. https://github.com/tl-its-umich-edu/instructor-tools/pull/373

PraveshKunwar:
Summary: resolves #504.
Problem was whenever running  `docker compose up` we encountered this warning: `Your models in app(s): 'canvas_oauth', 'rest_framework_tracking' have changes that are not yet reflected in a migration, and so won't be applied.` This workaround should fix it, based off this PR: https://github.com/tl-its-umich-edu/instructor-tools/pull/373


backend/rest_framework_tracking/migrations/0001_initial.py
```diff
@@ -0,0 +1,42 @@
...
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='APIRequestLog',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
```
pushyamig, line 19: I see that `id` is [AutoField in the drf model ](https://github.com/aschn/drf-tracking/blob/master/rest_framework_tracking/migrations/0001_initial.py#L22)and the migration file created is using `BigAutoField`.  I did not see in the further migration files the `id` Changed to `BigAutoField`

Does my question is relevant here?

pushyamig:
This change is working and I saw the warning message gone.

pushyamig:
@jonespm you will be final approver on this.

backend/rest_framework_tracking/migrations/0001_initial.py
pushyamig: In the [setting.py](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/develop/backend/settings.py#L177) I see this configuration `DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'` and probably this migration file is reflecting that.  This behavior is same for CanvasOauth and drf_tracking library . @jonespm is this ok?

@PraveshKunwar How did you generate this Migration file?

PraveshKunwar: I ran this command: `docker compose exec web python manage.py makemigrations rest_framework_tracking`, and it generated that file with `BigAutoField`

pushyamig: Ok, this is the deal with this change..... 
1. The warning will be gone for sure
2. If you have migration ran before then the `describe canvas_oauth_canvasoauth2token` will say `id = int`, but If this is a new migration them `id = bigint`
3. So Probably the prod will be `bigint` if this change goes in
4. https://docs.djangoproject.com/en/5.2/ref/settings/#default-auto-field
DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
This something was added by Django as default with newer Django version after 3.2

```
Migrating auto-created through tables

The value of DEFAULT_AUTO_FIELD will be respected when creating new auto-created through tables for many-to-many relationships.

Unfortunately, the primary keys of existing auto-created through tables cannot currently be updated by the migrations framework.

This means that if you switch the value of DEFAULT_AUTO_FIELD and then generate migrations, the primary keys of the related models will be updated, as will the foreign keys from the through table, but the primary key of the auto-created through table will not be migrated.

In order to address this, you should add a [RunSQL](https://docs.djangoproject.com/en/5.2/ref/migration-operations/#django.db.migrations.operations.RunSQL) operation to your migrations to perform the required ALTER TABLE step. You can check the existing table name through sqlmigrate, dbshell, or with the field’s remote_field.through._meta.db_table property.

Explicitly defined through models are already handled by the migrations system.

Allowing automatic migrations for the primary key of existing auto-created through tables [may be implemented at a later date](https://code.djangoproject.com/ticket/32674).
```
5. I think while drf_tracking and canvasOuth migration are still using AutoInt and the migration files we created use BigInt, at some point we remove these migration file and unless they change BigInt we are in a flux state
6. While `int` = 2,147,483,647 and `bigInt`= 9,223,372,036,854,775,807.  don't feel we will ever generate billon records with drf.

Over all I feel this change of migration file not similar to Package version. risk might be low, but I want to bring out the nuance here

pushyamig (approved):

backend/rest_framework_tracking/migrations/0001_initial.py
jonespm: I think since BigAutoField is explicitly set that it's appropriate to have here.

jonespm (approved):

Issue #522, opened 2025-05-05
Send email to the user once the Background task is done


pushyamig:
When you run the enroll user as background task, the user needs to be informed if success or failure. This is something new with Python Rewrite and is not present with Node app.

We will be sending the email from ITS Authenticated SMPT server More details in this [ticket](https://teamdynamix.umich.edu/TDNext/Apps/31/Tickets/TicketDet.aspx?TicketID=8959449)

https://documentation.its.umich.edu/authenticated-smtp?check_logged_in=1

pushyamig:
These are email message user will receive when success/failure to enroll uses 
**Success Email:** Simple says all success i Subject/body. No attachment is served. 

```
Content-Type: text/plain; charset="utf-8"
 MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: For course 483458, 7/7 enrollments finished successfully
From: 4HELP@UMICH.EDU
To: pushyami@umich.edu
Cc: CCM.ADMIN@UMICH.EDU
Date: Thu, 14 Aug 2025 17:37:49 -0000
Message-ID: <175519306972.51.13140393853026063744@cb5c04085b90>

For Course 483458 enrolling all users is success
```

**Faillure Email structure**: The below test case when all loginID(uniqname) are incorrect, so it is returning the reason for failure which is coming from Canvas(or source of error). In this example returns error message as `Not found`

```
Content-Type: multipart/mixed; boundary="===============7184612628964569362=="
 MIME-Version: 1.0
 Subject: For course 483458, 0/7 enrollments finished successfully (7 failed)
 From: 4HELP@UMICH.EDU
To: pushyami@umich.edu
Cc: CCM.ADMIN@UMICH.EDU
Date: Thu, 14 Aug 2025 17:39:41 -0000
Message-ID: <175519318165.52.12924321394095749538@cb5c04085b90>

--===============7184612628964569362==
 Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit

For Course 483458 enrolling users encountered some failures. See attachment for error list.
 --===============7184612628964569362==
Content-Type: text/csv; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
 Content-Disposition: attachment; filename="483458.csv"

sectionId,LoginId,role,ReasonForFailure
 878977,stud1,student,Not Found
 878977,stud2,teacher,Not Found
[...]

PR #560, opened 2025-08-13
I523 enroll user to course sections

### LINKED ISSUE #523: Enrolling user with Canvas api

pushyamig:
Fixes #523 

This PR only address Enrolling users to section both single and multiple section mode. UI changes are made just to fit running it as background Process using [Django-q2](https://django-q2.readthedocs.io/en/master/architecture.html) Like pulling the Course ID to serve for email sending purpose ( Email send will be handled as #522 ). We support Following roles through CCM:
```
ALLOWED_ROLES = (
    "student",
    "teacher",
    "ta",
    "observer",
    "designer",
    "assistant", (custom)
    "librarian",(custom)
)
```

Enroll users upto 5000 Some stats of the process running
```
1000 Users took 53 sec 
1767 Users took 1:20 min 
5000 Users took 3:43 min 
```
**Canvas OAuth**: Adding a [Canvas Token buffer](https://github.com/Harvard-University-iCommons/django-canvas-oauth?tab=readme-ov-file#settings) to renew 15min before expiry to support the background process , keeping the error of token expiry to minimum. The Longest process will be finished under 5min

**Canvas API**: The enroll user function doesn't support alternative id like `sis_login_id` to be used as `enrollment[user_id]": f"sis_login_id:{process_login_id(login_id)}",`, so we don't use that built-in enroll function as a work around we created our own enroll_user() still playing with in the limits of canvas api. Please checkout `enroll_users.py` about implementation of this.

### Config Changes and Testing resources:

1. [Testing Sample files](https://drive.google.com/drive/folders/1366DsbAvRs371nBKXf8HiRSlLNFBd0dF)
2. **Config Changes**: all are optional but defaults values are set
```
CANVAS_OAUTH_TOKEN_EXPIRATION_BUFFER=15
CUSTOM_CANVAS_ROLES='{"assistant": 34, "librarian": 21}' 
Q_CLUSTER_WORKERS=4
Q_CLUSTER_TIMEOUT=1800
Q_CLUSTER_RETRY=3600
Q_CLUSTER_BULK=5
Q_CLUSTER_MAX_ATTEMPTS=1
```
3. **Canvas Scopes changes**: Add Below scope to `Canvas Developer API Key`, since this is background process, intially you might delete the token in the DB. if you run the enroll process with out it, code handles it by deleting it. But the whole User authorization process will be triggered with next LTI launch. **Note: This is not a prod issue since this scope will be enabled when we go live with Python version of CCM**
 `url:POST|/api/v1/sections/:section_id/enrollments`. 
4. Full support to Unit Test cases with enroll users usecases


backend/ccm/canvas_api/constants.py
```diff
@@ -0,0 +1,20 @@
...
+    "teacher",
+    "ta",
+    "observer",
+    "designer",
+    "assistant",
+    "librarian",
+)
+
+# Maximum number of enrollments allowed in a single section enrollment request
+MAX_ALLOWED_ENROLLMENTS = 5000
```
jonespm, line 14: Do you think it would be useful to have this default in settings instead and make it configurable? I see some code to patch it to a smaller value for testing, so maybe it would be better there than in constants.

backend/ccm/canvas_api/enroll_users.py
```diff
@@ -0,0 +1,57 @@
...
+from django.conf import settings
+from .constants import ROLE_TO_ENROLLMENT_TYPE
+logger = logging.getLogger(__name__)
+
+def process_login_id(login_id: str) -> str:
+    """
+    Process the login ID to strip the domain and handle special cases. Handles both UMich and Non-UMich email doimains along
+    with simple login IDs.
+    """
+    stripped_login_id = re.sub(r'@([^@.]+\.)*umich\.edu$', '', login_id, flags=re.IGNORECASE)
```
jonespm: Is this regex needed? Would a split just work? This seems like it would replace the method

```
def process_login_id(login_id: str) -> str:
    umich_domain = "@umich.edu"
    return login_id.split('@')[0] if login_id.lower().endswith(umich_domain) else login_id.replace('@', '+', 1)
```

requirements.txt
```diff
@@ -30,6 +30,7 @@ canvasapi==3.3.0 # For Canvas API
 #django-q
 django-q2==1.8.0
 watchfiles==1.0.5 # For file watching
+blessed==1.21.0
```
jonespm, line 34: Blessed looks like an cool library but is this code using it or just left over from some testing?

pushyamig: The django-q code uses it, we just need to add this package. The purpose is to see the Django-q components from terminal. Some of it available from Django admin since we uses Django ORM ( as the Broker) for storing Django-q tasks updates.

I have add more detail on Django-q in [Readme](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/560/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)
https://django-q2.readthedocs.io/en/master/monitor.html
1. `python manage.py qinfo` . This is to just quickly check on-off Queue stats.  More info [here](https://django-q2.readthedocs.io/en/master/monitor.html#info)
2. `$ python manage.py qmonitor`. The cluster monitor shows live information about all the Q clusters. I did not see the purpose for stating this a command with `start.sh` since we have Django admin. With that said I welcome feedback on this. More info [here](https://django-q2.readthedocs.io/en/master/monitor.html#monitor)

pushyamig: `qinfo`
Screenshot 2025-08-25 at 12 18 45 PM

`qmonitor`
Screenshot 2025-08-25 at 12 19 25 PM

jonespm: Okay, I see. It's an optional dependency for that other one. I'd also add a comment about that to the requirements.txt and remove the commented out django-q. Is watchfiles dependency used for this or separate?

jonespm (approved):
There's a lot packed into this change but it looks like it's setup good to me. Can wait to see if @jaydonkrooss has any feedback.

jaydonkrooss:
code looks good to me and locally tests well on my end. I agree with @jonespm's suggestion to move `MAX_ALLOWED_ENROLLMENTS` to env variable unless you see an argument otherwise.

beyond that I think its ready to be put on the test environment.

jaydonkrooss (approved):

requirements.txt
pushyamig: I have made changes based on the comment. `watchfiles` are used since django-q is a supervisor process during development it needs to be reload code base changes.

backend/ccm/canvas_api/enroll_users.py
pushyamig: I think we need to account for following scenarios, the above code not quiet working as supposed to. 
```
print(process_login_id('janedoe'))
print(process_login_id('janedoe@umich.edu'))
print(process_login_id('janedoe@med.umich.edu'))
print(process_login_id('janedoe@gmail.com'))
print(process_login_id('janedoe.smith@dept.umich.edu')) 
print(process_login_id('janedoe@OTHER.EDU'))

output: 
janedoe
janedoe
janedoe+med.umich.edu
janedoe+gmail.com
janedoe.smith+dept.umich.edu
janedoe+OTHER.EDU

```

pushyamig: But I think your concern is coming from using Regex, so I change the code not use regex made some change, now it print as below
```
janedoe
janedoe
janedoe
janedoe+gmail.com
janedoe.smith
janedoe+OTHER.EDU
```

backend/ccm/canvas_api/constants.py
```diff
@@ -0,0 +1,22 @@
...
+
+ROLE_TO_ENROLLMENT_TYPE = {
+    role: f"{role.capitalize()}Enrollment"
+    for role in ALLOWED_ROLES
+    if role not in ("assistant", "librarian")
+}
+
+MAX_CONCURRENCY = 10
```
pushyamig, line 22: adding Concurrency as a constance as part of feedback received https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/552/files#r2298380130

backend/ccm/canvas_api/constants.py
pushyamig: I can make this as configurable but If I made it like that
1. I need to add that to `.env`
2. I Need to pass it as FE using context processors and made it available as props, which is doable

As I understood the intent is for purpose of testing, then during the QA'ing the person can opt to test for lower limit like 3000 is they want to.  am I getting your perspective?

I don't see a huge advantage making it as configurable in Backend only. Both Backend and Frontend validate for max limit.

pushyamig:
I am merging this now, but if the conversation about the [configuration limit](https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/560#discussion_r2298294230) still needs to set I will address it as separate issue. I need some clarity on this will chat tomorrow in team meeting. 

Thanks for all the Review

PR #562, opened 2025-08-25
Fix webpack and mariadb

jonespm:
There was this build error related to webpack and Mariadb not building. I'm putting this out there as a draft as this is what got me going. The webpack error was related to browserlist, not webpack directly.


jonespm:
Better issue in #564

PR #564, opened 2025-08-26
Fixes for some recent changes for npm and webpack related to trixie update

### LINKED ISSUE #563: Issue starting up with trixie, npm not found and error starting webpack
There was some new error today where npm wasn't found. Initially we weren't sure but this seems like the base image wasn't explicit and upgraded from bookworm to trixie. 

Also after starting with webpack there's another error related to caniuse-lite being outdated and the service won't start. This is a dependency from webpack and updating webpack updates this. 

I found this thread and some ways to work around this but a webpack upgrade might just be enough. 

https://stackoverflow.com/questions/55271798/browserslist-caniuse-lite-is-outdated-please-run-next-command-npm-update-cani

jonespm:
Fixes #563
Install npm
Change paths for webpack

Draft: Will revisit this one after the November release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20-bookworm-trixie AS node-build
+FROM node:20-trixie-slim AS node-build
 WORKDIR /build/
 
 COPY ccm_web .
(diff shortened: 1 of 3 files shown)


jonespm:
Closing this as the webpack changes have been made already so this should no longer be an issue

PR #566, opened 2025-08-26
Hold back build base images to bookworm

### LINKED ISSUE #565: Debian upgraded to trixie, hold back for now to bookworm
Debian base images upgrade to trixie in early August. Since we're not explicit our base images got upgraded. We should hold them back and do a separate task to upgrade to trixie.

jonespm:
Fixes #565

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20-slim AS node-build
+FROM node:20-bookworm-slim AS node-build
 WORKDIR /build/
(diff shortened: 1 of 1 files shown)


pushyamig:
The fix is correctly building and starting CCM, app is functional. npm watch is also working as expected.

pushyamig (approved):
Thank you for quickly coming up with a fix about the root cause of the problem.

PR #573, opened 2025-08-29
#561 adding a getting user info

### LINKED ISSUE #561: Check if non-um user in Canvas
For a single non-um user enrollment, we need to check if the id or user is already part of canvas. This is an API call to canvas.

This canvas API call use Canvas Admin token , we will be receiving the email id
```
const safeLoginId = loginId.replace('@', '+')
endpoint = `users/sis_login_id:${safeLoginId}
```

With canvasapi, the below API call is working fine
```
def user_stuff(canvas_apis: Canvas):
 
    login_id = "testuser+gmail.com"
    user = canvas_apis.get_user(login_id, 'sis_login_id')
    print(f"User object created: {user.login_id}")

### PR DESCRIPTION
**Pull request by pushyamig**
Fixes #561 

**Config Change needed**: Take the config from Dropbox, if you ran a Node app it should be available. 
`CANVAS_ADMIN_API_TOKEN`

The PR checking if the login_id is available or not.  Tests are written to support the UI. 
New function created in CanvasCredentialManager class to support the Admin token, this token don't have any expiry and scoped to user manage permission and don't do need to go via Canvas API key setup for scope setting.  

**Test Plan**
1. Give an existing Non-Umich email id. You can look at the course (403334) for an existing one
2. Give a new Non-Umich email Id, just make some random
3. UI should message appropriate to users. Either recognize as new user or existing user

### DIFF
--- a/backend/ccm/canvas_api/canvas_user_handler.py
+++ b/backend/ccm/canvas_api/canvas_user_handler.py
@@ -1,0 +1,46 @@
+import logging
+from http import HTTPStatus
+from rest_framework.views import APIView
+from rest_framework_tracking.mixins import LoggingMixin
+from rest_framework import authentication, permissions
+from rest_framework.request import Request
+from rest_framework.response import Response
+from canvasapi import Canvas
+from canvasapi.exceptions import CanvasException
+from backend.ccm.canvas_api.canvas_credential_manager import CanvasCredentialManager
+from backend.ccm.canvas_api.canvasapi_serializer import CanvasObjectROSerializer, LoginIdSerializer
+from .exceptions import CanvasErrorHandler, HTTPAPIError
+
+logger = logging.getLogger(__name__)
+
+class CanvasUserHandler(LoggingMixin, APIView):
+    logging_methods = ['GET']
+    authentication_classes = [authentication.SessionAuthentication]
+    permission_classes = [permissions.IsAuthenticated]
+    user_allowed_fields = ['name']   
+    serializer_class = LoginIdSerializer
+
+    def __init__(self, credential_manager=None):
+        self.credential_manager = credential_manager or CanvasCredentialManager()
+        self.canvas_error = CanvasErrorHandler()
+        super().__init__()
+    
+    def get(self, request: Request, login_id: str) -> Response:
+        # Validate login_id is an email address
+        serializer = LoginIdSerializer(data={'login_id': login_id})
+        if not serializer.is_valid():
+            self.canvas_error.handle_serializer_errors(serializer.errors,str(login_id))
+            return Response(self.canvas_error.to_dict(), status=self.canvas_error.to_dict().get('statusCode'))
+        
+        canvas_api: Canvas = self.credential_manager.get_canvasapi_admin_instance()
+        safe_login_id: str = login_id.replace('@', '+')
+        try:
+            user_info = canvas_api.get_user(safe_login_id, 'sis_login_id')
+            append_fields ={"login_id": login_id}
+            serializer = CanvasObjectROSerializer(user_info, allowed_fields=self.user_allowed_fields, append_fields=append_fields)
+            return Response(serializer.data, status=HTTPStatus.OK)
+        except (CanvasException, Exception) as e:
+            self.canvas_error.handle_canvas_api_exceptions(HTTPAPIError(str(login_id), e))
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by pushyamig** [comment]
I am resolving conflicts with this PR due to the Merges to `develop` branch today

**Comment by pushyamig** [comment]
conflicts are resolved and got the latest from develop

File: backend/ccm/canvas_api/canvas_user_handler.py [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+        super().__init__()
+    
+    def get(self, request: Request, login_id: str) -> Response:
+        # Validate login_id is an email address
+        serializer = LoginIdSerializer(data={'login_id': login_id})
+        if not serializer.is_valid():
+            self.canvas_error.handle_serializer_errors(serializer.errors,str(login_id))
+            return Response(self.canvas_error.to_dict(), status=self.canvas_error.to_dict().get('statusCode'))
+        
+        canvas_api: Canvas = self.credential_manager.get_canvasapi_admin_instance()
```
jaydonkrooss: nitpick: I would rename canvas_api in this case to canvas_api_admin to not cause any confusion, since we've used variable name "canvas_api" to represent the user's api permission

backend/tests/test_canvas_user_handler.py
jaydonkrooss, line 1: I think we should have a test for 404 response, since that is a common case if the user search results in "not found" from the canvas api

backend/ccm/canvas_api/canvasapi_serializer.py
```diff
@@ -116,4 +116,8 @@ def to_representation(self, instance):
                 if key not in data:
                     data[key] = value
         
-        return data
\ No newline at end of file
+        return data
+    
+# Serializer to validate login_id as an email address
+class LoginIdSerializer(serializers.Serializer):
+    login_id = serializers.EmailField()
```
jaydonkrooss, line 123: I'm assuming we don't have any non-UM user emails in production that would fail this validation, but are we sure that is the case? The Node backend didn't have any validation before, so I'd hope this doesn't block finding users that already exist.

jaydonkrooss (approved):
Local testing in app and in swagger looks to be working correctly for found (200), not found (404), and invalid (500). Unit tests pass as well, I only have the one suggestion to add one more unit test for the 404 case. 

I know this PR is a blocker for continuing non-Um enrollment feature, so I'm approving this PR. My rename variable suggestion and my question for email validation are not essential

backend/ccm/canvas_api/canvasapi_serializer.py
pushyamig: This Backend validation is just checking if the email received from FE is proper email format like `test@example.com`. Frontend End is already checking for it, we do check in the Backend as an extra step.

>we don't have any non-UM user emails in production that would fail this validation, but are we sure that is the case?

I am not clear what you mean by this,  if an non-umich email already added to Prod, the API response is the only way we know if the email present or not and hence we need to make this API call.  Validation here is all for the basic formatting check that's all

backend/ccm/canvas_api/canvas_user_handler.py
pushyamig: Done

backend/tests/test_canvas_user_handler.py
pushyamig: OK, I added test for 404 use case.

backend/ccm/canvas_api/canvasapi_serializer.py
jaydonkrooss: Right, I have no problem with backend email format checking, it's a good step to add. I'm only thinking of a (probably impossible) edge case, if there's a login_id out there that _would_ be a match in production but instead fails format checking before the Canvas API searches. The search would proceed in the old Node version, but would be blocked in the new Python version.

I'm not suggesting to make any changes because its not really a valid use case. I'm just curious if you're aware that all login_id's that are a possible match would 100% be the correct email format that `serializers.EmailField()` expects

pushyamig: Even it passed this step of checking if user exists or not.Next step with Creating user step is again validating the it in [Node code](https://github.com/pushyamig/canvas-course-manager-next/blob/i569_add_non_um_user/ccm_web/server/src/api/dtos/api.external.users.dto.ts#L19), So probably it is taken care. So chances of that might be less (edited)

We chatted offline on this ....I am resolving this issue now.

PR #575, opened 2025-08-29
#574 Remove is_root_admin custom variable

### LINKED ISSUE #574: Remove LTI Custom variable about is_root_admin
CCM is not using the LTI Custom Variable `Canvas.user.isRootAccountAdmin` anywhere in the code, only launch variables that are sent to Frontend is `"course": {"id": 12345, "roles": ["TeacherEnrollment", "Account Admin"]`. So this issue is removing the redundant custom variable not being used

pushyamig:
Fixes #574 
**LTI config change needed**
LTI launch should work for admin user and show all the features for them

### DIFF
--- a/backend/ccm/lti_config.py
+++ b/backend/ccm/lti_config.py
@@ -22,7 +22,7 @@ class LTILaunchError(Exception):
 
 class CCMLTILaunchView(LtiLaunchBaseView):
 
-    LTI_CUSTOM_PARAMS: List[str] = ['roles', 'is_root_account_admin', 'login_id', 'course_id']
+    LTI_CUSTOM_PARAMS: List[str] = ['roles', 'login_id', 'course_id']
     LTI_CUSTOM_PARAMS_URL: str = 'https://purl.imsglobal.org/spec/lti/claim/custom'
     
     def validate_custom_lti_launch_data(self, lti_launch: TLaunchData) -> None:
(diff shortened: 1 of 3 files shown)


jonespm:
So this wasn't used? 

Do we have or need some way to tell they're root admin? Was this something in the existing version of CCM?

pushyamig:
>So this wasn't used?

yes

pushyamig:
>Do we have or need some way to tell they're root admin? Was this something in the existing version of CCM?

We are using this custom variable, `"roles": "$Canvas.membership.roles",` to determine admin account or not. this cutom variable return are Canvas roles and not LTI roles.
we are sending it as `ccm_globals` to FE: `"course": {"id": 403334, "roles": ["TeacherEnrollment", "Account Admin"]`
Logic for enabling feature is done here https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/develop/ccm_web/client/src/models/feature.ts

pushyamig:
IN Node CCM, we send it `/globals` API call, in Python CCM set via Context processor sent as script tag to Frontend

jaydonkrooss (approved):
python app is functional with local testing and all backend unit tests pass

zqian:
verified on CCM dev: the`is_root_account_admin=$Canvas.user.isRootAccountAdmin` is not included in LTI custom fields

PR #585, opened 2025-10-01
#584 LTI misc role check and denial access

### LINKED ISSUE #584: CCM should show error msg for invalid roles
if user doesn't have atleast below allowed roles, then CCM should throw an message that they are not allowed to access this tool. The requirement for this is given [here](https://docs.google.com/spreadsheets/d/1R0fqbKfTzICtqeWmysVouudReFbi6sfnxcTniCDJPvI/edit?gid=471747967#gid=471747967) 

ALLOWED_LTI_ROLES = {
    "Account Admin",
    "Sub-Account Admin",
    "Support Consultant",
    "TeacherEnrollment",
    "DesignerEnrollment",
    "TaEnrollment",
    "Assistant",}

So Backend should throw error before the App load and when LTI steps are happening.

Note that Node code is handling it from UI, but the Python the static assets is only loaded at this point so we can't have similar experience UI experience. But the goal of  informing the user is accomplished.

pushyamig:
Fixes #584

The similar LTI check in Node code is [here](https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/develop/ccm_web/server/src/lti/lti.service.ts#L73)

### DIFF
--- a/backend/ccm/lti_config.py
+++ b/backend/ccm/lti_config.py
@@ -19,11 +19,24 @@ class LTILaunchError(Exception):
     Exception class for errors that occur while processing data from the LTI launch
     """
 
+class LTINotAllowedRolesError(Exception):
+    """
+    Exception class for users with roles that are not allowed to access the tool
+    """
 
 class CCMLTILaunchView(LtiLaunchBaseView):
 
     LTI_CUSTOM_PARAMS: List[str] = ['roles', 'login_id', 'course_id']
     LTI_CUSTOM_PARAMS_URL: str = 'https://purl.imsglobal.org/spec/lti/claim/custom'
+    ALLOWED_LTI_ROLES = {
+    "Account Admin",
+    "Sub-Account Admin",
+    "Support Consultant",
+    "TeacherEnrollment",
+    "DesignerEnrollment",
+    "TaEnrollment",
+    "Assistant",
+    }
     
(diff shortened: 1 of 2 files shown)


backend/ccm/lti_config.py
```diff
@@ -79,10 +92,25 @@ def login_user_store_session(self, request, launch_data, user_obj):
...
+        """
+        Validate if the user has at least one of the allowed LTI roles.
+        Raises an exception if the user has no allowed roles.
+        """
+        roles: str = launch_data[self.LTI_CUSTOM_PARAMS_URL].get('roles', '')
+        roles_list: List[str] = roles.split(',') if roles else []
+        
+        # Check if user has at least one allowed role
+        if not any(role in self.ALLOWED_LTI_ROLES for role in roles_list):
+            error_msg = f'To use Canvas Course Manager, you must launch the tool from a Canvas course in which you have a course management role.<br />'
```
jonespm: Minor but no need to use f strings here, just remove the f' from both of these.

jonespm (approved):
This looks good, we may want to have a future Issue where we make these roles customizable in settings or admin UI. But as discussed on the meeting that would also involve changes in the frontend and these don't need to be updated very often. So doesn't have to happen in this release.

backend/ccm/lti_config.py
pushyamig: Done

PR #590, opened 2025-10-08
#589 exit getting course early when limit is reached

### LINKED ISSUE #589: Merge section: Fix the timeout issue on Openshift for lazy searches
When user types searches like  `101` , `001` and `102`, this will result in fetching lot of courses and then "Too many course... error is thrown' this would result in user waiting.. waiting.. on fetching all courses then throw the error. 

So this fix is to stop the pagination if it met the limit. Currently we limit this to >=400, if we say >400 then it is another API call. I feel 400 should be good enough count to mitigate the problem

pushyamig:
Fixes #589 

Check for Limit is happening at 2 places
1. Getting courses for single account call
2. Top level after fetching all the courses in multiple account before proceeding to section call. This way sections are not called
3. If user is part of multiple account like [1, 232, 34], only searches courses from [1]. Root account has all courses , this will save some time and is an improvement.

One side effect of this is getting courses in multiple account is an async process so if user is part of multiple accounts there is no way to stop the process until all the calls that are finished. since we are using a combination of coroutines + Python thread. There is no way a thread can be stopped once started. we are using this combination approach since `canvasapi` is sync HTTP client. ~~If we have to stop `gather` tasks it will be a complicated solution~~.  

Since Getting courses for all the accounts the user belongs to is async call, and Sub-account admin might be practically be part of 2 or 3 account, so the async call (let say for fetching from 3 accounts) all starts at same time, so the amount of time take seem to be same when API call hits 400 count limit. I personally feel this is a Optimal solution since the blocking of time is coming from Pagination getting more courses from the account due to sync HTTP client.  

Currently for seeing `too many courses error` taking about **~8sec**, Node CCM takes **5sec**

I feel we still should extend the timeout on the router, currently without nothing it will times out 30s. Current setting is 30min. since 400 courses making sections call might take another ~17secs

Unit Test written to check this specific use case

----------------------------------
Test Plan 1: 
1. Add your self to Aerospace Engineering [211] account, you should be root admin
2. Search by "AEROSP" 
3. This should only fetch courses from Root account

-----------------------------------
Test Plan 2: 
1. Pick a user who is part of multiple account ( but not root)
2.  Search by  common course text '101'
3. This should exit in about 10 sec and also fetches courses of those 2 sub-accounts

### DIFF
--- a/backend/ccm/canvas_api/admin_sections_api_handler.py
+++ b/backend/ccm/canvas_api/admin_sections_api_handler.py
@@ -9,6 +9,7 @@ from rest_framework.request import Request
 from asgiref.sync import async_to_sync
 from drf_spectacular.utils import extend_schema, OpenApiParameter
 from drf_spectacular.types import OpenApiTypes
+from itertools import islice
 
 from canvasapi.exceptions import CanvasException
 from canvasapi.account import Account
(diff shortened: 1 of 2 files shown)


pushyamig:
I converted this to draft, I am doing some additional use case testing, will demo in the meeting!

pushyamig:
This PR is ready for review!

jaydonkrooss (approved):
Looks good, nice solution with islice! Local testing works in cases of admins of multiple sub-accounts (with faster error throwing for too many). Unit tests pass.

PR #598, opened 2025-10-16
#597 make app TZ as configuraable

### LINKED ISSUE #597: Set APP TZ to configurable option
Currently the TZ is set to UTC as part of Django Default, make it configurable to `America/Detroit`

pushyamig:
Fixes #597

 The best way to test is 
1. Launch CCM
2. Go to Admin --> API Request Logs
3. this should show the timestamp based on the `TIME_ZONE` setting

### DIFF
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -155,8 +155,7 @@ AUTH_PASSWORD_VALIDATORS = [
 
 LANGUAGE_CODE = 'en-us'
 
-TIME_ZONE = os.getenv('TIME_ZONE', 'America/Detroit')
-
+TIME_ZONE = os.getenv('TIME_ZONE', os.getenv("TZ", "America/Detroit"))
 USE_I18N = True
 
 USE_TZ = True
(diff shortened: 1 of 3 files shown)


jonespm:
So the Dockerfile is setting an environment variable already TZ based on if a build value was passed.

https://github.com/tl-its-umich-edu/canvas-course-manager-next/blob/1f43913f3716c0092bb732b2d4a930f85ae5c97f/Dockerfile#L61

I think it's probably fine to have a build time zone and a application timezone separated and also check if TZ is set. This is what the code from MyLA did.

So I'd call something like.

`os.getenv('TIME_ZONE', os.getenv('TZ', 'America/Detroit'))
`

pushyamig:
>So I'd call something like.
os.getenv('TIME_ZONE', os.getenv('TZ', 'America/Detroit')) 

Done

pushyamig:
I made the changes as requested, and added test to support that.

jonespm (approved):
Looks good, time zone of Django is using the configured timezone now.

PR #600, opened 2025-10-20
#599 Disable Django backend URL

### LINKED ISSUE #599: Disable existing admin backend
This issue when try to use /admin enable /admin/login, which should be shown. we are using `ENABLE_BACKEND` to disable this by default. But this is useful when doing the Load testing.

pushyamig:
Fixes #599 

Test plan: https://docs.google.com/spreadsheets/d/1pPJ5dRe-Za1ZVV2mFHKBISUvcRL7SxRdxuH28jyOl-0/edit?gid=0#gid=0

### DIFF
--- a/backend/urls.py
+++ b/backend/urls.py
@@ -18,11 +18,22 @@ from django.conf import settings
 from django.contrib import admin
 from django.urls import include, path
 from lti_tool.views import jwks, OIDCLoginInitView
-from backend.ccm.canvas_api.course_api_handler import CanvasCourseAPIHandler
 from backend.ccm.lti_config import CCMLTILaunchView
+from django.http import HttpResponseForbidden
 import watchman.views
 
 from backend import views
+
+def disabled_admin_login(request, *args, **kwargs):
+    return HttpResponseForbidden("View is disabled.")
+
+# Disable the Django admin login page by assigning the simple view directly.
+if not getattr(settings, "ENABLE_BACKEND", False):
+    # In normal deployments, disable the admin login page so the app doesn't expose admin url.
+    # For local load testing where you need the backend enabled set ENABLE_BACKEND=True
+    admin.site.login = disabled_admin_login
+
+
 
 urlpatterns = [
     path('admin/', admin.site.urls),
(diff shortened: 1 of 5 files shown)


backend/urls.py
```diff
@@ -18,12 +18,23 @@
...
 from backend import views
 
+def disabled_admin_login(request, *args, **kwargs):
+    return HttpResponseForbidden("View is disabled.")
+
+# Disable the Django admin login page by assigning the simple view directly.
+if not getattr(settings, "ENABLE_BACKEND", False):
+    # In normal deployments, disable the admin login page so the app doesn't expose admin url.
+    # For local load testing where you need the backend enabled set ENABLE_BACKEND=True
+    admin.site.login = disabled_admin_login
```
pushyamig, line 34: By Default, Admin ( with super_user and is_staff) role are given access to `/admin` page. it's Django checking pattern. We don't control unless we disable `path('admin/', admin.site.urls),`

pushyamig:
More or less this check is influenced by this change in Instructor Tools: admin.site.login = user_passes_test(False, admin.site.login)

is did not use this `user_passes_test(False, admin.site.login) ` since that just throws error in the `Server Error`, just felt some message "View disabled` would be little descriptive ( this is a syntax sugar). Open with ideas. 

With CCM we only use the tool via LTI and not using login page, which Myla do only locally. Felt this fix is suffice. But I Open to change and welcomed. Test Plan is written when I tested this change with.

config/.env.sample
```diff
@@ -108,4 +114,7 @@ CANVAS_ADMIN_API_TOKEN=<your_canvas_admin_api_token>
 GUEST_ACCOUNT_CREATION_LINK=https://friend.weblogin.umich.edu/friend/
 
 # Set the time zone for the application
-# TIME_ZONE=America/Detroit
\ No newline at end of file
+# TIME_ZONE=America/Detroit
+
+# Enable Backend
+# ENABLE_BACKEND=False
```
jaydonkrooss: Remove from sample, I see it's already added above

pushyamig: Done

jaydonkrooss (approved):
local testing passes successfully on my end. code looks good.

I think eventually, we'll want to add load testing instructions in the readme, but that is obviously not in scope yet.

PR #675, opened 2026-06-16
#666 adding ssl support

### LINKED ISSUE #666: Add database .pem SSL file to DB config
With the recent MySQL upgrade to 8.0 and 8.4, it is recommended to add a Pem file making sure connections are encrypted.. this needs a code change

So currently connections are on a default connection mode because we do not specify that ssl mode. So it defaults to a preferred. However, if you specify the mode to verify_ca or required, then you present that global-bundle.pem and it establishes a secured connection because the cert you present is verified with what is installed in the db and encryots your connection.

pushyamig:
Fixes #666

### DIFF
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -118,7 +118,6 @@ def get_mysql_options() -> dict:
     if db_ssl_ca:
         options['ssl'] = {'ca': db_ssl_ca}
         options['ssl_mode'] = db_ssl_mode
-    print(options)
 
     return options
 
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -108,11 +108,18 @@ WSGI_APPLICATION = 'backend.wsgi.application'
 
 
 def get_mysql_options() -> dict:
+    """
+    Build database OPTIONS dict with conditional SSL support.
+    
+    SSL Configuration:
+    - DB_SSL_MODE: Values include 'REQUIRED', 'VERIFY_IDENTITY', 'VERIFY_CA', 'DISABLED', 'PREFERRED'
+    - DB_SSL_CA: Path to CA certificate (optional, used for verification)
+    """
     options = {
         'charset': 'utf8mb4',
     }
     db_ssl_ca = os.getenv('DB_SSL_CA')
-    db_ssl_mode = os.getenv('DB_SSL_MODE', 'VERIFY_CA')
+    db_ssl_mode = os.getenv('DB_SSL_MODE', 'REQUIRED')
 
     # Configure MySQL SSL only when a CA cert path is provided.
     if db_ssl_ca:
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -106,6 +106,22 @@ WSGI_APPLICATION = 'backend.wsgi.application'
 # Database
 # https://docs.djangoproject.com/en/5.2/ref/settings/#databases
 
+
+def get_mysql_options() -> dict:
+    options = {
+        'charset': 'utf8mb4',
+    }
+    db_ssl_ca = os.getenv('DB_SSL_CA')
+    db_ssl_mode = os.getenv('DB_SSL_MODE', 'VERIFY_CA')
+
+    # Configure MySQL SSL only when a CA cert path is provided.
+    if db_ssl_ca:
+        options['ssl'] = {'ca': db_ssl_ca}
+        options['ssl_mode'] = db_ssl_mode
+    print(options)
+
+    return options
+
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.mysql',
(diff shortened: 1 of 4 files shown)


backend/settings.py
```diff
@@ -106,6 +106,22 @@
...
+        'charset': 'utf8mb4',
+    }
+    db_ssl_ca = os.getenv('DB_SSL_CA')
+    db_ssl_mode = os.getenv('DB_SSL_MODE', 'VERIFY_CA')
+
+    # Configure MySQL SSL only when a CA cert path is provided.
+    if db_ssl_ca:
+        options['ssl'] = {'ca': db_ssl_ca}
+        options['ssl_mode'] = db_ssl_mode
+    print(options)
```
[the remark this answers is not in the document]
pushyamig: Done

docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
       - redis
 
   database:
-    image: mysql:8-oracle
+    image: mysql:8.4-oracle
```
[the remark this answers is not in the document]
pushyamig: Done

pushyamig:
I took the feedback from PR https://github.com/tl-its-umich-edu/instructor-tools/pull/700 and implemented here

backend/settings.py
```diff
@@ -106,6 +106,21 @@
 # Database
 # https://docs.djangoproject.com/en/5.2/ref/settings/#databases
 
+
+def get_mysql_options() -> dict:
+    options = {
+        'charset': 'utf8mb4',
+    }
+    db_ssl_ca = os.getenv('DB_SSL_CA')
+    db_ssl_mode = os.getenv('DB_SSL_MODE', 'VERIFY_CA')
```
zqian: Copilot suggested use 'VERIFY_IDENTITY' instead of 'VERIFY_CA' as the default setting, since the later does not verify hostname matches. 

Can we do VERIFY_IDENTITY?

zqian (commented):
I followed the test steps documented in https://github.com/tl-its-umich-edu/instructor-tools/pull/700, and each step completed successfully.

Could you please copy those test steps in https://github.com/tl-its-umich-edu/instructor-tools/pull/700 into this PR?

backend/settings.py
jonespm: I didn't pay attention to this but the SSL MODE we'll be ultimately setting this to at launch is `REQUIRED`. That's the only one that will work with our current certificate so might as well use that as default.

`VERIFY_IDENTITY `is the highest security level but we cannot use that or `VERIFY_CA `with our current certificate. So should just use it `REQUIRED `to match placement exams. I can understand the suggestion but it doesn't know our situation.

https://github.com/tl-its-umich-edu/placement-exams/pull/94/changes

Here was a nice breakdown of the options, with our current certificate we can't go higher than `REQUIRED`. 

Mode | Security Level | Requirement
-- | -- | --
DISABLED | None | Connection is strictly unencrypted (plain text).
PREFERRED | Low | Tries SSL; if the server fails, it falls back to unencrypted.
REQUIRED | Medium | Requires SSL. The connection fails if the server doesn't support it.
VERIFY_CA | High | Requires SSL and verifies the Server's CA certificate against your .pem file.
VERIFY_IDENTITY | Maximum | Same as VERIFY_CA, but also checks if the hostname matches the certificate.

zqian: Sure! REQUIRED is enough for current needs.

To be consistent, the [Instructor Tools SSL MODE setting](https://github.com/tl-its-umich-edu/instructor-tools/pull/700/changes) needs to be REQUIRED, too.

pushyamig: I added the fix with default to REQUIRED. 
so initially i had REQUIRED and opted for `VERIFY_CA`. as from description `Requires SSL and verifies the Server's CA certificate against your .pem file.` so we only have a CA file one file, typically from the [MYSQL docs ](https://dev.mysql.com/doc/refman/8.4/en/using-encrypted-connections.html)I noticed there are 2 more certs added  client `key` and `cert`.  

In the email Clement shared that since we have CA file and he suggested `VERIFY_IDENTITY`

The .pem file doesn't work Locally, and in order to understand what option works we need to test this in Openshift only. 

@jonespm I forwarded the email what clement shared about the Pem and ssl mode options.

zqian (approved):

backend/settings.py
jonespm: So it looks like a lot of this was merged before I could reply. 

We could use `VERIFY_IDENTITY `if we changed which domain name we used in the connection string. If we used the actual full internal `amazonaws.com` domain name then I believe that setting would work, but we'd there are situations that internal name will change.

If we used the CNAME we have `tl.mysql.` we can only use REQUIRED as the highest setting since this hostname isn't part of the certificate. Clement did say it could be possible in the future to both use this hostname and the certificate but wouldn't be done soon. 

So realistically I think for our cases just requiring SSL is fine for now and was less changes needed. I think eventually we may update it to `VERIFY_IDENTITY` but there isn't a lot of value and isn't really much difference for our case.

